PR #6, opened 2017-07-28
Add django tracer

liyanhui1228:
Should merge the flask PR first then merge this one.

### DIFF
--- a/trace/tests/unit/tracer/test_django_tracer.py
+++ b/trace/tests/unit/tracer/test_django_tracer.py
@@ -1,0 +1,108 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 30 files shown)


trace/tests/unit/tracer/test_django_tracer.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_no_context_header(self):
+        header = self._call_fut()
+        self.assertIsNone(header)
+
+    def test_valid_context_header(self):
+        from django.test import RequestFactory
+        from opencensus.trace.tracer.middleware import request
+
+        django_trace_header = 'HTTP_X_CLOUD_TRACE_CONTEXT'
+        trace_id = 'testtraceiddjango'
```
duggelz: Shouldn't this be an error?

liyanhui1228: This is just test getting trace header from a django request, not include checking the format. Format is checked in `google_cloud_format.from_header`.

PR #8, opened 2017-08-01
Add a logging reporter to send the traces to logging

liyanhui1228:


duggelz:
Would be nice to have a unit test

liyanhui1228:
Unit test in progress. Will add it and let the CircleCI pass before merge. Thanks!

PR #9, opened 2017-08-01
Add decorator to trace a function

liyanhui1228:
Usage as below:

```
from opencensus.trace.tracer.context_tracer import ContextTracer
tracer = ContextTracer()

@tracer.trace_decorator()
def hello(a):
    print(a)

hello()

>>> tracer.cur_trace.spans
[<opencensus.trace.trace_span.TraceSpan object at 0x7fad81d42160>]
>>> tracer.cur_trace.spans[0].name
'hello'
```

### DIFF
--- a/trace/opencensus/trace/tracer/context_tracer.py
+++ b/trace/opencensus/trace/tracer/context_tracer.py
@@ -78,6 +78,21 @@ class ContextTracer(object):
         else:
             return False
 
+    def trace_decorator(self):
+        """Decorator to trace a function."""
+
+        def decorator(func):
+
+            def wrapper(*args, **kwargs):
+                self.start_span(name=func.__name__)
+                return_value = func(*args, **kwargs)
+                self.end_span()
+                return return_value
+
+            return wrapper
+
+        return decorator
+
     def trace(self):
         """Create a trace using the context information.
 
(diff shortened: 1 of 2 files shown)


trace/opencensus/trace/tracer/context_tracer.py
```diff
@@ -78,6 +78,21 @@ def set_enabled(self):
         else:
             return False
 
+    def trace_decorator(self):
+        """Decorator to trace a function."""
+
+        def decorator(func):
+
+            def wrapper(*args, **kwargs):
```
duggelz: I'm not sure this is exactly what we want.  This will trace every execution of the wrapped function in the same context.  It seems like the wrapper should fetch the trace context every time it is executed.

liyanhui1228: To me it's reasonable to trace the function for every execution in a context. Because it might belong to different parent span, and worth to know the latency of each execution. And the `start_span` and `end_span` method in the wrapper has the information in context tracer and span context, like whether this function should be sampled and what's the parent span.

duggelz: Maybe I'm misunderstanding this.  Suppose I have two http request handlers, for `/foo` and `/bar`.  They each call `hello`.  `/foo` has a trace context, `/bar` has a different trace context.  The handlers both call `hello`.  The calls to hello will add spans to a third context, unrelated to the `/foo` and `/bar` contexts?

liyanhui1228: Yes, that's correct. The calls to hello will add spans to the context belong to the decorator which is specified in `@tracer.trace_decorator()`. I think the decorator usage for tracing is intended to use in the client side to trace the function that makes a request to the server.

PR #10, opened 2017-08-03
Add webapp2 tracer

liyanhui1228:


trace/opencensus/trace/tracer/webapp2_tracer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    :rtype: str
+    :returns: Trace context header in HTTP request headers.
+    """
+    if webapp2 is None:  # pragma: NO COVER
+        return None
+
+    try:
+        request = webapp2.get_request()
+    except AssertionError:
```
duggelz: Why AssertionError?

liyanhui1228: When it's not running inside a webapp2 request context, it will raise `AssertionError: Request global variable is not set.`

PR #11, opened 2017-08-05
Add text format propagation

liyanhui1228:
This is for generating/extracting span context which is not in google cloud format.

### DIFF
--- a/trace/opencensus/trace/propagation/text_format.py
+++ b/trace/opencensus/trace/propagation/text_format.py
@@ -1,0 +1,79 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module provides the basic utilities for extracting the trace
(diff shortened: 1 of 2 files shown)


trace/opencensus/trace/propagation/text_format.py
```diff
@@ -0,0 +1,79 @@
...
+    for key in carrier:
+        key = key.lower()
+        if key == _TRACE_ID_KEY:
+            trace_id = carrier[key]
+        if key == _SPAN_ID_KEY:
+            span_id = carrier[key]
+        if key == _ENABLED_TRACE_KEY:
+            enabled = bool(carrier[key])
+
+    return SpanContext(
```
duggelz: We should handle the cases where span_id and/or options exist, but trace_id doesn't.  Probably by throwing away the span_id and options?

liyanhui1228: The default value of trace_id is None, so if it doesn't exist in the carrier, it will just be None. And the constructor of SpanContext will generate a new trace_id for it.

duggelz: Ok.  I'm looking at the SpanContext [constructor](https://github.com/census-instrumentation/opencensus-python/blob/master/trace/opencensus/trace/span_context.py#L53), and thinking that where we have:

```
        if trace_id is None:
            trace_id = trace.generate_trace_id()
```
We might add
```
            # If generating trace_id, then generate a new span_id as well
            span_id = None
            enabled = True
```

But it's an error case either way (right?), so there's no clear "correct" answer.

liyanhui1228: The `span_id` should be generated when the user calls `tracer.start_span()`. The `span_context` is for holding the information that should be propagated to the child spans, so the `span_context.span_id` is the child span's `parent_span_id`. But the `trace_id` is shared by all the things in the context, so it should be generated when the `SpanContext` is generated.

Issue #3, opened 2017-07-26
gRPC integration using client and server interceptors


liyanhui1228:
We need to support tracing Grpc requests automatically.

- [x] Implement interceptors for client and server
- [x] Add hello world example on how to use the interceptors
- [x] Update README documentation

PR #12, opened 2017-08-05
Add interceptors for grpc tracing in client side and server side

liyanhui1228:
For #3


trace/opencensus/trace/grpc/client_interceptor.py
```diff
@@ -0,0 +1,59 @@
...
+
+from opencensus.trace.enums import Enum
+
+class OpenCensusClientInterceptor(grpc_ext.UnaryClientInterceptor):
+
+    def __init__(self, tracer):
+        self._tracer = tracer
+
+    def _start_client_span(self, method):
+        logging.warn('test client span')
```
duggelz: Remove this?  Or change to `logging.debug('Start client span for %s', method)`?

trace/opencensus/trace/grpc/client_interceptor.py
```diff
@@ -0,0 +1,59 @@
...
+        span = self._tracer.start_span(name=str(method))
+        span.add_label(label_key='component', label_value='grpc')
+        span.kind = Enum.SpanKind.RPC_CLIENT
+        return span
+
+    def intercept_unary(self, method, request, metadata, invoker):
```
duggelz: Document the arguments?  In particular, what is `metadata`?

trace/opencensus/trace/grpc/client_interceptor.py
```diff
@@ -0,0 +1,59 @@
...
+        metadata = metadata + tuple(six.iteritems(headers))
+
+        with self._start_client_span(method) as span:
+            span.add_label(label_key='metadata', label_value=str(metadata))
+
+            try:
+                result = invoker(request, metadata)
+            except:
+                e = sys.exc_info()[0]
+                span.add_label('error', str(e))
```
duggelz: This is going to be the type of the exception, but not the exception's error message, if I'm reading exc_info() right.  Is that what we want?

trace/opencensus/trace/grpc/server_interceptor.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class OpenCensusServerInterceptor(grpc_ext.UnaryServerInterceptor):
+
+    def __init__(self, sampler=None, reporter=None):
+        self.sampler = sampler
+        self.reporter = reporter
+
+    def _start_server_span(self, tracer, servicer_context, method):
+        logging.warn('test server span')
```
duggelz: Remove?

liyanhui1228: Removed.

trace/opencensus/trace/grpc/client_interceptor.py
liyanhui1228: Changed to record the exception's error message.

trace/opencensus/trace/grpc/client_interceptor.py
liyanhui1228: Done.

trace/opencensus/trace/grpc/client_interceptor.py
liyanhui1228: Removed.

duggelz:
Approved, and now we just wait for GRPC work.

trace/opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -0,0 +1,137 @@
...
+            header = self._propagator.to_header(span_context)
+            grpc_trace_metadata = {
+                oc_grpc.GRPC_TRACE_KEY: header,
+            }
+            metadata = metadata + tuple(six.iteritems(grpc_trace_metadata))
+
+            span.name = '[gRPC_client][{}]{}'.format(request_type, str(method))
+
+            try:
+                result = invoker(method, request, metadata=metadata, **kwargs)
```
duggelz: It should work to do:

```
except Exception as e:
  span.add_label('error', str(e))
```

I thought a bit about `BaseException` vs `Exception`: `BaseException` includes things like `SystemExit` and out of memory errors.  In those cases, the process will probably exit/crash before the trace data is ever sent anywhere, so it's probably not worth trying to handle it.

trace/opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -0,0 +1,137 @@
...
+
+        with self._start_client_span(method) as span:
+            span_context = self._tracer.span_context
+            header = self._propagator.to_header(span_context)
+            grpc_trace_metadata = {
+                oc_grpc.GRPC_TRACE_KEY: header,
+            }
+            metadata = metadata + tuple(six.iteritems(grpc_trace_metadata))
+
+            span.name = '[gRPC_client][{}]{}'.format(request_type, str(method))
```
duggelz: As per our discussion, the hostname/port is part of the grpc Channel object, and we should probably record that.  I'm not sure how we dig out the channel from the interceptor, we might have to change `intercept_channel`.

duggelz: Actually, I'm not sure where the hostname/port is stored, looking at the grpc docs.

trace/opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -0,0 +1,136 @@
...
+        result.add_done_callback(self._future_done_callback())
+
+        return result
+
+    def _future_done_callback(self):
+        def callback(future_response):
+            code = future_response.code()
+
+            if code != grpc.StatusCode.OK:
+                span.add_label('error in response', str(code))
```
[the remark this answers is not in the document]
liyanhui1228: Yes, currently I stored all the label keys in `labels_helper.py`, will be good to unify all the labels in opencensus-specs repo, prepare that PR later.

trace/opencensus/trace/ext/grpc/client_interceptor.py
liyanhui1228: Just confirmed with Mehrdad, we cannot access the host:port in the interceptors since the interceptors are operating at the front-end of the RPC stack, and RPCs are dispatched at lower level. But we can let the user pass in the host:port when initialize the `OpenCensusClientInterceptor`, and then add this information to span labels.

trace/opencensus/trace/ext/grpc/client_interceptor.py
liyanhui1228: Changed.

opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -0,0 +1,185 @@
...
+            self._trace_future_exception(future_response)
+            self._tracer.end_span()
+
+        return callback
+
+    def _trace_future_exception(self, response):
+        # Trace the exception for a grpc.Future if any
+        exception = response.exception(timeout=TIMEOUT)
+
+        # If there is not exception, the attribute with null value will be
```
duggelz: I don't quite understand this comment.  Are you saying that exception might be None, or that str(exception) might be None?

opencensus/trace/propagation/binary_format.py
```diff
@@ -108,7 +108,7 @@ def from_header(self, binary):
         # from binary.
         try:
             data = Header._make(struct.unpack(BINARY_FORMAT, binary))
-        except struct.error:
+        except Exception:
```
duggelz: What other types of exceptions were you seeing?

liyanhui1228: Added check for None input and changed the exception type back to `struct.error`.

opencensus/trace/ext/grpc/client_interceptor.py
liyanhui1228: Thanks for noticing this, just added a check for whether the exception is None or not.

PR #17, opened 2017-08-16
Django integration for tracing requests automatically

liyanhui1228:
Remove the django tracer, instead adding the trace logic in the django middleware to enable tracing requests automatically.


trace/opencensus/trace/ext/django/config.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if user_settings is not None:
+            settings = user_settings
+
+        self.settings = settings
+
+    def __getattr__(self, attr):
+        # If not in defaults, it is something we cannot parse.
+        if attr not in DEFAULT_DJANGO_TRACER_CONFIG:
+            raise AttributeError('Attribute {} not exists.'.format(attr))
```
duggelz: `not exists` -> `does not exist`

trace/opencensus/trace/ext/django/middleware.py
```diff
@@ -0,0 +1,142 @@
...
+
+"""Django middleware helper to capture and trace a request."""
+import logging
+import threading
+
+from opencensus.trace.ext import utils
+from opencensus.trace.ext.django.config import settings
+from opencensus.trace.propagation.google_cloud_format import from_header
+
+_thread_locals = threading.local()
```
duggelz: This is ok for now, but we need to investigate the various concurrency packages.  Some use multiple threads, but others use continuations or similar.  So we'll probably want to look at ContextLocals or similar: http://werkzeug.pocoo.org/docs/0.12/local/

trace/opencensus/trace/ext/django/middleware.py
```diff
@@ -0,0 +1,142 @@
...
+    return getattr(_thread_locals, TRACER_THREAD_LOCAL_KEY, None)
+
+def _set_django_labels(span, request):
+    """Set the django related labels."""
+    django_user = getattr(request, 'user', None)
+
+    if django_user is None:
+        return
+
+    user_id = django_user.getattr(django_user, 'pk', None)
```
duggelz: I guess `pk` is the primary key of the `django.contrib.auth.models.User` object?  Maybe a comment or a link to documentation?

trace/opencensus/trace/ext/django/middleware.py
```diff
@@ -0,0 +1,142 @@
...
+
+        try:
+            # Span name is being set at process_view
+            span = tracer.start_span()
+            span.add_label(label_key=HTTP_METHOD, label_value=request.method)
+            span.add_label(label_key=HTTP_URL, label_value=request.path)
+
+            # Add current span to thread local
+            setattr(_thread_locals, TRACER_THREAD_LOCAL_KEY, tracer)
+        except Exception:
```
duggelz: Is there a more specific Exception type you can catch?

trace/opencensus/trace/samplers/fixed_rate.py
```diff
@@ -23,7 +23,7 @@ class FixedRateSampler(Sampler):
     :type rate: float
     :param rate: The rate of sampling.
     """
-    def __init__(self, rate):
+    def __init__(self, rate=0.5):
```
liyanhui1228: Done. Yeah I remember there was a discussion about this. We can always change the default when we have a conclusion later.

trace/opencensus/trace/reporters/file_reporter.py
```diff
@@ -23,7 +23,7 @@ class FileReporter(object):
     :param file_name: The name of the output file.
     """
 
-    def __init__(self, file_name):
+    def __init__(self, file_name='traces'):
```
liyanhui1228: Done.

trace/opencensus/trace/ext/utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc. All Rights Reserved.
+# Copyright 2017 Google Inc.
```
liyanhui1228: Fixed.

trace/opencensus/trace/ext/utils.py
```diff
@@ -12,6 +12,13 @@
...
-from opencensus.trace.tracer.middleware.request import RequestMiddleware
 
-__all__ = ['RequestMiddleware']
+def get_func_name(func):
+    """Return a name which includes the module name and function name."""
+    func_name = getattr(func, '__name__', func.__class__.__name__)
+
+    if hasattr(func, '__module__'):
+        module_name = func.__module__
+        return '{}.{}'.format(module_name, func_name)
```
liyanhui1228: It's the fully qualified package name.

trace/opencensus/trace/ext/django/config.py
```diff
@@ -0,0 +1,66 @@
...
+    'TRACER': 'opencensus.trace.tracer.context_tracer.ContextTracer',
+    'SAMPLER': 'opencensus.trace.samplers.always_on.AlwaysOnSampler',
+    'REPORTER': 'opencensus.trace.reporters.print_reporter.PrintReporter'
+}
+
+PATH_DELIMITER = '.'
+
+
+class DjangoTraceSettings(object):
+    """Set the params for a django tracer, including the tracer,sampler and
```
liyanhui1228: Done.

trace/opencensus/trace/ext/django/config.py
liyanhui1228: Fixed.

trace/opencensus/trace/ext/django/middleware.py
liyanhui1228: ContextLocal looks good, thank for sharing this link. I'll keep investigating the differences between the packages.

trace/opencensus/trace/ext/django/middleware.py
liyanhui1228: Yeah it is the django auto field for User model, which is the primary key.

trace/opencensus/trace/ext/django/middleware.py
liyanhui1228: I prefer a broad except here to prevent unexpected errors which might fail the user's application.

trace/opencensus/trace/ext/django/middleware.py
duggelz: Also, I just ran across this PEP, literally created last week: https://www.python.org/dev/peps/pep-0550/

trace/opencensus/trace/ext/django/config.py
```diff
@@ -0,0 +1,66 @@
...
+    """Set the params for a django tracer, including the tracer, sampler and
+    reporter. If the user did not define the settings in django settings file.
+    then use the dafaults.
+    """
+
+    def __init__(self, settings=DEFAULT_DJANGO_TRACER_CONFIG):
+        # Try to read the user settings from django settings file
+        user_settings = getattr(django_settings, 'OPENCENSUS_TRACE', None)
+
+        if user_settings is not None:
```
duggelz: Could simplify this to `self.settings = getattr(django_settings, 'OPENCENSUS_TRACE', user_settings)`

trace/tests/unit/ext/django/app/settings.py
```diff
@@ -0,0 +1,92 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Django settings for test app."""
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import os
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+SECRET_KEY = '-c&qt=71oi^e5s8(ene*$b89^#%*0xeve$x_trs91veok9#0h0'
```
duggelz: Even though this should be ok, because it's a test app, it would make me feel much better to even avoid the appearance on insecurity.  Maybe `SECRET_KEY = str(uuid.uuid4.hex())` or something?

PR #20, opened 2017-08-23
Update the license to OpenCensus Authors

liyanhui1228:


trace/nox.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc.
```
bogdandrutu: For 2016 use 2016-17

liyanhui1228: Done.

PR #21, opened 2017-08-24
Flask integration to trace requests automatically

liyanhui1228:
Sample usage is like below:
```
from opencensus.trace.ext.flask.flask_middleware import FlaskMiddleware

app = flask.Flask(__name__)
middleware = FlaskMiddleware(app)
```


trace/opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -63,7 +63,7 @@ def from_header(header):
         enabled = bool(int(options) | _ENABLED_BITMASK)
 
         span_context = SpanContext(
-            trace_id=trace_id,
+            trace_id=str(trace_id),
```
duggelz: Why is this necessary?

trace/opencensus/trace/trace_span.py
```diff
@@ -76,21 +76,24 @@ def __init__(
...
         self.parent_span_id = parent_span_id
-        self.labels = labels
         self.start_time = start_time
         self.end_time = end_time
 
         if span_id is None:
             span_id = generate_span_id()
 
+        if labels is None:
+            labels = {}
```
duggelz: Good catch

PR #27, opened 2017-09-08
Add mysql integration

liyanhui1228:


trace/opencensus/trace/ext/mysql/trace.py
```diff
@@ -0,0 +1,101 @@
...
+def trace():
+    """Wrap the mysql connector to trace it."""
+    log.info('Integrated module: {}'.format(MODULE_NAME))
+    conn_func = mysql.connector.connect
+    conn_module = inspect.getmodule(conn_func)
+    wrapped = wrap_conn(conn_func)
+    setattr(conn_module, conn_func.__name__, wrapped)
+
+
+def wrap_conn(conn_func):
```
duggelz: @jonparrott Do you have any particular opinions on wrapping, e.g., MySQL or other third-party objects with wrappers that call tracing functions?

trace/opencensus/trace/ext/mysql/trace.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(self, cursor):
+        super(TraceCursor, self).__init__(cursor)
+        self.cursor = cursor
+        self.tracer = thread_local.get_opencensus_tracer()
+
+    def trace_cursor_query(self, method, query, *args, **kwargs):
+        self.tracer.start_span()
+        self.tracer.add_label_to_current_span('mysql.query', query)
+        self.tracer.add_label_to_current_span(
+            'cursor.method.name',
```
duggelz: The private methods we add should probably use double underscore, i.e. `__trace_cursor_query` to avoid collisions (unlikely though they are).  Also, I'd prefer `mysql.cursor.method.name` just to keep all the mysql labels together alphabetically.

trace/opencensus/trace/ext/mysql/trace.py
```diff
@@ -0,0 +1,101 @@
...
+            method.__name__)
+
+        result = method(*args, **kwargs)
```
duggelz: This should be `result = method(query, *args, **kwargs)`?  And the `query, query` in the function below should just be one `query`?

trace/opencensus/trace/thread_local.py
```diff
@@ -0,0 +1,40 @@
...
+    """Get the opencensus tracer from thread local."""
+    return getattr(_thread_local, 'tracer', noop_tracer.NoopTracer())
+
+
+def set_opencensus_tracer(tracer):
+    """Add the tracer to thread local."""
+    setattr(_thread_local, 'tracer', tracer)
+
+
+def set_opencensus_attrs(attr_key, attr_value):
```
duggelz: Can you make this singular, i.e. `set_opencensus_attr`?  Also the get function below?

trace/opencensus/trace/thread_local.py
```diff
@@ -0,0 +1,40 @@
...
+    attrs = get_opencensus_attrs()
+    attrs[attr_key] = attr_value
+
+    setattr(_thread_local, 'attrs', attrs)
+
+
+def get_opencensus_attrs():
+    return getattr(_thread_local, 'attrs', {})
```
duggelz: Probably `None` is a better default, to match the behavior of similar functions in Python like get().

trace/opencensus/trace/tracer/base.py
```diff
@@ -81,3 +84,17 @@ def add_label_to_spans(self, label_key, label_value):
     def list_collected_spans(self):
         """List collected spans."""
         raise NotImplementedError
+
+
+class NullObject(object):
```
duggelz: Can we call this `NullContextManager`?  `NullObject` is a little non-specific.

trace/setup.py
```diff
@@ -20,6 +20,7 @@
 install_requires = [
     'google-gax>=0.15.7, <0.16dev',
     'googleapis-common-protos[grpc]>=1.5.2, <2.0dev',
+    'wrapt',
```
duggelz: This is probably ok, but I want to make sure that we only add external dependencies rarely, when they are really needed.  Although it's been around for 3 years and has had several releases, I'm concerned that wrapt only has one maintainer.  It seems like we could use regular decorators and subclasses and get most of the way there.

trace/tests/unit/ext/flask/app/main.py
```diff
@@ -13,16 +13,36 @@
...
 
 
 @app.route('/')
 def hello():
     return 'hello'
 
 
+@app.route('/mysql')
+def query():
+    conn = mysql.connector.connect(user='root', password='19931228')
```
duggelz: Does this create an in-memory database, or connect to an external database, or do nothing?  Can you add a comment?

If we're adding a dependency that tests require a running mysql database, we'll need to do some work around that.

trace/opencensus/trace/config_integration.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+import logging
+
+log = logging.getLogger(__name__)
+
+DEFAULT_INTEGRATIONS = {
+    'mysql': True,
```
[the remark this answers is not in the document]
liyanhui1228: Framework integration should be separated from this. Because the usage is different, user needs to use the middleware to config the framework integrations.

trace/opencensus/trace/config_integration.py
```diff
@@ -0,0 +1,54 @@
...
+    'redis': False,
+}
+
+PATH_PREFIX = 'opencensus.trace.ext'
+
+
+def trace_integrations(integrations=None):
+    """Enable tracing on the selected integrations.
+
+    :type integrations: dict
```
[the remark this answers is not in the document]
liyanhui1228: I changed it to list, the values are the name of the services to integrate, like 'mysql'.

trace/opencensus/trace/config_integration.py
```diff
@@ -0,0 +1,54 @@
...
+    :param integrations: The integrations to be traced.
+    """
+    integrated = []
+
+    if integrations is None:
+        integrations = DEFAULT_INTEGRATIONS
```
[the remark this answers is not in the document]
liyanhui1228: Agree. Changed to let the user specify what to integrate with.

trace/opencensus/trace/config_integration.py
```diff
@@ -0,0 +1,54 @@
...
+
+    for item in integrations.keys():
+        if integrations.get(item):
+            try:
+                path_to_module = '{}.{}.trace'.format(PATH_PREFIX, item)
+                module = importlib.import_module(path_to_module)
+                module.trace()
+                integrated.append(item)
+            except Exception:
+                log.debug(
```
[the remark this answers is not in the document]
liyanhui1228: Fixed.

trace/opencensus/trace/thread_local.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2017, OpenCensus Authors
```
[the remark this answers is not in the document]
liyanhui1228: Changed.

trace/opencensus/trace/thread_local.py
liyanhui1228: Done.

trace/opencensus/trace/thread_local.py
liyanhui1228: Done.

trace/opencensus/trace/ext/mysql/trace.py
liyanhui1228: Done.

trace/opencensus/trace/ext/mysql/trace.py
liyanhui1228: Good call, thanks!

trace/opencensus/trace/tracer/base.py
liyanhui1228: Done.

trace/setup.py
liyanhui1228: Removed the wrapt dependency, using subclasses works right now.

trace/tests/unit/ext/flask/app/main.py
liyanhui1228: I haven't setup circle ci to run the end-to-end test yet, I just run this application locally to test. So it's fine right now. The database is pre-installed and configured in my local machine.

liyanhui1228:
@duggelz Ready for another review, thanks!

trace/opencensus/trace/ext/mysql/trace.py
```diff
@@ -0,0 +1,89 @@
...
+            _span.name = '[mysql.query]{}'.format(query)
+            _tracer.add_label_to_current_span('mysql/query', query)
+            _tracer.add_label_to_current_span(
+                'mysql/cursor/method/name',
+                query_func.__name__)
+
+            result = query_func(query, *args, **kwargs)
+
+            _tracer.end_span()
+        except Exception:  # pragma: NO COVER
```
duggelz: What exceptions can be raised here?  I feel like this try block might just hide bugs, and retrying transactions is going to very surprising to users.

liyanhui1228: Thanks for pointing out! No specific exception can be raised here, and I removed the try catch block.

PR #28, opened 2017-09-16
Add postgreSQL integration

liyanhui1228:


trace/nox.py
```diff
@@ -28,7 +28,8 @@ def unit_tests(session, python_version):
     # Install all test dependencies, then install this package in-place.
     session.install('mock', 'pytest', 'pytest-cov',
                     'django', 'flask', 'webapp2', 'webob',
-                    'google-cloud-trace', 'mysql-connector==2.1.6')
+                    'google-cloud-trace', 'mysql-connector==2.1.6',
```
duggelz: Can we move these into a `requirements-test.txt` file?

trace/tests/unit/ext/django/app/views.py
```diff
@@ -71,6 +74,29 @@ def mysql_trace(request):
         return HttpResponse(msg)
 
 
+def postgresql_trace(request):
+    try:
+        conn = psycopg2.connect(
+            host='localhost',
+            user='postgres',
```
duggelz: `POSTGRES_USER`?

trace/tests/unit/ext/django/app/views.py
```diff
@@ -71,6 +74,29 @@ def mysql_trace(request):
...
+
+        result = []
+
+        for item in cursor.fetchall():
+            result.append(item)
+
+        return HttpResponse(str(result))
+
+    except Exception:
+        return "Query failed. Check your env vars for connection settings."
```
duggelz: This should return a non-200 HTTP status code.  Probably 500.

trace/tests/unit/ext/flask/app/main.py
```diff
@@ -38,10 +40,9 @@ def hello():
 
 
 @app.route('/mysql')
-def query():
-
+def mysql_query():
     try:
-        conn = mysql.connector.connect(user=USER, password=PASSWORD)
+        conn = mysql.connector.connect(user='root', password='19931228')
```
duggelz: Are these values left over from debugging?

trace/tests/unit/ext/flask/app/main.py
```diff
@@ -58,5 +59,29 @@ def query():
         return "Query failed. Check your env vars for connection settings."
 
 
+@app.route('/postgresql')
+def postgresql_query():
+    try:
+        conn = psycopg2.connect(
+            host='localhost',
+            user='postgres',
```
duggelz: We should probably have a single file shared between the flask and django tests with the username, password, dbname, hostname, and sample queries and expected results.

trace/nox.py
liyanhui1228: Done.

trace/tests/unit/ext/django/app/views.py
liyanhui1228: Moved to config file.

trace/tests/unit/ext/django/app/views.py
liyanhui1228: Done.

trace/tests/unit/ext/flask/app/main.py
liyanhui1228: Fixed.

trace/tests/unit/ext/flask/app/main.py
liyanhui1228: Done.

trace/requirements.txt
```diff
@@ -1 +1,10 @@
-google-gax>=0.15.7, <0.16dev
+Django==1.11.5
```
duggelz: Can you move this to `requirements-test.txt`, and leave the existing `requirements.txt` as is?  We want to separate the dependencies a customer needs to install this (just google-gax), from the dependencies needed to test this (everything).

trace/tests/unit/ext/django/app/views.py
```diff
@@ -20,12 +20,17 @@
 from opencensus.trace import config_integration
 
 import mysql.connector
+import psycopg2
+
 import time
 import os
+import sys
+
+sys.path.insert(0, os.path.abspath(__file__+"/../../../.."))
```
duggelz: We really, really, want to avoid modifying `sys.path` at runtime if at all possible.  What is the underlying issue?

trace/requirements.txt
liyanhui1228: Done.

trace/tests/unit/ext/django/app/views.py
liyanhui1228: Leave it for now as discussed.

PR #31, opened 2017-09-26
Add requests integration

liyanhui1228:


trace/opencensus/trace/ext/requests/trace.py
```diff
@@ -0,0 +1,61 @@
...
+    def call(url, *args, **kwargs):
+        _tracer = execution_context.get_opencensus_tracer()
+        _span = _tracer.start_span()
+        _span.name = '[requests]{}'.format(requests_func.__name__)
+
+        # Add the requests url to labels
+        _tracer.add_label_to_current_span('requests/url', url)
+
+        # Add the kwargs to labels
+        for key, value in kwargs.items():
```
duggelz: This might be tracing too much.  What do the kwargs look like?  We don't want to trace the whole request body for example, and tracing all the request headers may record data which should be ephemeral and/or protected, like passwords.

liyanhui1228: Thanks for pointing out. The kwargs can include the url parameters or request headers, and users might not want to expose them in the spans. I removed the code for tracing kwargs.

PR #32, opened 2017-09-28
Make tracer params configurable in django middleware

liyanhui1228:
Get the params like sampling rate from the django settings file, and use them to initialize samplers and reporters.

### DIFF
--- a/trace/opencensus/trace/ext/django/middleware.py
+++ b/trace/opencensus/trace/ext/django/middleware.py
@@ -20,6 +20,7 @@ from opencensus.trace.ext.django.config import settings
 from opencensus.trace import labels_helper
 from opencensus.trace import request_tracer
 from opencensus.trace import execution_context
+from opencensus.trace.samplers import fixed_rate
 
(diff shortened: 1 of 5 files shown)


trace/opencensus/trace/ext/django/middleware.py
```diff
@@ -87,6 +94,35 @@ def __init__(self, get_response=None):
         self._reporter = settings.REPORTER
         self._propagator = settings.PROPAGATOR
 
+        # Initialize the sampler
+        if self._sampler.__name__ == 'FixedRateSampler':
```
duggelz: Are we going to have very similar code for Flask, webapp2, etc?  It seems like it might be better if the sampler could figure this out itself.  Lookups based on __name__ tend to be fragile.  Could we have a generic (not Django-specific) settings object that we pass into the sampler and report constructors that does this lookup?

liyanhui1228: This is just for Django, because we have to do the tracing configuration in the Django middleware. Need to think about better ways to do the config though.

PR #34, opened 2017-10-11
System test

liyanhui1228:


.circleci/config.yml
```diff
@@ -8,12 +8,37 @@ jobs:
...
+                  -in trace/tests/system/credentials.json.enc \
+                  -out trace/$GOOGLE_APPLICATION_CREDENTIALS
+            else
+              echo "No credentials. System tests will not run."
+            fi
+      - run:
+          name: Install packages
+          command: |
+            apt-get update
+            apt-get install netcat-openbsd
```
duggelz: What's this for?

trace/nox.py
```diff
@@ -39,12 +40,43 @@ def unit_tests(session, python_version):
...
         '--cov-report=',
         '--cov-fail-under=97',
-        'tests/',
+        'tests/unit/',
         *session.posargs
     )
 
 
 @nox.session
+@nox.parametrize('python_version', ['2.7', '3.5'])
```
duggelz: Can we change this from 3.5 to 3.6?

.circleci/config.yml
duggelz: What's this for?

trace/nox.py
```diff
@@ -39,12 +40,43 @@ def unit_tests(session, python_version):
...
+    session.install('.')
+
+    # Run py.test against the system tests.
+    session.run(
+        'py.test',
+        '-vvv',
+        '-s',
+        'tests/system/',
+        *session.posargs,
+        success_codes=range(0, 100)
```
duggelz: What does this do?  It seems like it ignores all error codes in the [1,99] range?  Can we reduce this list, and document the expected error codes?

trace/tests/system/set_credential.sh
```diff
@@ -0,0 +1,26 @@
...
+# Set credential env var
+export GOOGLE_APPLICATION_CREDENTIALS='google_application_credentials'
+
+# Decrypt credentials
+if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
+    openssl aes-256-cbc -d -a -k "$GOOGLE_CREDENTIALS_PASSPHRASE" \
+        -in credentials.json.enc \
+        -out "$GOOGLE_APPLICATION_CREDENTIALS"
+else
+    echo "No credentials. System tests will not run."
```
duggelz: Can you add `exit 1` here?

trace/tests/system/flask/main.py
```diff
@@ -24,15 +24,22 @@
...
+
+# MySQL settings
+MYSQL_PASSWORD = os.environ.get('SYSTEST_MYSQL_PASSWORD')
+
+# PostgreSQL settings
+POSTGRES_PASSWORD = os.environ.get('SYSTEST_POSTGRES_PASSWORD')
+
 app = flask.Flask(__name__)
 
 # Enbale tracing, send traces to Stackdriver Trace
```
duggelz: Enbale -> Enable

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+import uuid
+
+import unittest
+
+PROJECT = os.environ.get('GCLOUD_PROJECT_PYTHON')
+
+
+def wait_app_to_start():
+    """Wait the application to start running."""
+    cmd = 'until nc -z -v -w30 127.0.0.1 8080; do sleep 2; done'
```
duggelz: Is there a way to make this fail after a certain timeout?  It looks like it would loop forever?

I had used `wget --retry-connrefused --tries=5 http://localhost:$PORT/` elsewhere, which returns with an error code after 5 tries (~60 seconds I think).

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+        self.assertEqual(trace.get('projectId'), PROJECT)
+        self.assertEqual(trace.get('traceId'), str(self.trace_id))
+        self.assertEqual(len(trace.get('spans')), 1)
+        self.assertEqual(
+            trace.get('spans')[0].get('parentSpanId'),
+            str(self.span_id))
+
+    def test_mysql_trace(self):
+        requests.get(
+            'http://127.0.0.1:8080/mysql',
```
duggelz: At some point, we'll need to change all the 127.0.0.1 to `localhost` for IPv6 compatibility.

liyanhui1228:
@duggelz I couldn't see your comment on the UI, the `netcat-openbsd` is for checking the connection to the flask/django app and wait for it to fully start and ready to response to requests.

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+
+import unittest
+
+PROJECT = os.environ.get('GCLOUD_PROJECT_PYTHON')
+
+
+def wait_app_to_start():
+    """Wait the application to start running."""
+    cmd = 'until nc -z -v -w30 127.0.0.1 8080; do sleep 2; done'
+    os.system(cmd)
```
duggelz: I'd prefer `subprocess.check_call(cmd)` which raises an exception if the command fails.

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+    def tearDown(self):
+        # Kill the flask application process
+        os.killpg(os.getpgid(self.process.pid), signal.SIGTERM)
+
+    def test_flask_request_trace(self):
+        requests.get(
+            'http://127.0.0.1:8080',
+            headers=self.headers_trace)
+
+        time.sleep(5)
```
duggelz: We'll want to go back later and change the hardcoded sleep times to use `retry` or similar, but this is ok for now.

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.assertEqual(trace.get('projectId'), PROJECT)
+        self.assertEqual(trace.get('traceId'), str(self.trace_id))
+
+        # Should have 2 spans, one for flask request, one for mysql query
+        self.assertEqual(len(trace.get('spans')), 2)
+        self.assertEqual(
+            trace.get('spans')[0].get('parentSpanId'),
+            str(self.span_id))
+
```
duggelz: Can you check the span attributes also?

trace/tests/system/flask/Dockerfile
```diff
@@ -12,17 +12,15 @@
...
-POSTGRES_USER = os.environ.get('POSTGRES_USER')
-POSTGRES_HOST = os.environ.get('POSTGRES_HOST')
-POSTGRES_DB = os.environ.get('POSTGRES_DB')
+# Set virtualenv environment variables. This is equivalent to running
+# source /env/bin/activate
+ENV VIRTUAL_ENV /env
+ENV PATH /env/bin:$PATH
+ADD . /app/
+RUN pip install -r requirements.txt
+EXPOSE 8000
```
duggelz: What's running on port 8000, and what's running on port 8080?  Is there an inconsistency with some of the other code below?

trace/tests/system/flask/flask_system_test.py
```diff
@@ -0,0 +1,170 @@
...
+    cmd = 'python tests/system/flask/main.py'
+    process = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        shell=True,
+        preexec_fn=os.setsid)
+    return process
+
+
+class TestFlaskTrace(unittest.TestCase):
```
duggelz: If we could separate the "test harness" from the "start flask application", we could reduce some of the duplicated code between the django and flask tests.

trace/tests/system/basic_trace/basic_trace_system_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+        with tracer.span(name='root_span') as root:
+            func_to_trace()
+            parent_span_id = root.span_id
+            with root.span(name='child_span'):
+                func_to_trace()
+
+        tracer.end_trace()
+
+        file = open(file_reporter.DEFAULT_FILENAME, 'r')
```
duggelz: If we later make tracing asynchronous (which we should), we'll have to remember to add some kind of wait or flush here, between closing the last span and opening the reporter file.

.circleci/config.yml
liyanhui1228: See comment above.

trace/nox.py
liyanhui1228: Done.

trace/nox.py
liyanhui1228: Removed. Will add documentation after several observations of the system tests failure(if any).

trace/tests/system/basic_trace/basic_trace_system_test.py
liyanhui1228: Yes, will do that.

trace/tests/system/flask/Dockerfile
liyanhui1228: Thanks for pointing out! Actually I'm going to remove this dockerfile since it's not used in the system test, just for running locally.

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Thanks for the suggestion! Changed.

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Changed.

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Yes, will use retry in later PR.

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Changed.

trace/tests/system/flask/main.py
liyanhui1228: Fixed.

trace/tests/system/set_credential.sh
liyanhui1228: Added

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Yes, will fix in separate PR.

trace/tests/system/flask/flask_system_test.py
liyanhui1228: Added.

PR #39, opened 2017-10-20
Remove trace class

### LINKED ISSUE #37: Remove Trace class
This library only exports spans, trace should not be used internally.

liyanhui1228:
Closes #37.


trace/opencensus/trace/utils.py
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import uuid
+
+
+def generate_trace_id():
```
duggelz: Would it make sense to put generate_trace_id() and generate_span_id() in span_context.py?  Related, from the discussion yesterday, should we remove span_id from Span, and have Span contain a SpanContext?

liyanhui1228: Moved to `span_context.py`. Yes will move SpanContext to Span in later PR.

PR #48, opened 2017-10-25
Change sampling rule of FixedRateSampler

### LINKED ISSUE #45: Change FixedRateSampler sampling rule

liyanhui1228:
Closes #45

### DIFF
--- a/trace/opencensus/trace/samplers/fixed_rate.py
+++ b/trace/opencensus/trace/samplers/fixed_rate.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import random
-
 from opencensus.trace.samplers.base import Sampler
 
 DEFAULT_SAMPLING_RATE = 0.5
+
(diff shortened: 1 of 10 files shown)


trace/opencensus/trace/samplers/fixed_rate.py
```diff
@@ -12,30 +12,46 @@
...
 
-    @property
-    def should_sample(self):
-        random_number = random.uniform(0, 1)
+    def should_sample(self, trace_id):
+        lower_long = get_lower_long_from_trace_id(trace_id)
+        bound = self.rate * MAX_VALUE
 
-        if random_number <= self.rate:
+        if lower_long <= bound:
```
duggelz: Do we know that this the same logic that Java et al uses?  Has this been documented in the specs repo yet?

liyanhui1228: Yes, but as there is no max value for long type in Python, I use the max value for a 8 bytes hex as the max value.

PR #52, opened 2017-10-30
Remove span_stack and store parent_span in span object

### LINKED ISSUE #51: Add current span to execution context

liyanhui1228:
Add get/set_current_span methods in execution_context, use the current span as the parent span when creating new spans. Remove span_stack, and instead use the parent_span object in span to record the relationship between spans.

Closes #51

### DIFF
--- a/trace/opencensus/trace/span.py
+++ b/trace/opencensus/trace/span.py
@@ -37,8 +37,8 @@ class Span(object):
                  distinguished using RPC_CLIENT and RPC_SERVER to identify
                  queueing latency associated with the span.
 
-    :type parent_span_id: int
-    :param parent_span_id: (Optional) ID of the parent span.
+    :type parent_span: :class:`~opencensus.trace.span.Span`
+    :param parent_span: (Optional) Parent span.
 
     :type labels: dict
     :param labels: Collection of labels associated with the span.
@@ -70,7 +70,7 @@ class Span(object):
             self,
             name,
             kind=Enum.SpanKind.SPAN_KIND_UNSPECIFIED,
-            parent_span_id=None,
+            parent_span=None,
             labels=None,
             start_time=None,
             end_time=None,
@@ -78,7 +78,7 @@ class Span(object):
             context_tracer=None):
         self.name = name
         self.kind = kind
-        self.parent_span_id = parent_span_id
+        self.parent_span = parent_span
         self.start_time = start_time
         self.end_time = end_time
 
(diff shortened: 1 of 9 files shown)


trace/opencensus/trace/tracer/context_tracer.py
```diff
@@ -76,14 +74,19 @@ def start_span(self, name='span'):
         :rtype: :class:`~opencensus.trace.span.Span`
         :returns: The Span object.
         """
-        parent_span_id = self.span_context.span_id
+        parent_span = self.current_span()
+
+        if parent_span is None:
```
duggelz: Can you add a comment describing the django case you explained to me?

trace/opencensus/trace/tracer/context_tracer.py
```diff
@@ -94,23 +97,23 @@ def end_span(self):
         """
         cur_span = self.current_span()
 
-        self._span_stack.pop()
+        if cur_span is None:
+            logging.warning('No active span, cannot do end_span.')
+            return
+
         cur_span.finish()
+        self.span_context.span_id = cur_span.parent_span.span_id
```
duggelz: Is it valid to manually create a span like `Span(parent_span=None,...)`?  In that case, we would need to check for None here, and set `span_id` to 0 or something.  It seems like we have 3 cases:

1) parent_span is a Span object
2) parent_span is a NullContextManager because we got the span_id from an HTTP header
3) parent_span is None because we manually created a root span.

trace/opencensus/trace/tracer/context_tracer.py
liyanhui1228: Done.

trace/opencensus/trace/span.py
```diff
@@ -88,6 +89,11 @@ def __init__(
         if labels is None:
             labels = {}
 
+        # Do not manipulate spans directly using the methods in Span Class,
+        # make usre to use the RequestTracer.
```
duggelz: usre -> sure

trace/opencensus/trace/tracer/context_tracer.py
liyanhui1228: Yes we have these 3 cases, I added check for parent_span in span constructor and documented that user should start and end span using RequestTracer instead of directly using the span class.

trace/opencensus/trace/span.py
liyanhui1228: Fixed.

PR #53, opened 2017-10-31
Build docs using sphinx

liyanhui1228:
The empty .nojekyll file is for letting github pages render the css configurations.


scripts/update_docs.sh
```diff
@@ -0,0 +1,26 @@
...
+# Build docs
+function build_docs {
+    rm -rf docs/build/
+    make html
+    return $?
+}
+
+build_docs
+
+cp -R docs/build/html/* docs/
```
duggelz: Maybe delete `docs/build` after the copy?

make.bat
```diff
@@ -0,0 +1,36 @@
+@ECHO OFF
```
duggelz: Is this just copied from the quickstart?  I'd rather not include it if we haven't tried it.  We can just say that one must build the docs on Linux, not Windows.

scripts/update_docs.sh
liyanhui1228: Deleted.

make.bat
liyanhui1228: Yes this is auto generated. Removed.

PR #63, opened 2017-11-03
Batch exporting spans once a while

### LINKED ISSUE #59: Exporters should batch
The GoogleCloudExporter should batching multiple traces and upload them to the backend once a while. The batching options could be configurable.

See the bundler utility in Go for an example list of batching/buffering options: https://godoc.org/google.golang.org/api/support/bundler#pkg-constants.

liyanhui1228:
Closes #59
This PR is doing the following things:

- [x] Add background transport module for batch exporting spans, one thread per exporter.
- [x] Add sync transport module for exporting span once it's finished


trace/opencensus/trace/exporters/file_exporter.py
```diff
@@ -27,14 +28,19 @@ class FileExporter(base.Exporter):
...
     """
 
-    def __init__(self, file_name=DEFAULT_FILENAME):
+    def __init__(self, file_name=DEFAULT_FILENAME,
+                 transport=sync.SyncTransport):
         self.file_name = file_name
+        self.transport = transport(self)
 
-    def export(self, trace):
+    def emit(self, trace):
```
duggelz: Can you add emit() to base.Exporter, and add doc comments explaining emit() and export()?

The naming is a little weird, since I don't usually think of "transport" for something that's just printing or appending to a file.  But I don't have a better suggestion.

trace/opencensus/trace/exporters/transports/background_thread.py
```diff
@@ -0,0 +1,146 @@
...
+
+_DEFAULT_GRACE_PERIOD = 5.0  # Seconds
+_WAIT_PERIOD = 3.0  # Seconds
+_DEFAULT_MAX_BATCH_SIZE = 2
+_WORKER_THREAD_NAME = 'opencensus.trace.Worker'
+_WORKER_TERMINATOR = object()
+
+
+class _Worker(object):
+    def __init__(self, exporter, grace_period=_DEFAULT_GRACE_PERIOD,
```
duggelz: The google-cloud-python code which inspired this is well documented, e.g. https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/logging/google/cloud/logging/handlers/transports/background_thread.py#L65  .  Can you write/copy doc comments into this code to match?

It would be nicest if we could both use the same code, but I don't think that's feasible right now.

trace/opencensus/trace/exporters/zipkin_exporter.py
```diff
@@ -62,12 +63,14 @@ def __init__(
             service_name='my_service',
             host_name=DEFAULT_HOST_NAME,
             port=DEFAULT_PORT,
-            endpoint=DEFAULT_ENDPOINT):
+            endpoint=DEFAULT_ENDPOINT,
+            transport=sync.SyncTransport):
```
duggelz: Can you update the doc comment above?

trace/tests/system/flask/main.py
```diff
@@ -38,8 +38,11 @@
 
 app = flask.Flask(__name__)
 
-# Enable tracing, send traces to Stackdriver Trace
-exporter = stackdriver_exporter.StackdriverExporter()
+# Enable tracing, send traces to Stackdriver Trace using background thread transport
+from opencensus.trace.exporters.transports import background_thread
+exporter = stackdriver_exporter.StackdriverExporter(
+    transport=background_thread.BackgroundThreadTransport)
+
```
duggelz: To check my understanding, the flask test uses async transport and the django test uses sync transport?  I think this is a good idea, but we should add a comment here and in the django test explaining that they are intentionally different to test both code paths.

liyanhui1228: Yes correct. Comment added.

trace/opencensus/trace/exporters/zipkin_exporter.py
liyanhui1228: Done.

trace/opencensus/trace/exporters/transports/background_thread.py
liyanhui1228: Done.

trace/opencensus/trace/exporters/file_exporter.py
liyanhui1228: Done.

PR #68, opened 2017-11-09
Export span once it's finished

### LINKED ISSUE #65: Send span to exporter once the span is ended
For resolving #63. Currently the spans are stored in spans_list and sent to exporter all together when calling tracer.finish(). Need to change the implementation to export span once the span is ended to support the batch exporting.

liyanhui1228:
Closes #65

### DIFF
--- a/trace/opencensus/trace/ext/flask/flask_middleware.py
+++ b/trace/opencensus/trace/ext/flask/flask_middleware.py
@@ -49,7 +49,8 @@ class FlaskMiddleware(object):
                      :class:`.PrintExporter`. The rest option is
                      :class:`.FileExporter`.
     """
-    def __init__(self, app, sampler=None, exporter=None, propagator=None):
+    def __init__(self, app, sampler=None, exporter=None, propagator=None,
+                 transport=None):
         if sampler is None:
             sampler = always_on.AlwaysOnSampler()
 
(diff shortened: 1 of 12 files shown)


trace/opencensus/trace/ext/flask/flask_middleware.py
```diff
@@ -49,7 +49,8 @@ class FlaskMiddleware(object):
                      :class:`.PrintExporter`. The rest option is
                      :class:`.FileExporter`.
     """
-    def __init__(self, app, sampler=None, exporter=None, propagator=None):
+    def __init__(self, app, sampler=None, exporter=None, propagator=None,
+                 transport=None):
```
duggelz: I think you didn't mean to have `transport` in this PR?

trace/opencensus/trace/tracer/context_tracer.py
```diff
@@ -59,6 +61,7 @@ def span(self, name='span'):
         :returns: The Span object.
         """
         span = self.start_span(name=name)
+        span.__class__.__exit__ = self.end_span
```
duggelz: I don't quite understand this.  Why are we modifying `span.__class__` at runtime?  Doesn't the existing `Span.__exit__()` call `ContextTracer.end_span()`?

Also, in `Span.__exit__`, I wonder if we should be calling `self.finish()` before calling `end_span()`, instead of after?

trace/opencensus/trace/ext/flask/flask_middleware.py
liyanhui1228: Removed.

trace/opencensus/trace/tracer/context_tracer.py
liyanhui1228: I was removing the `ContextTracer` from span class and modifying the span.__class__.__exit__ to use the `ContextTracer.end_span()` instead. Changed back to the previous implementation as it's not that safe to modify the span.__class__.
And for `span.__exit__`, if it's calling `ContextTracer.end_span()`, the `span.finish()` is called inside `tracer.end_span()` method and then return, the finish function won't be executed twice.

PR #71, opened 2017-11-14
Add tracing blacklist

### LINKED ISSUE #70: Implement path based blacklist to config tracing on specific path
Use case: On GAE Flex, the default setting for health check is on, and user's application will receive a lot of health check requests which would be automatically traced and thus the Trace UI will show a lot of health check spans. We need a way to enable the users to config what path they do not want to trace.

liyanhui1228:
Fixes #70 

- [x] Flask config
- [x] Django config

### DIFF
--- a/trace/opencensus/trace/ext/django/config.py
+++ b/trace/opencensus/trace/ext/django/config.py
@@ -31,6 +31,7 @@ DEFAULT_DJANGO_TRACER_PARAMS = {
     'ZIPKIN_EXPORTER_SERVICE_NAME': 'my_service',
     'ZIPKIN_EXPORTER_HOST_NAME': 'localhost',
     'ZIPKIN_EXPORTER_PORT': 9411,
+    'BLACKLIST_PATHS': ['/_ah/health'],
 }
 
 
--- a/trace/opencensus/trace/ext/django/config.py
+++ b/trace/opencensus/trace/ext/django/config.py
@@ -26,12 +26,15 @@ DEFAULT_DJANGO_TRACER_CONFIG = {
 }
 
 DEFAULT_DJANGO_TRACER_PARAMS = {
-    'SAMPLING_RATE': 0.5,
+    # https://cloud.google.com/appengine/docs/flexible/python/
+    # how-instances-are-managed#health_checking
+    'BLACKLIST_PATHS': ['/_ah/health'],
     'GCP_EXPORTER_PROJECT': None,
+    'SAMPLING_RATE': 0.5,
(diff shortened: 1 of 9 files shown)


trace/opencensus/trace/ext/django/config.py
```diff
@@ -31,6 +31,7 @@
     'ZIPKIN_EXPORTER_SERVICE_NAME': 'my_service',
     'ZIPKIN_EXPORTER_HOST_NAME': 'localhost',
     'ZIPKIN_EXPORTER_PORT': 9411,
+    'BLACKLIST_PATHS': ['/_ah/health'],
```
duggelz: These should probably be in alphabetical order.  Also, can you add a comment next to the url with a link to https://cloud.google.com/appengine/docs/flexible/python/how-instances-are-managed#health_checking .

duggelz: Also, we should document that these are literal string matches, rather than regular expressions or something like that.

trace/opencensus/trace/ext/utils.py
```diff
@@ -23,3 +29,20 @@ def get_func_name(func):
...
+    the health check request. True if not tracing, False if tracing.
+
+    :type blacklist_paths: list
+    :param blacklist_paths: Paths that not tracing.
+    """
+    if blacklist_paths is None:
+        blacklist_paths = DEFAULT_BLACKLIST_PATHS
+
+    for path in blacklist_paths:
+        if path in url:
```
duggelz: I worry a bit about false positives here, like the url is `foo.com/something?blahblahblah&arg=_/ah/health`.  What would you think about requiring the path to be at the start of the URL, or be a regular expression that matches the entire URL?  You'd have to do a little more work to split the hostname/port from the path from the fragment/query parameters.

trace/opencensus/trace/ext/django/config.py
liyanhui1228: Done.

trace/opencensus/trace/ext/utils.py
liyanhui1228: Done. Split the url path from the provided url, and see whether the url path starts with the blacklisted path or not.

jtwarren:
@liyanhui1228 is there any way to blacklist `/` but not `/everything`?

The use case is on GCP (GKE) with GLB hitting `/` as a health check. Currently it's not possible, afaik, to configure a different healthcheck path as the ingress controller does not support HTTPS readiness probes (https://github.com/kubernetes/ingress-gce/blob/master/examples/health-checks/README.md) which we're using.

PR #83, opened 2017-11-22
Add support for configure transport in Django middleware

### LINKED ISSUE #82: Add support for configuring transport in Django settings
Currently there is no way for users to set the transport type(sync or async) in Django settings, we need to provide support for them to choose the transport type.

liyanhui1228:
Closes #82


trace/opencensus/trace/ext/django/config.py
```diff
@@ -74,6 +81,20 @@ def __getattr__(self, attr):
...
 
+def _set_default_configs(user_settings, default):
+    """Set the default value to user settings if user not specified
+    the value.
+    """
+    config_keys = user_settings.keys()
+    supported_keys = default.keys()
+
+    for key in supported_keys:
+        if key not in config_keys:
```
chingor13: This could probably be `if key not in user_settings` to check if the key exists rather than checking if the value is found in the list of keys.

liyanhui1228: Done. Thanks!

PR #86, opened 2017-11-28
Increase batch size in background thread transport

### LINKED ISSUE #85: Increase batch size in background thread transport
Increase batch size and do not wait between each export to ensure all spans are sent within the grace period when terminating the thread.

liyanhui1228:
Closes #85

### DIFF
--- a/trace/opencensus/trace/exporters/transports/background_thread.py
+++ b/trace/opencensus/trace/exporters/transports/background_thread.py
@@ -14,7 +14,6 @@
 
 import atexit
 import threading
-import time
 
 from six.moves import queue
 from six.moves import range
--- a/trace/opencensus/trace/exporters/transports/background_thread.py
+++ b/trace/opencensus/trace/exporters/transports/background_thread.py
@@ -22,8 +22,7 @@ from six.moves import range
 from opencensus.trace.exporters.transports import base
 
 _DEFAULT_GRACE_PERIOD = 5.0  # Seconds
-_WAIT_PERIOD = 3.0  # Seconds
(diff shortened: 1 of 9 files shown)


trace/opencensus/trace/exporters/transports/background_thread.py
```diff
@@ -107,14 +106,12 @@ def _thread_main(self):
...
                     'spans': spans,
                 }
+
                 self.exporter.emit(spans_json)
 
             for _ in range(len(items)):
                 self._queue.task_done()
 
-            # Wait for a while before next export
-            time.sleep(_WAIT_PERIOD)
```
chingor13: Will removing the sleep make this into a busy loop and eat CPU?

trace/opencensus/trace/exporters/transports/background_thread.py
```diff
@@ -158,11 +155,14 @@ def stop(self):
             return True
 
         with self._lock:
+            print('Number of spans pending sent: {}'.format(
```
chingor13: Is the printing meant to stay or is it just for debugging?

If intentional, should it be some kind of optional, configurable logger instead?

trace/opencensus/trace/exporters/transports/background_thread.py
liyanhui1228: Yeah agree, changed to reduce the sleep time to 1 second. Just want to make sure the spans are sent within the grace period.

trace/opencensus/trace/exporters/transports/background_thread.py
liyanhui1228: Removed.

PR #87, opened 2017-11-28
Add Stackdriver Trace V2 models

liyanhui1228:


trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,60 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def _format_attribute_value(value):
+    if type(value).__name__ == 'str':
+        value_type = 'string_value'
```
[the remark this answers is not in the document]
liyanhui1228: Good catch! Done.

liyanhui1228:
Yes will change the StackdriverExporter to use the models in a separate PR.

trace/opencensus/trace/span.py
```diff
@@ -129,6 +165,14 @@ def add_label(self, label_key, label_value):
         """
         self.labels[label_key] = label_value
 
+    def add_time_event(self, time_event):
+        """Add a TimeEvent."""
+        self.time_events.append(time_event)
```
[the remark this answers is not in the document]
liyanhui1228: Done.

trace/opencensus/trace/exporters/stackdriver_exporter.py
```diff
@@ -90,34 +90,43 @@ class StackdriverExporter(base.Exporter):
...
 
     def export(self, trace):
         self.transport.export(trace)
 
-    def translate_to_stackdriver(self, trace):
+    def translate_to_stackdriver(self, spans):
         """
-        :type trace: dict
-        :param trace: Trace collected.
+        :type spans: dict
```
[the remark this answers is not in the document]
liyanhui1228: Done.

trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def format_attributes_json(self):
+        """Convert the Attributes object to json format."""
+        attributes_json = {}
+
+        for key in self.attributes:
+            value = self.attributes.get(key)
+            value_json = _format_attribute_value(value)
+            attributes_json[key] = value_json
+
```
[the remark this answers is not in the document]
liyanhui1228: Done.

trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,63 @@
...
+                       to 128 bytes long. The value can be a string up to 256
+                       bytes, an integer, or the Boolean values true and false.
+    """
+    def __init__(self, attributes=None):
+        if attributes is None:
+            attributes = {}
+
+        self.attributes = attributes
+
+    def set_attribute(self, key, value):
```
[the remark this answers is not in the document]
liyanhui1228: Added.

trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,63 @@
...
+    :type attributes: dict
+    :param attributes: The set of attributes. Each attribute's key can be up
+                       to 128 bytes long. The value can be a string up to 256
+                       bytes, an integer, or the Boolean values true and false.
+    """
+    def __init__(self, attributes=None):
+        if attributes is None:
+            attributes = {}
+
+        self.attributes = attributes
```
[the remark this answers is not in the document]
liyanhui1228: Good catch! Done.

trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return {value_type: value}
+
+
+class Attributes(object):
+    """A set of attributes, each in the format [KEY]:[VALUE].
+
+    :type attributes: dict
+    :param attributes: The set of attributes. Each attribute's key can be up
+                       to 128 bytes long. The value can be a string up to 256
```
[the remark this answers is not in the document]
liyanhui1228: Added check for length of the attribute key. The length of string value is checked in `opencensus.trace.utils._get_truncatable_str` when converting to TruncatableString.

trace/opencensus/trace/attributes.py
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+
+from opencensus.trace.utils import _get_truncatable_str
+
+
+def _format_attribute_value(value):
+    if type(value).__name__ == 'str':
+        value_type = 'string_value'
+        value = _get_truncatable_str(value)
+    elif type(value).__name__ == 'int':
```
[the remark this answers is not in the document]
liyanhui1228: `isinstance(True, int)` will also return True. Changed to put the check for bool at the first if statement and check int later.

duggelz:
You are going to have a massive merge conflict between this and the big rename.  Which one are you going to do first?

liyanhui1228:
The Stackdriver V2 won't be included in the first release. So I'll merge the restructure PR first.

trace/opencensus/trace/utils.py
```diff
@@ -0,0 +1,67 @@
...
+# Max length is 128 bytes for a truncatable string.
+MAX_LENGTH = 128
+
+
+def _get_truncatable_str(str_to_convert):
+    """Truncate a string if exceed limit and record the truncated bytes
+    count.
+    """
+    str_bytes = str_to_convert.encode(UTF8)
+    str_len = len(str_bytes)
```
duggelz: This seems kind of problematic, because you could convert a code point to a 2 or 3 byte utf-8 sequence, and then truncate the string in the middle of that sequence, creating an invalid string that causing an exception to be thrown.

But I think you can get around this by changing the code below to

```
truncated = str_bytes.decode(UTF8, errors='ignore')
```

This seemed to do the right thing in both Python 2 and 3.  Can you add a unit test where you call this on a string with 127 ascii characters followed by a some non-ascii characters?

I used the following code to play with truncation: `('\u2603\u2603'.encode('utf-8')[0:5]).decode('utf-8', errors='ignore')`.

trace/opencensus/trace/utils.py
```diff
@@ -0,0 +1,67 @@
...
+    if limit is None:
+        limit = MAX_LENGTH
+
+    str_bytes = str_to_check.encode(UTF8)
+    str_len = len(str_bytes)
+
+    if str_len > limit:
+        str_bytes = str_bytes[:limit]
+
+    result = str_bytes.decode(UTF8)
```
duggelz: Same comment here as above.  Perhaps one of these functions could call the other, or have a helper function that takes a (string, limit) and returns (truncated_string, truncated_byte_count)?

trace/opencensus/trace/utils.py
liyanhui1228: Thanks for the explanation! Added the unit test for that.

trace/opencensus/trace/utils.py
liyanhui1228: Good suggestion, done.

trace/opencensus/trace/link.py
```diff
@@ -0,0 +1,71 @@
...
+        self.span_id = span_id
+
+        if type is None:
+            type = Type.TYPE_UNSPECIFIED
+
+        self.type = type
+        self.attributes = attributes
+
+    def format_link_json(self):
+        """Conver a Link object to json format."""
```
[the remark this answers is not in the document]
liyanhui1228: Fixed.

opencensus/trace/span_context.py
```diff
@@ -150,13 +150,13 @@ def check_trace_id(self, trace_id):
...
 def generate_span_id():
-    """Return the random generated span ID for a span.
+    """Return the random generated span ID for a span. Must be 16 digits.
 
     :rtype: int
     :returns: Identifier for the span. Must be a 64-bit integer other
               than 0 and unique within a trace.
     """
-    span_id = random.getrandbits(64)
+    span_id = random.randint(10**15, 10**16 - 1)
```
duggelz: Why did this change?

opencensus/trace/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    :param str_to_check: String to check.
+
+    :type limit: int
+    :param limit: The upper limit of the length.
+
+    :rtype: tuple
+    :returns: The string it self if not exceeded length, or truncated string
+              if exceeded and the truncated byte count.
+    """
+    if limit is None:
```
duggelz: Could also just do `limit=MAX_LENGTH` above in the function signature.

opencensus/trace/span_context.py
liyanhui1228: Because the V2 API only accepts 16-digit span_id, but we can also trim the span_id in the stackdriver_exporter.

duggelz: Ok.  Maybe document that rationale, and/or make those named constants.

liyanhui1228: Added.

opencensus/trace/utils.py
liyanhui1228: Done.

PR #92, opened 2017-12-05
Restructure the package

liyanhui1228:
For #54


setup.py
```diff
@@ -14,27 +14,29 @@
...
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: Implementation :: CPython',
+        'Programming Language :: Python :: 3.6',
```
duggelz: I think you accidentally replaced Python 3.5 here, instead of adding 3.6 on the line below.

nox.py
```diff
@@ -12,9 +12,100 @@
...
 
-import os
+from __future__ import absolute_import
 
 import nox
+import os
+
+
+@nox.session
+@nox.parametrize('python_version', ['2.7', '3.4', '3.5', '3.6'])
```
duggelz: The google-cloud-python packages did 'python_version' -> 'py', 'system_tests' -> 'system', 'unit_tests' -> 'unit'.  Do you have any opinion on doing the same here?  I think I'm fine with either.

nox.py
```diff
@@ -12,9 +12,100 @@
...
+
+    # Install all test dependencies, then install this package into the
+    # virutalenv's dist-packages.
+    session.install('-r', 'requirements-test.txt')
+    session.install('.')
+
+    # Run py.test against the system tests.
+    session.run(
+        'py.test',
+        '-vvv',
```
duggelz: Is this deliberately verbose, or a debugging line left in by accident?

setup.py
liyanhui1228: Good catch, added back.

nox.py
liyanhui1228: It's verbose to better see what happens in the system tests, but we can remove it for now.

nox.py
liyanhui1228: Changed.

PR #97, opened 2017-12-12
Restructure tracers

liyanhui1228:
- [x] Rename RequestTracer to Tracer
- [x] Remove the webapp2_tracer which should be in the ext module


README.rst
```diff
@@ -30,24 +30,24 @@ Installation & basic usage
 
     .. code:: python
 
-        from opencensus.trace import request_tracer
+        from opencensus.trace import tracer as tracer_module
```
duggelz: Yes, this looks good.  I always struggle with the "package/module/class/variable" naming, since I have up to four things but only two ways (Uppercase/lowercase) to name them.

README.rst
```diff
@@ -258,7 +258,7 @@ Webapp2
 
 .. code:: python
 
-    from opencensus.trace.tracer import webapp2_tracer
+    from opencensus.trace.tracers import webapp2_tracer
```
duggelz: I can't find the webapp2 files, are you going to add them to the 'ext' directory in this PR or a later one?

liyanhui1228: Deleted it for now, probably add it in later PR.

README.rst
liyanhui1228: :+1:

PR #99, opened 2017-12-15
Add a hello world example for grpc integration

liyanhui1228:
For #3.


examples/trace/grpc/hello_world_client.py
```diff
@@ -0,0 +1,43 @@
...
+        host_port=HOST_PORT)
+    channel = grpc.insecure_channel(HOST_PORT)
+    channel = grpc.intercept_channel(channel, tracer_interceptor)
+    stub = hello_world_pb2_grpc.HelloWorldStub(channel)
+    response = stub.SayHello(hello_world_pb2.HelloRequest(name='you'))
+    print("Message received: " + response.message)
+
+
+if __name__ == '__main__':
+    run()
```
duggelz: Style-wise, I would call this `main` rather than `run`.

examples/trace/grpc/hello_world_client.py
```diff
@@ -0,0 +1,43 @@
...
+    exporter = stackdriver_exporter.StackdriverExporter()
+    tracer = Tracer(exporter=exporter)
+    tracer_interceptor = client_interceptor.OpenCensusClientInterceptor(
+        tracer,
+        host_port=HOST_PORT)
+    channel = grpc.insecure_channel(HOST_PORT)
+    channel = grpc.intercept_channel(channel, tracer_interceptor)
+    stub = hello_world_pb2_grpc.HelloWorldStub(channel)
+    response = stub.SayHello(hello_world_pb2.HelloRequest(name='you'))
+    print("Message received: " + response.message)
```
duggelz: We should check that the response is what we expect.

examples/trace/grpc/hello_world_client.py
```diff
@@ -0,0 +1,43 @@
...
+import grpc
+
+import hello_world_pb2
+import hello_world_pb2_grpc
+
+from opencensus.trace.tracer import Tracer
+from opencensus.trace.exporters import stackdriver_exporter
+from opencensus.trace.ext.grpc import client_interceptor
+
+HOST_PORT = 'localhost:50051'
```
duggelz: You might look into the 'portpicker' package on PyPI for picking unused ports instead of hardcoding.  Hardcoding can run into problems when you're running tests on CI machines and can have collisions which lead to hard-to-debug failures.

duggelz: ... but you can do this later.

examples/trace/grpc/hello_world_client.py
liyanhui1228: Done.

examples/trace/grpc/hello_world_client.py
liyanhui1228: Yeah similar as above, will check that in the system test.

examples/trace/grpc/hello_world_client.py
liyanhui1228: Good suggestion! This is just an example and will not be running on CI. I'll do that when I added the system test for grpc integration.

duggelz: You're right, for a sample app, hard-coding ports is probably the right thing to, because it's simpler for the person trying to learn from the code.

PR #108, opened 2018-01-24
Update grpc server_interceptor to report accurate timing

### LINKED ISSUE #107: OpenCensusServerInterceptor reports incorrect timing
There is a bug in OpenCensusServerInterceptor that causes server spans to be finished immediately. You can reproduce it by running the helloworld grpc example in examples/trace/grpc. You should see a server span with a very short execution time, for example when I run it locally with a print exporter I get:

`{'traceId': 'a8fdeb7813144b178cb4b3d3d9b8fee4', 'spans': [{'childSpanCount': 0, 'displayName': {'truncated_byte_count': 0, 'value': 'grpc_server'}, 'parentSpanId': 4417962580138865, 'startTime': '2018-01-23T16:01:17.826761Z', 'attributes': {'attributeMap': {'/component': {'string_value': {'truncated_byte_count': 0, 'value': 'grpc'}}}}, 'spanId': 7098108131070600, 'endTime': '2018-01-23T16:01:17.826771Z'}]}`

The problem is that [this line](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/grpc/server_interceptor.py#L62) is being traced, but it isn't actually executing the server handler code. It just returns an RPCMethodHandler, which defers execution. 

I hacked up a [quick fix](https://gist.github.com/wkiser/d086a3f623f135265914e2b51bb722e8) that wraps the RPCMethodHandler and instruments the span when the handler is evaluated. A better solution might be to have the RpcMethodHandlerWrapper accept arbitrary callbacks, or to update grpc to have RPCMethodHandler accept callbacks natively.

Happy to submit a PR if you have a preferred approach. Thanks!

wkiser:
fixes #107. Updated my example to accept pre handler and post handler callbacks as requested. I also updated the code to set the tracer in the interceptor as the current tracer in the execution context.

### DIFF
--- a/opencensus/trace/ext/grpc/server_interceptor.py
+++ b/opencensus/trace/ext/grpc/server_interceptor.py
@@ -17,7 +17,8 @@ import logging
 import grpc
 
 from opencensus.trace import attributes_helper
-from opencensus.trace import tracer as tracer_module, execution_context
+from opencensus.trace import tracer as tracer_module
+from opencensus.trace import execution_context
 from opencensus.trace.ext import grpc as oc_grpc
 from opencensus.trace.propagation import binary_format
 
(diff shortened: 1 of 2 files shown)


opencensus/trace/ext/grpc/server_interceptor.py
```diff
@@ -11,37 +11,71 @@
...
+RpcRequestInfo = collections.namedtuple('RPCRequestInfo', ('request', 'context'))
+RpcResponseInfo = collections.namedtuple('RPCCallbackInfo', ('request', 'context', 'response', 'exc'))
+
+
+class RpcMethodHandlerWrapper(object):
+    """Wraps a grpc RPCMethodHandler and records the variables about the execution context and response"""
+
+    def __init__(self, handler, pre_handler_callbacks=None, post_handler_callbacks=None):
+        """
+        :param handler: instance of RpcMethodHandler
```
liyanhui1228: Please add one blank line between the param docstrings.

opencensus/trace/ext/grpc/server_interceptor.py
```diff
@@ -11,37 +11,71 @@
...
 # limitations under the License.
+import collections
+import logging
 
 import grpc
-import logging
 
 from opencensus.trace import attributes_helper
-from opencensus.trace import tracer as tracer_module
+from opencensus.trace import tracer as tracer_module, execution_context
```
liyanhui1228: Change this to two lines to import the `tracer` module and `execution_context` module separately.

liyanhui1228:
System tests are sometimes flaky which are not related to grpc integrations, merging for now.

Issue #117, opened 2018-02-13
Test pollution due to execution context not being cleared


bplotnick:
While writing some new tests, I hit test pollution (i.e. pytest passes with just my new test suite, but fails when running with others), which I think is from the httplib tests (added in #95) not clearing the execution context in the `tearDown` method.

Here's a snippet from my postmortem pdb session from pytest (from within the test I am writing that fails due to pollution):

```python
        span_id = 1234
        tracer = execution_context.get_opencensus_tracer()
        self.assertIsNotNone(tracer)

        span = tracer.current_span()

>       self.assertEqual(span.parent_span.span_id, span_id)
E       AssertionError: <Mock name='mock.span_id' id='4596810808'> != 1234

tests/unit/trace/ext/pyramid/test_pyramid_middleware.py:233: AssertionError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py(822)_baseAssertEqual()
-> raise self.failureException(msg)
(Pdb) up
> /usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py(829)assertEqual()
-> assertion_func(first, second, msg=msg)
(Pdb) up
> /Users/plotnick/opencensus-python/tests/unit/trace/ext/pyramid/test_pyramid_middleware.py(233)test__after_request()
-> self.assertEqual(span.parent_span.span_id, span_id)
(Pdb) p execution_context._thread_local.attrs
{'httplib/current_span_id': <Mock name='mock.span_id' id='4596869664'>}
```

This does not occur after adding `execution_context.clear()` to the `setUp()` test suite method (I already had it in the `tearDown()` method)

liyanhui1228:
[...]

PR #118, opened 2018-02-13
Clean up execution context in httplib unit tests

liyanhui1228:
For #117.


bplotnick:
Thanks! This fixes the test pollution I reported in #117 

Since we might hit this again, Maybe we should have either a TestCase baseclass that does this or a pytest `autouse` fixture that does this. There's some tradeoffs to each, but I think some mechanism would be useful.

PR #121, opened 2018-02-20
Pyramid support

bplotnick:
This adds support for the [pyramid](https://docs.pylonsproject.org/projects/pyramid/en/latest/index.html) web framework. It has mostly been cribbed from both the flask and django integrations, with a few pyramid-specific things.

Of note:
* It uses settings like django for the propagator/exporter/sampler, but does not take a class string. Instead, it takes an instance. I didn't really like the idea of putting implementation-specific logic in the middleware like the django one does. But I do like the global and per-request configurability of `settings`. e.g. swapping out samplers on the fly.
* I did not add a system test, because I am not able to get the existing system tests working. It looks like the system tests actually talk to stackdriver or something, which I am not able to do.
* I did add an example, which I think gives a good idea of how to use this thing

Let me know what you think!

### DIFF
--- a/opencensus/trace/ext/pyramid/pyramid_middleware.py
+++ b/opencensus/trace/ext/pyramid/pyramid_middleware.py
@@ -1,0 +1,107 @@
+# Copyright 2017, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+from opencensus.trace.ext import utils
+from opencensus.trace.ext.pyramid.config import PyramidTraceSettings
+
+from opencensus.trace import attributes_helper
+from opencensus.trace import execution_context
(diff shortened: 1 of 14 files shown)


opencensus/trace/ext/pyramid/pyramid_middleware.py
```diff
@@ -0,0 +1,107 @@
...
+
+            tracer.end_span()
+            tracer.finish()
+        except Exception:  # pragma: NO COVER
+            log.error('Failed to trace request', exc_info=True)
+
+
+def get_context_header(request):
+    """Get trace context header from pyramid request headers."""
+
```
liyanhui1228: Remove this blank line

opencensus/trace/ext/pyramid/pyramid_middleware.py
```diff
@@ -0,0 +1,107 @@
...
+_PYRAMID_TRACE_HEADER = 'X_CLOUD_TRACE_CONTEXT'
+
+BLACKLIST_PATHS = 'BLACKLIST_PATHS'
+
+
+log = logging.getLogger(__name__)
+
+
+class OpenCensusTweenFactory(object):
+    def __init__(self, handler, registry):
```
liyanhui1228: Add comments about what `handler` and `registry` are?

liyanhui1228:
System tests need the key file for the google cloud project that Stackdriver Trace uses, which is stored as an environment variable in Circle CI. So that it cannot run locally (for testing) if you do not have the key file. I'll add system tests after this is merged. And the example looks awesome!

bplotnick:
@liyanhui1228 Sorry for the delay. I updated this with the changes. Let me know if there's something else that this needs.

liyanhui1228:
@bplotnick Thanks for your contribution!

PR #144, opened 2018-03-23
Refactor exporters to accept SpanData objects instead of Spans

### LINKED ISSUE #127: Pass raw spans to exporters
Currently, span objects are converted to an intermediate json format before being sent to an exporter. As a result, there are some stackdriver specific conversions [being applied](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span.py#L222) to all spans (such as string truncation) that may not necessarily apply to all export formats.

I am proposing to update the [base.Exporter](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/exporters/base.py#L18) emit and export methods to take a raw [span.Span](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span.py#L28) object instead of a dict. Not only will this push stackdriver specific details to the stackdriver exporter, but it would also make the development of new exporters easier since there are less layers of conversion.

Happy to make a PR if everyone is ok with the change.

wkiser:
Fixes #127

### DIFF
--- a/opencensus/trace/exporters/file_exporter.py
+++ b/opencensus/trace/exporters/file_exporter.py
@@ -16,6 +16,7 @@
 
 import json
 
+from opencensus.trace import span_data
 from opencensus.trace.exporters import base
 from opencensus.trace.exporters.transports import sync
 
@@ -47,14 +48,25 @@ class FileExporter(base.Exporter):
         self.transport = transport(self)
         self.file_mode = file_mode
 
-    def emit(self, trace):
+    def emit(self, span_datas):
         """
-        :type trace: dict
-        :param trace: Trace collected.
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
         """
         with open(self.file_name, self.file_mode) as file:
-            trace_str = json.dumps(trace)
+            # convert to the legacy trace json for easier refactoring
+            # todo refactor this to use the span data directly
+            legacy_trace_json = span_data.format_legacy_trace_json(span_datas)
+            trace_str = json.dumps(legacy_trace_json)
             file.write(trace_str)
 
-    def export(self, trace):
-        self.transport.export(trace)
+    def export(self, span_datas):
+        """
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to export
+        """
+        self.transport.export(span_datas)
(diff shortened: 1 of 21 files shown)


wkiser:
@liyanhui1228, could you kick off circle again? It looks like it's stuck.

opencensus/trace/exporters/stackdriver_exporter.py
```diff
@@ -132,33 +132,46 @@ def __init__(self, client=None, project_id=None,
...
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
         """
         name = 'projects/{}'.format(self.project_id)
-        stackdriver_spans = self.translate_to_stackdriver(spans)
+
+        # convert to the legacy trace json for easier refactoring
+        # todo refactor this to use the span data directly
```
liyanhui1228: Please change the `todo` to uppercase `TODO:`.

opencensus/trace/exporters/file_exporter.py
```diff
@@ -47,14 +48,25 @@ def __init__(self, file_name=DEFAULT_FILENAME,
...
-        :param trace: Trace collected.
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
         """
         with open(self.file_name, self.file_mode) as file:
-            trace_str = json.dumps(trace)
+            # convert to the legacy trace json for easier refactoring
+            # todo refactor this to use the span data directly
```
liyanhui1228: Please change the `todo` to uppercase `TODO:`.

opencensus/trace/span_data.py
```diff
@@ -81,6 +89,81 @@ class SpanData(_SpanData):
...
+    :return: Legacy 'trace' dictionary representing given SpanData tuples
+    """
+    if not span_datas:
+        return {}
+    top_span = span_datas[0]
+    assert isinstance(top_span, SpanData)
+    trace_id = None
+    if top_span.context is not None:
+        trace_id = top_span.context.trace_id
+    return {
```
liyanhui1228: assert the `trace_id` is not None before return?

opencensus/trace/span_data.py
```diff
@@ -81,6 +89,81 @@ class SpanData(_SpanData):
...
+    :rtype: dict
+    :return: Legacy 'trace' dictionary representing given SpanData tuples
+    """
+    if not span_datas:
+        return {}
+    top_span = span_datas[0]
+    assert isinstance(top_span, SpanData)
+    trace_id = None
+    if top_span.context is not None:
+        trace_id = top_span.context.trace_id
```
liyanhui1228: Could be simplified to ` trace_id = top_span.context.trace_id if top_span.context is not None else None`.

liyanhui1228:
We need to disable circleci running the default nosetests, which runs the unittest before installing the test requirements. And override it with our nox tests command. I'm preparing the PR for updating the circleci config.

PR #152, opened 2018-04-10
Update gRPC span names to match spec

wkiser:


opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -211,3 +211,9 @@ def intercept_stream_stream(self, continuation, client_call_details,
         self._end_span_between_context(current_span)
 
         return response_it
+
+
+def _get_span_name(client_call_details):
+    """Generates a span name based off of the gRPC client call details"""
+    method_name = client_call_details.method[1:].replace('/', '.')
+    return '{}.{}'.format(SENT_PREFIX, method_name)
```
bogdandrutu: FYI: The prefix (sent, recv) will be replaced by the Span.Kind (https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/trace.proto#L66).

In SD exporter we will still attach the prefixes, but in Jagger and Zipkin we will just set the kind.

Here https://github.com/census-instrumentation/opencensus-specs/issues/85 (not documented) but the issue is about that.

liyanhui1228: Acknowledged, thanks!

PR #154, opened 2018-04-15
[WIP] Implement Jaeger Trace Exporter

sonmezonur:
This PR adds support for [Jaeger tracing](https://www.jaegertracing.io). It will allow exporting Trace data to a Jaeger instance using: 

- Compact Thrift Protocol over UDP
- Binary Thrift Protocol over HTTP

Thanks! Looking forward to receiving your comments.


opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+        self.collector = self._collector
+
+    @property
+    def _agent_client(self):
+        return AgentClientUDP(
+            host_name=self.agent_host_name,
+            port=self.agent_port)
+
+    @property
+    def _collector(self):
```
wkiser: You may want to cache this so you don't need to make a new thrift client for each batch.

opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+            end_microsec = calendar.timegm(end_datetime.timetuple()) * 1000
+
+            duration_microsec = end_microsec - start_microsec
+
+            tags = _extract_tags(span)
+
+            status = span.get('status', None)
+            if status is not None:
+                tags.append(jaeger.Tag(
+                    key='status.code',
```
wkiser: Can you use the error tags from the [opentracing spec](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#captured-errors)?
Also, can you add an error=true tag so that error spans will render as errors in the Jaeger UI?

opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+        return Collector(thrift_url=thrift_url, auth=auth)
+
+    def emit(self, span_datas):
+        """
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
+        """
+        trace = span_data.format_legacy_trace_json(span_datas)
```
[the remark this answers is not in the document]
wkiser: Thanks! You are actually the first person to make a new exporter since I made that change. I should probably go back and update some of the other exporters for reference.

opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+
+from opencensus.trace import span_data, link
+from opencensus.trace.exporters import base
+from opencensus.trace.exporters.transports import sync
+from opencensus.trace.exporters.gen.jaeger import jaeger
+from opencensus.trace.exporters.gen.jaeger import agent
+
+DEFAULT_HOST_NAME = 'localhost'
+DEFAULT_AGENT_PORT = 6931
+DEFAULT_ENDPOINT = '/api/traces'
```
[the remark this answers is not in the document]
sonmezonur: Updated

opencensus/trace/exporters/jaeger_exporter.py
sonmezonur: Agreed.

opencensus/trace/exporters/jaeger_exporter.py
wkiser: I meant for non-OK statuses you should additionally add an `error=True` tag.

I misspoke earlier, the other error information is stored as span logs and not tags. It looks like the only error related tag in OT is `error`. Maybe in a follow up PR (or this one) we should additionally encode the Status message and Stack Trace information into span logs.

opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+
+    @property
+    def _agent_client(self):
+        return AgentClientUDP(
+            host_name=self.agent_host_name,
+            port=self.agent_port)
+
+    @property
+    def _collector(self):
+        if self.host_name is None or self.port is None:
```
liyanhui1228: Can we make the `host_name` and `port` as required parameters in `__init__`? So that we don't need to do the check here.

opencensus/trace/exporters/jaeger_exporter.py
```diff
@@ -0,0 +1,451 @@
...
+        if self.username != '' and self.password != '':
+            auth = (self.username, self.password)
+
+        return Collector(thrift_url=thrift_url, auth=auth)
+
+    def emit(self, span_datas):
+        """
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
```
liyanhui1228: Remove `opencensus.trace.span_data.SpanData`?

opencensus/trace/exporters/jaeger_exporter.py
sonmezonur: I intended to only activate UDP Client(agent) by default. If host_name and port are undefined, collector object will not be constructed, thereby batch will not be uploaded into thrift HTTP server. Go-Client uses a similar pattern as I understand. But happy to update/change it if I'm wrong.

liyanhui1228: Thanks for clarifying! That makes sense!

sonmezonur:
Sorry for the delay. A quick summary of what I have done so far after reviews:

- Allow working with SpanData objects rather than the mystery json
- Refactor `_convert_hex_str_to_int()` to accept signed integers
- Set basic auth header in Collector
- Other minor improvements(caching properties, lint issues, sorting imports, etc.) 

Also, @wkiser suggests using error tags for non-OK statuses([spec](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#captured-errors)). But, while testing I realized that REST Middlewares(Django, Pyramid) do no update span status after the request sent, instead, it is written into attributes(Status always return empty for all my tests). So, I need clarification to add error tags and make it work.

Thanks! BTW, any comments and suggestions will be appreciated!

Issue #134, opened 2018-03-08
Error when using google_cloud_clientlibs: "integration 'NoopTracer' object has no attribute 'span_context'"


bshaffer:
With the latest release, `0.1.3`, I am getting the following exception when using the `google_cloud_clientlibs` trace integration:

```
Exception in thread Thread-19:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 701, in channel_spin
    completed_call = event.tag(event)
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 174, in handle_event
    callback()
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 313, in <lambda>
    self._state.callbacks.append(lambda: fn(self))
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/ext/grpc/client_interceptor.py", line 122, in callback
    self._tracer.end_span()
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/tracer.py", line 116, in end_span
    self.tracer.end_span()
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/tracers/context_tracer.py", line 112, in end_span
    self.exporter.export(span_json)
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/exporters/stackdriver_exporter.py", line 145, in export
    self.transport.export(trace)
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/exporters/transports/sync.py", line 23, in export
    self.exporter.emit(trace)
  File "/usr/local/lib/python2.7/dist-packages/opencensus/trace/exporters/stackdriver_exporter.py", line 142, in emit
    self.client.batch_write_spans(name, stackdriver_spans)
  File "/usr/local/lib/python2.7/dist-packages/google/cloud/trace/client.py", line 104, in batch_write_spans
    timeout=timeout)
  File "/usr/local/lib/python2.7/dist-packages/google/cloud/trace/_gax.py", line 91, in batch_write_spans
    timeout=timeout)
[...]

PR #157, opened 2018-04-18
Update gRPC client interceptor to use tracer from execution context

wkiser:
Ran into this when calling a gRPC client from within a flask endpoint handler. The tracer was set once when the interceptor was created and never updated afterwards.

### DIFF
--- a/opencensus/trace/ext/grpc/client_interceptor.py
+++ b/opencensus/trace/ext/grpc/client_interceptor.py
@@ -50,33 +50,30 @@ class OpenCensusClientInterceptor(grpc.UnaryUnaryClientInterceptor,
                                   grpc.StreamUnaryClientInterceptor,
                                   grpc.StreamStreamClientInterceptor):
 
-    def __init__(self, tracer=None, host_port=None):
-        if tracer is None:
-            tracer = execution_context.get_opencensus_tracer()
-
-        self._tracer = tracer
+    def __init__(self, host_port=None):
         self.host_port = host_port
         self._propagator = binary_format.BinaryFormatPropagator()
 
     def _start_client_span(self, client_call_details):
-        span = self._tracer.start_span(
+        tracer = execution_context.get_opencensus_tracer()
+        span = tracer.start_span(
             name=_get_span_name(client_call_details)
         )
 
         # Add the component grpc to span attribute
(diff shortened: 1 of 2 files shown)


opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -50,41 +50,39 @@ class OpenCensusClientInterceptor(grpc.UnaryUnaryClientInterceptor,
                                   grpc.StreamUnaryClientInterceptor,
                                   grpc.StreamStreamClientInterceptor):
 
-    def __init__(self, tracer=None, host_port=None):
```
liyanhui1228: Please keep the `tracer` parameter here since we also need to handle the case that sharing the tracer between threads. See #133, which was addressing issue #134 caused by `tracer` not shared between threads when instrumenting google cloud client library calls.

opencensus/trace/ext/grpc/client_interceptor.py
```diff
@@ -50,41 +50,39 @@ class OpenCensusClientInterceptor(grpc.UnaryUnaryClientInterceptor,
...
+        tracer.add_attribute_to_current_span(
             attribute_key=attributes_helper.GRPC_ATTRIBUTES.get(
                 GRPC_HOST_PORT),
             attribute_value=self.host_port)
 
         # Add the method to span attribute
-        self._tracer.add_attribute_to_current_span(
+        tracer.add_attribute_to_current_span(
             attribute_key=attributes_helper.GRPC_ATTRIBUTES.get(GRPC_METHOD),
             attribute_value=str(client_call_details.method))
```
liyanhui1228: Probably add code below to here to update the `tracer` in execution context:
```
execution_context.set_opencensus_tracer(self._tracer)
execution_context.set_current_span(span)
```

opencensus/trace/ext/grpc/client_interceptor.py
wkiser: Will do! 

Would the current integration in `google_cloud_clientlibs` be facing this same issue if it is passing in the tracer at initialization time? For instance, if a Flask endpoint calls a gcloud service, would the gcloud span be nested under the flask one?

PR #165, opened 2018-04-22
trace: Span.add_annotation method

### LINKED ISSUE #164: span: provide a helper to add annotations
I kindly would like to the addition of a helper method on Span, `add_annotation`

My use case is to add an annotation to the span like this
```python
with tracer.span(name='close') as span:
     if self.sock is None:
           span.add_annotation('Socket was already closed')
     else:
           span.add_annotation('Socket is still open')
           try:
                self.sock.close()
           except Exception as e:
                span.add_annotation('Encountered an exception while closing', recoverable=True, method='memcache-close'))
```

I ask because as a helper method makes it easy for annotations to be made, without having to import too much or go through all the docs.

Perhaps something that wraps `add_time_event` say:

```python
def add_annotation(self, description, **attributes):
    tevt = TimeEvent(datetime.utcnow(), Annotation(description, Attributes(attributes)))
    self.add_time_event(tevt)
```

Thank you.

odeke-em:
Fixes #164

Adds a convenience method so that users of the API
don't have to import a lot in order to add an annotation.
I found the need for this API while instrumenting a client.
It avoids userland imports of:
* datetime
* attributes.Attributes
* trace.time_event.Annotation
* trace.time_event.TimeEvent

abstracting the above work, to no imports and just:
```python
span.add_annotation('Here is my annotation', key1='Foo', key2=12)
```

I believe the other implementations have a
method that helps add annotations.

Language|MethodName|Resource
---|---|---
Go|Span.Annotate|https://godoc.org/go.opencensus.io/trace#Span.Annotate
Java|Span.addAnnotation|https://www.javadoc.io/doc/io.opencensus/opencensus-api/0.12.2

### Exhibit
screen shot 2018-04-21 at 5 17 40 pm


opencensus/trace/span.py
```diff
@@ -176,6 +176,16 @@ def add_attribute(self, attribute_key, attribute_value):
         """
         self.attributes[attribute_key] = attribute_value
 
+    def add_annotation(self, description, **attrs):
+        """Add an annotation to span.
+
+        :type description: str
```
liyanhui1228: nit: please update the documentation like below,
```
:type description: str
:param description: A user-supplied message describing the event.
                    The maximum length for the description is 256 bytes.

:type attrs: dict
:param attrs: keyworded arguments e.g. failed=True, name='Caching'
```

odeke-em: Nice doc update request, let me do that, thank you!

odeke-em: Although the `type attrs: dict` perhaps isn't that accurate as they are keyworded arguments. Passing in a dict d (without unfurling it i.e. **d) will register that as a non-keyworded argument :)

odeke-em:
I've updated the docs update request, PTAL @liyanhui1228

opencensus/trace/span.py
liyanhui1228: Right, I was thinking what type is accurate to describe the keyword arguments, but cannot think of a better one. `kwargs` is fine.

odeke-em: Cool.

odeke-em:
@liyanhui1228, unrelated to this change, the latest code from master that got merged in here breaks the threaded tests as we can see from CircleCI https://circleci.com/gh/census-instrumentation/opencensus-python/249?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

odeke-em:
Haha, not a problem at all, I also looked at the comment length but decided to wing it given that one of the comments below looked close in length to it.

odeke-em:
Thank you for the review and merge @liyanhui1228!

PR #166, opened 2018-04-23
Update the should_sample logic for tracer

### LINKED ISSUE #159: Missing support for "forced tracing"
I think you use the trace options from StackDriver incorrectly, based on this:
https://cloud.google.com/trace/docs/support?hl=sr#how_do_i_force_a_request_to_be_traced

`o=1` should mean forced tracing, regardless of the decision of the sampler. So here:

https://github.com/census-instrumentation/opencensus-python/blob/cfcfda584ba984e6f0e89684120c1ea94271b2c9/opencensus/trace/tracer.py#L68-L77

you should probably say `or` instead of `and` but please correct me if I'm missing something. To me it's more intuitive from a developer POV to have a way to force tracing for debugging reasons when needed and otherwise use some constant sampling rate. And maybe I would rename `enabled` to `forced`. I'm happy to open PR ofc if this makes sense by first look to you

liyanhui1228:
Fixes #159


opencensus/trace/tracer.py
```diff
@@ -74,7 +74,7 @@ def should_sample(self):
         :returns: Whether to trace the request or not.
```
wkiser: can you update the docstring?

liyanhui1228: Done.

Issue #167, opened 2018-04-25
Trace header should be tied to propagator and not web framework


justinwp:
for example: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/flask/flask_middleware.py#L34

This is related to https://github.com/census-instrumentation/opencensus-python/issues/46#issuecomment-384423270

Issue #46, opened 2017-10-24
Create a formatter class to handle different trace header format


liyanhui1228:
Need to automatically figure out what trace format the user is using, including Zipkin, AWS, StackDriver and TraceContext headers.

justinwp:
Running into this trying to support the [zipkin b3 headers](https://github.com/openzipkin/b3-propagation) being injected by istio. Seems like the single header format is fairly pervasive in the codebase. Any suggestions on structural changes to support this.

PR #168, opened 2018-04-25
Allow Zipkin B3 headers

justinwp:
My use case it tying the application level tracing into [Istio](https://istio.io/docs/tasks/telemetry/distributed-tracing.html#understanding-what-happened) which uses the [openzipkin B3 propogation format](https://github.com/openzipkin/b3-propagation).

related to #167 and #46

### DIFF
--- a/opencensus/trace/ext/django/middleware.py
+++ b/opencensus/trace/ext/django/middleware.py
@@ -20,6 +20,7 @@ from opencensus.trace.ext.django.config import (settings, convert_to_import)
 from opencensus.trace import attributes_helper
 from opencensus.trace import tracer as tracer_module
 from opencensus.trace import execution_context
+from opencensus.trace.propagation import zipkin_b3_format
 from opencensus.trace.samplers import probability
 
 try:
@@ -85,11 +86,28 @@ def get_django_header():
     :rtype: str
     :returns: Trace context header in HTTP request headers.
     """
+    headers = get_django_headers([_DJANGO_TRACE_HEADER])
+
+    return headers.get(_DJANGO_TRACE_HEADER)
+
+
+def get_django_headers(header_names):
+    """Get trace context headers from django request headers."""
     request = _get_django_request()
 
-    header = request.META.get(_DJANGO_TRACE_HEADER)
+    headers = {}
+    for name in header_names:
+        django_name = name
+        if not django_name.startswith("HTTP_"):
+            django_name = "HTTP_{}".format(django_name)
 
-    return header
+        django_name = django_name.replace("-", "_").upper()
+
+        header = request.META.get(django_name)
+
+        headers[name] = header
+
+    return headers
 
 
 class OpencensusMiddleware(MiddlewareMixin):
@@ -155,8 +173,13 @@ class OpencensusMiddleware(MiddlewareMixin):
 
         try:
             # Start tracing this request
-            header = get_django_header()
-            span_context = self.propagator.from_header(header)
+            if isinstance(self.propagator,
+                          zipkin_b3_format.ZipkinB3FormatPropagator):
+                headers = get_django_headers(self.propagator.HEADERS)
+                span_context = self.propagator.from_carrier(headers)
+            else:
+                header = get_django_header()
+                span_context = self.propagator.from_header(header)
 
             # Reload the tracer with the new span context
             tracer = tracer_module.Tracer(
--- a/opencensus/trace/ext/flask/flask_middleware.py
+++ b/opencensus/trace/ext/flask/flask_middleware.py
@@ -14,6 +14,7 @@
 
 import inspect
 import logging
+import six
 import sys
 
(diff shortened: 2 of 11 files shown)


wkiser:
Ideally we could keep logic around which headers to use out of the middleware and instead put it into the propagators. This would require some refactoring of the current HTTP propagators since they only act on a single header value.

This might also be a good opportunity to standardize the HTTP propagator interface. The [spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#propagation) suggests that the propagator should modify the request object directly, but that makes less sense in Python where there is no common http request type.

Maybe we could standardize the HTTP propagators to act on a dictionary of headers instead? That seems to be the most pythonic way. Something like:
```python
class HttpPropagator(object):

    @classmethod
    def span_context_from_headers(cls, headers):
        """Creates a SpanContext object from a dictionary of http headers"""
        raise NotImplementedError

    @classmethod
    def headers_from_span_context(cls, span_context):
        """Creates a dictionary of http headers from a SpanContext"""
        raise NotImplementedError
```

justinwp:
>Maybe we could standardize the HTTP propagators to act on a dictionary of headers instead? That seems to be the most pythonic way. 

This was already there with the `from_carrier` and `to_carrier` of the `TextFormatPropagator`.

wkiser:
>This was already there with the from_carrier and to_carrier of the TextFormatPropagator.

True. I think what I was trying to say is that we should use the interface on `TextFormatPropagator` to standardize how https traces are propagated.  Probably by making `GoogleCloudFormatPropagator` use the same interface.

That way you can have header agnostic code in your middleware. For example flask would look like:
```python
span_context = self.propagator.from_carrier(dict(request.headers))
```
and it would be up to the developer to instantiate the middleware with the appropriate propagator e.x. B3, Opencensus, GoogleCloud, etc

opencensus/trace/ext/django/middleware.py
```diff
@@ -85,11 +86,28 @@ def get_django_header():
     :rtype: str
     :returns: Trace context header in HTTP request headers.
     """
+    headers = get_django_headers([_DJANGO_TRACE_HEADER])
```
liyanhui1228: If we want to add `get_django_headers(header_names)`, we could remove the `get_django_header()` method, since they basically have no difference. And it would be easy to make a typo calling the wrong method.

opencensus/trace/ext/django/middleware.py
```diff
@@ -155,8 +173,13 @@ def process_request(self, request):
 
         try:
             # Start tracing this request
-            header = get_django_header()
-            span_context = self.propagator.from_header(header)
+            if isinstance(self.propagator,
```
liyanhui1228: After we combine the `get_django_headers` and `get_django_header`, we don't need the if else here any more. And the `propagator.from_carrier` could be renamed to use `from_header`.

opencensus/trace/ext/flask/flask_middleware.py
```diff
@@ -244,10 +250,28 @@ def get_flask_header():
...
     """
-    header = flask.request.headers.get(_FLASK_TRACE_HEADER)
+    return get_flask_headers([_FLASK_TRACE_HEADER]).get(_FLASK_TRACE_HEADER)
 
-    # In case the header is unicode, convert it to string.
-    if header is not None:
-        header = str(header.encode('utf-8'))
 
-    return header
+def get_flask_headers(header_names):
```
liyanhui1228: Same as `get_django_header`, we don't need a separate method for `get_flask_header`.

opencensus/trace/ext/pyramid/pyramid_middleware.py
```diff
@@ -121,4 +127,17 @@ def _after_request(self, request, response):
 
 def get_context_header(request):
```
liyanhui1228: Remove this and combine with `get_context_headers`? We can set the default value for `header_names` to `[_PYRAMID_TRACE_HEADER]`.

opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -70,7 +70,7 @@ def from_header(self, header):
         else:
             logging.warning(
                 'Cannot parse the header {}, generate a new context instead.'
-                .format(header))
+                .format(header.encode('utf-8', errors='ignore')))
```
liyanhui1228: Why this encoding is needed?

negz:
https://github.com/census-instrumentation/opencensus-go/pull/755

FWIW I'm interested in adding linkerd style propagation support (which propagates traces in the `l5d-ctx-trace` header rather than `X-B3-*`) to opencensus-python. It seems like it might be best for me to wait for this PR to be merged first?

reyang:
B3 support has been added by #265. Closing this PR as a dup.
@justinwp @kornholi @negz

PR #172, opened 2018-05-08
Removed unused propagator from Tracer

wkiser:


liyanhui1228:
Thanks for this! Could you provide more context about why we want to remove the propagator from tracer? And how we are going to set the propagator for a tracer?

reyang:
Propagators are now used by tracer, closing this PR.

PR #184, opened 2018-05-25
Stats: fix a few bugs in measure_to_view_map.

songy23:
1. The arg `stats` that `measure_to_view_map.record()` takes is a `measurement_map`. Its key is a `Measure`, rather than a `Measurement`.
2. `measure_to_view_map.map` is a multi-map, the value of it is a list. `view_datas` needs to extend this list rather than append it.


songy23:
>  please fix the lint (line too line) before merge.

Fixed, thanks!

PR #188, opened 2018-05-29
changed time precision on zipkin exporter from seconds to microseconds

felippe-mendonca:
I had a problem when using zipkin exporter provided by opencensus-python with another [zipkin exporter (c++)](https://github.com/rnburn/zipkin-cpp-opentracing). Due to seconds precision, spans generated by this exporter usually appear before parent span generate by c++ exporter.


liyanhui1228:
HI @felippe-mendonca, thanks for this change! Please also update the expected_timestamp in the unit tests with the updated precision.

PR #194, opened 2018-05-30
Fixing NoneType Error in ViewData._tag_map

kcooperstein:


opencensus/stats/view_data.py
```diff
@@ -72,13 +73,17 @@ def end(self):
...
-            return tags
+        try:
+            if context.items() <= self.tag_map.items():
+                return self.tag_map
+            else:
+                tags = self.tag_map
+                for tag_key, tag_value in context.items():
+                    tags[tag_key] = tag_value
+                return tags
+        except ValueError:  # pragma: NO COVER
```
songy23: Suggestion: is it possible to check the type of `self.tag_map` instead of doing the try-except logic?

kcooperstein: We can check is self.tag_map is None with if/else, should I do that instead?

opencensus/stats/view_data.py
```diff
@@ -72,13 +73,17 @@ def end(self):
...
+            if context.items() <= self.tag_map.items():
+                return self.tag_map
+            else:
+                tags = self.tag_map
+                for tag_key, tag_value in context.items():
+                    tags[tag_key] = tag_value
+                return tags
+        else:  # pragma: NO COVER
+            logging.warning("Tag Map cannot be none",
+                            self.tag_map)
```
liyanhui1228: Just a message in logging is enough, it will throw `TypeError` when `self.tag_map` is None in the logging call here.

songy23:
Please rebase.

PR #196, opened 2018-05-30
Trace: Refactor exporters to use SpanData objects

sonmezonur:
Under this PR, we may update each trace exporter to directly use SpanData objects. This will both reduce the LoC and eliminate the additional step to convert 'span_datas' to JSON format. 

**Updates:** 

* Directly use SpanData objects instead of JSON to export spans in StackDriver/Zipkin exporters

* Log SpanData object directly in LoggingExporter

* Add pkl format to serialize objects to the file in FileExporter

* Additionally, update docstrings in JaegerExporter

**Note:** We may either log formatted JSONs(previous one) or SpanData objects in LoggingExporter. Which one should be?

### DIFF
--- a/opencensus/trace/exporters/file_exporter.py
+++ b/opencensus/trace/exporters/file_exporter.py
@@ -15,12 +15,13 @@
 """Export the trace spans to a local file."""
 
 import json
+import pickle
 
 from opencensus.trace import span_data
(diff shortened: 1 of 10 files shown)


sonmezonur:
Closing this one. I will try to split this PR into small separate PRs

PR #198, opened 2018-06-04
Fixing Non-Shared Measure to View Map

kcooperstein:


opencensus/stats/stats_recorder.py
```diff
@@ -24,7 +26,16 @@ class StatsRecorder(object):
...
     def __init__(self, measure_to_view_map=None):
-        self.measure_to_view_map = measure_to_view_map or {}
+        if measure_to_view_map is None:
+            if execution_context.get_measure_to_view_map() is None:
+                self.measure_to_view_map = MeasureToViewMap()
+                execution_context.set_measure_to_view_map(MeasureToViewMap())
+            else:
+                self.measure_to_view_map = \
+                    execution_context.get_measure_to_view_map()
+        else:
```
songy23: Consider this case:
```python
# Create a new MeasureToViewMap and store it in thread local
view_manager = ViewManager()

# Create another MeasureToViewMap
measure_to_view_map_new = MeasureToViewMap()
# Use the new one and store it in thread local, but view_manager is still using the old measure_to_view_map
stats_recorder = StatsRecorder(measure_to_view_map_new)
```

And the `measure_to_view_map` in `ViewManager` and `StatsRecorder` could still go unsync. Thus I think `StatsRecorder` should always use the `measure_to_view_map` from thread local, instead of passing it as an arg.

PR #199, opened 2018-06-04
Fixing NoneType Error in ViewData._tag_map

kcooperstein:


opencensus/stats/view_data.py
```diff
@@ -72,13 +73,16 @@ def end(self):
...
+        if self.tag_map is not None:
+            if context.items() <= self.tag_map.items():
+                return self.tag_map
+            else:
+                tags = self.tag_map
+                for tag_key, tag_value in context.items():
+                    tags[tag_key] = tag_value
+                return tags
+        else:  # pragma: NO COVER
+            logging.warning("Tag Map cannot be none")
```
liyanhui1228: Do we need to return something if `self.tag_map` is `None`?

PR #201, opened 2018-06-05
Update Readme to use OPENCENSUS_TRACE.EXPORTER

dhendry:
Trivial documentation update. It looks like OPENCENSUS_TRACE.REPORTER is an old thing. As per https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/django/config.py#L22


dhendry:
I signed it!

PR #202, opened 2018-06-06
Run integration tests with different permutations of dependencies.

brianquinlan:
I'm not sure if this is a good idea or not.


nox.py
```diff
@@ -27,6 +27,7 @@ def unit(session, py):
     session.interpreter = 'python{}'.format(py)
 
     # Install all test dependencies, then install this package in-place.
+    session.install('google-cloud-trace')
```
liyanhui1228: Are we going to run unit tests against the latest version of `google-cloud-trace`? While this can make sure our library works well with the latest release version of it, but it would be confusing whether the tests failed because of the version incompatibility issue or the tested code itself.

I would suggest we pin the version used for tests and add tooling for ensuring we work well with the latest release when the tools is built.

brianquinlan: Should we pin the version or test with the versions matching setup.py i.e.
google-cloud-trace=0.17.0, google-cloud-trace<0.20
google-api-core = 0.1.1, google-api-color < 2.0.0
?

PR #203, opened 2018-06-08
Fix app engine flex environment vars in stackdriver_exporter and be more defensive in postgresql/trace.py

dhendry:
This PR updates the environment variables used by the stackdriver_exporter to their latest format (as per https://cloud.google.com/appengine/docs/flexible/python/runtime#environment_variables). I guess google dropped `_FLEX_` at some point?

Also it seems like `g.co/gae/app/service` just does not do anything in the stackdriver UI but `g.co/gae/app/module` does. This ruby code is the most official documentation I could find regarding which attributes are actually special: https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/label_key.rb#L46. I have tested this in our python services but version filtering still does not work (it seems no matter what special attribute is used, including `g.co/gae/app/module_version` - I guess this is a bug in slackdriver?).

Finally, the change to the postgress tracer is needed since the requests integration blindly calls `execution_context.set_opencensus_tracer(tracer)` even when the `tracer` argument is None - which is what I _think_ is expected in most cases (see: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/requests/trace.py#L34 and https://github.com/census-instrumentation/opencensus-python#service-integration which does not set the tracer to anything). I think this is a bug in the requests integration but regardless, the postgres tracer should be more defensive. Note that this whole change is to protect against a problem I was experiencing where db quires were failing if a tracer had not yet been created.


dhendry:
@liyanhui1228 Test added for null condition on postgres driver.

liyanhui1228:
Thanks for the PR!

PR #219, opened 2018-07-02
refactor - move header name to propagator (part 2)

### LINKED ISSUE #167: Trace header should be tied to propagator and not web framework

reyang:
Fix #167

### DIFF
--- a/examples/stats/helloworld/main.py
+++ b/examples/stats/helloworld/main.py
@@ -22,11 +22,12 @@ from opencensus.stats import view as view_module
 from opencensus.tags import tag_key as tag_key_module
 from opencensus.tags import tag_map as tag_map_module
 from opencensus.tags import tag_value as tag_value_module
+from pprint import pprint
 
 MiB = 1 << 20
 FRONTEND_KEY = tag_key_module.TagKey("my.org/keys/frontend")
 VIDEO_SIZE_MEASURE = measure_module.MeasureInt(
-    "my.org/measure/video_size", "size of processed videos", "By")
+    "my.org/measures/video_size", "size of processed videos", "By")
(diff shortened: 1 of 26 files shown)


reyang:
@liyanhui1228, is this a transient failure while running system test? Would you help to trigger the CI? Thanks!

reyang:
@liyanhui1228, would you help to look in to the CI error?
I can look into the issue if there a way to reproduce them on my local environment - do you have instructions on how to run system test locally?

SergeyKanzhelev:
I created a work item to track contributing.md update: https://github.com/census-instrumentation/opencensus-python/issues/224

kornholi:
Hi @reyang, the CI failures are actually valid because of how Django works.

In `request.META`, all incoming headers are prefixed with `HTTP_`, so `X-Cloud-Trace-Context` gets stored as `HTTP_X_CLOUD_TRACE_CONTEXT` instead of `X_CLOUD_TRACE_CONTEXT`.

I think the best solution here is to wrap `request.META` with an object where `__getitem__` rewrites the key `X-Foo` to `HTTP_X_FOO`, then pass that object into `from_headers`. What do you think?

reyang:
@kornholi, I see, thanks for the hint!
I agree with the your suggestion, let me make the code change now.

kornholi:
@liyanhui1228 PTAL. I will refactor the Zipkin/B3 PR after this is in.

PR #220, opened 2018-07-03
Support trace propagation in httplib integration

reyang:
With this change, the httplib will propagate context while sending requests.

Sample code:

```import requests

config_integration.trace_integrations(['httplib'])
tracer = Tracer()

def main():
	with tracer.span(name = 'client'):
		response = requests.get('http://localhost:5000')
		print(response)

if __name__ == '__main__':
	main()
```


opencensus/trace/ext/httplib/trace.py
```diff
@@ -57,7 +57,7 @@ def wrap_httplib_request(request_func):
     """Wrap the httplib request function to trace. Create a new span and update
     and close the span in the response later.
     """
-    def call(self, method, url, *args, **kwargs):
```
SergeyKanzhelev: I think adding this feature requires new unit tests

tests/unit/trace/ext/httplib/test_httplib_trace.py
```diff
@@ -76,7 +76,7 @@ def test_wrap_httplib_request(self):
         method = 'GET'
 
         with patch:
-            wrapped(mock.Mock(), method, url)
```
SergeyKanzhelev: it will be really cool beyond the unit test have some functional test with actual httplib doing actual POST and GET requests.

SergeyKanzhelev: This may be a separate work item though

opencensus/trace/ext/httplib/trace.py
reyang: updated unit test to reflect the change

SergeyKanzhelev:
I signed it!

reyang:
Created a new PR #237 instead.

PR #221, opened 2018-07-06
Store fake Tracer when parent_span is None

therc:
Right now, the code checks if parent_span is None only after it's stored.
Either the checks runs when it's too late or  it's not needed in the first place. Assume the former.


therc:
Uh, this breaks TestSpan.test_constructor_defaults. So perhaps the code serves no purpose? @liyanhui1228, you added the code in https://github.com/therc/opencensus-python/commit/a155f47d8225d2c399165bfc24da25e1f6e062a0, is it a bad copy and paste from context_tracer.py?

therc:
Ping?

PR #222, opened 2018-07-09
Add a propagator for jaeger's uber-trace-id header.

tcolgate:


tcolgate:
I'll be raising PRs for node and Go, assuming there aren't outright objections to supporting this at all.

opencensus/trace/propagation/jaeger_format.py
```diff
@@ -0,0 +1,121 @@
...
+from opencensus.trace.trace_options import TraceOptions
+
+_TRACE_CONTEXT_HEADER_NAME = 'UBER_TRACE_ID'
+_TRACE_CONTEXT_HEADER_FORMAT = '([0-9a-f]{0,32}|)\:([0-9a-f]{0,16})\:([0-9a-f]{0,16}|)\:([0-9a-f]{2})'
+_TRACE_CONTEXT_HEADER_RE = re.compile(_TRACE_CONTEXT_HEADER_FORMAT)
+
+
+class JaegerFormatPropagator(object):
+    """This class is for converting the trace header in Jaeger's format
+    and generate a SpanContext, or converting a SpanContext to a google cloud
```
SergeyKanzhelev: >converting a SpanContext to a google cloud 
copy/paste issue?

opencensus/trace/propagation/jaeger_format.py
```diff
@@ -0,0 +1,121 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import re
+
+from opencensus.trace.span_context import SpanContext
+from opencensus.trace.trace_options import TraceOptions
+
+_TRACE_CONTEXT_HEADER_NAME = 'UBER_TRACE_ID'
```
SergeyKanzhelev: I'd suggest to name it `_UBER_HEAER_NAME` to avoid confusions

tcolgate: Updated

opencensus/trace/propagation/jaeger_format.py
tcolgate: blast! caught out ;)
Fixed

opencensus/trace/propagation/jaeger_format.py
```diff
@@ -0,0 +1,121 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import re
+
+from opencensus.trace.span_context import SpanContext
+from opencensus.trace.trace_options import TraceOptions
+
+_UBER_HEADER_NAME = 'UBER_TRACE_ID'
```
adriancole: is jaeger literally still using the name uber in its header format? if not, and this feature is desirable in core codebase... maybe parameterize it?

tcolgate: It is, but it wont be changing. They intend to move to the w3c trace header spec which is entirely different to the existing header.

opencensus/trace/propagation/jaeger_format.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def to_header(self, span_context):
+        """Convert a SpanContext object to header string.
+
+        :type span_context:
+            :class:`~opencensus.trace.span_context.SpanContext`
+        :param span_context: SpanContext object.
+
+        :rtype: str
+        :returns: A trace context header string in google cloud format.
```
SergeyKanzhelev: another mention of Google =) Sorry didn't spot it first time.

opencensus/trace/propagation/jaeger_format.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def to_headers(self, span_context):
+        """Convert a SpanContext object to HTTP request headers.
+
+        :type span_context:
+            :class:`~opencensus.trace.span_context.SpanContext`
+        :param span_context: SpanContext object.
+
+        :rtype: dict
+        :returns: Trace context headers in google cloud format.
```
SergeyKanzhelev: and one more

tcolgate: @SergeyKanzhelev you'd think I'd know how to use sed or something! Sorry about that, fixed and pushed.

SergeyKanzhelev:
Please fix CI:

```
opencensus/trace/propagation/jaeger_format.py:22:80: E501 line too long (93 > 79 characters)
opencensus/trace/propagation/jaeger_format.py:65:72: E231 missing whitespace after ','
```

tcolgate:
things are now failing on code coverage it seems. I'll need to get the tests running locally and see if I can improve the unit tests.

adriancole:
I find this change wholly objectional especially if it is to be repeated on all repos as a backdoor addition to the spec. This particular change [highlights how uber, despite being well staffed and assisted by red hat, were unwilling to change](https://github.com/jaegertracing/jaeger/issues/121#issuecomment-295798165) when it was easier to do to in order to facilitate others. Now, we are expected to stain census (forever?) with the word Uber, which is already objectional to some, due to this, and knowing that it is a dead format.

This should be very optional imho as no-one is emulating uber aka jaeger. No firewalls are premade to pass this header.. it has practically zero merit outside uniform deployments of jaeger libraries. I would have a different opinion if this were the Amazon header as at least that is a cloud and there are places with high reuse where it makes sense.

This specifically issue aside.. this is an open project which should focus on least future weight balanced with pragmatism when deciding what to add to core libraries. I'm very surprised I am the only one mentioning this!! A healthy project doesn't add debt for itself due to temporary issues by one implementor!

Anyway, if this change is to be accepted, I'd recommend it be contained into a uber/jaeger namespace especially when this change is considered for other libraries like java.

tcolgate:
@adriancole on a practical note, jaeger is one of the more accessible and practical tracers to deploy right now, and this is the default behaviour of its clients. The same criticism could be levelled at stackdriver for not just using the zipkin b3 headers (or, frankly, anyone that came after zipkin). It's a shame some random company I don't care about will have its name in some of my http headers, but I'm not sure I can raise quite your level of ire about it.

I'm happy to keep this somewhere else.

Feel free to shoot down https://github.com/census-instrumentation/opencensus-node/pull/74 too.

adriancole:
>
> The same criticism could be levelled at stackdriver for not just using the
> zipkin b3 headers (or, frankly, anyone that came after zipkin). It's a
> shame some random company I don't care about will have its name in some of
> my http headers, but I'm not sure I can raise quite your level of ire about
> it.
>
The problem isn't a different header, it isn't about B3. It is a number of
things.

* Firstly stackdriver was not an "open project" at the time, and I would
have no idea if customers asked them to change as I wouldn't see their
issues list. That's a distraction.
* Also, my ire is not about them not using B3.. in fact they do support
that!. I don't care about that part.
* The issue I linked shows clearly them making a decision to make the
public code not be brandstamped in header form forever.. not with jaeger
but with uber. You have to permanently link jaeger with uber when
troubleshooting tracing?!

This problem is pushed down here now? great at least make it optional like
that can be removed from the core library when uber decides it is ok to do
that.

tcolgate:
The issue you linked shows them wanting to avoid a massive pile of internal work. I am unconviinced that http headers are an effective marketing channel.
Feel free to close, I can keep this out of tree.

tcolgate:
@SergeyKanzhelev I'll wait for you to tell me if this stands a chance of getting merged before I continue;  Seems like making the `uber` part configurable might be desirable for jaeger integration anyway (since some of jaegers client libs supported that)

SergeyKanzhelev:
@adriancole I agree that python repo as it is today is not flexible as everything is in the same package. Exporters and header formats should be moved to addons. This is something that needs to be done anyway.

For this particular PR - I believe supporting easy way to configure OC for all-jaeger environments gives a way for customers to migrate to OC, which is a spirit of the project. It is our goal to create a single library for distributed tracing so there is no question how to instrument the client library to expose telemetry.

This is why I believe this PR is OK with two work items to follow:

1. Split one module into core and add-ons.
2. Discuss/develop the way to combine multiple headers so there is a story to move to W3C one component at a time.

I see nine thumbsups on your comment. I'd like to understand if what I said is unreasonable. Also, @tcolgate - do you think you can help with those two items I listed?

tcolgate:
@SergeyKanzhelev 

> 1 Split one module into core and add-ons.
Do you mean the package itself? I've zero experience publishing python packages, and am still learning this code base.
> 2 Discuss/develop the way to combine multiple headers so there is a story to move to W3C one component at a time.
This seems like it would need to be a higher level OC discussion. Apps would need to agree on the approaches I guess. @adriancole seems more qualified to discuss that.

I'm personally more tempted at this point to work on the w3c header support in the libraries I care about and do what I can to help that migration. 

All my bandwidth is currently stolen by attempting to get OC to play nicely with various nodejs and python async frameworks (which is a level of pain I can barely comprehend in two language I'd really rather not be using, but I'm here to help the devs right!)

wu-sheng:
@SergeyKanzhelev From my understanding, the most important thing is OC Python should have an add-on first, then talk about moving this PR in.

tcolgate:
FWIW, perhaps people are misunderstanding slightly. The PR as is is a separate propagator which can be passed to the Tracer. It's already optional, nothing uses unless people select it. The obejction seems to be to having this support in the repository at all, so I'm not entirely sure why having this in a different OC repo would make any difference at all.
  I can publish it separately if the general association with jaeger/uber is so objectionable.

wu-sheng:
@tcolgate I can understand this is optional for now. I just ask @SergeyKanzhelev to help on format project structure, to avoid any further understand.

tcolgate:
@wu-sheng so it would be acceptable if it was a separate module, people want to pip install opencensus-python-propagator-jaeger? The node project is structured like that and I can't say it made working with it any easier. The code is what it is, the license is going to be the same. Putting it in a different git repo seem like excessive maintenance burden.

What is the actual objection? Because it seemed to be very much about polluting the code base with the name of a taxi firm. Shifting the boxes around doesn't make the word go away. We can force the user to pass the word as an argument, but that seems rather silly.

adriancole:
More important objection is that this is a temporary thing requested to be
added into the main codebase. in zipkin even amazon propagation which has
far wider impact is in its own area due to coupling concerns. Same with
stackdriver incidentally, though I know the history of this codebase could
make a viable exception.

If you solve the coupling concerns such that this can be deleted later and
folks here don't have to maintain it if the case occurs you are no longer
personally interested, sounds fine to me.

Usually having multiple diverse folks buying into a change is a norm. Right
now, we have one person very interested in this and who feels that alone is
sufficient. Having some rules could be helpful.  Here's an example
https://github.com/openzipkin/brave/blob/master/HACKING.md

tcolgate:
I don't get the coupling concern. This is as decoupled as it is ever going to be. 
If you want to get rid of it you delete two files a drop the requirements.
What other area possibly solves the concern? Directory,? Repo? Org?

Granted it is "temporary" all be it with no obvious w3c adoption anywhere yet.

As it stands I cannot use oc-python at all as there is no propagation mechanism that works for me (this, or b3 at a stretch) . 

I get maintenance burden. I did the easiest thing for me to maintain. I was intending doing likewise for node and Go.

Clearly it's easier to just move this elsewhere.

tcolgate:
I've moved this to a separate repository. I'm not sure how it makes sense to support jaeger as an exporter but not the default/native propagation format. The spec clearly says implementation may support other propagation formats. The hostility to this is frankly, a bit odd.

rakyll:
> I can publish it separately if the general association with jaeger/uber is so objectionable.

Two cents: Jaeger documents this as its canonical header, we should provide support for this or we should not claim we have support for Jaeger.

I think there is an agreement that closed-source backend support should not live in the core repos but having a vendor name in the header name is not the same.

rakyll:
Btw, I wish we can discuss these type of issues outside of the language issue trackers. Many contributors are not watching this repo and we cannot participate into the discussion.

tcolgate:
@rakyll I'm hoping to get this integrated with some of our production services next week which should mean getting some integration testing between traefik, this and the node services. I'll re-raise this PR once I've ironed out any issues.

It would be good to have some clear indications/guidelines on which issues should be on the per-language repos and which should go somewhere else (e.g. spec?), some kind of "escalation" path I guess.

bogdandrutu:
My 2 cents here:
1) OpenCensus library should support new formats easily.
2) If this format is legacy and they intend to remove it I think we should not have it in the "core" library because we try to offer backwards compatibility and follow the deprecation rules strictly. I think we can put this into a contrib directory/repository and say that we are not intended to support this when Jagger supports the w3c standard.
3) Also I know that Jagger supports b3 format https://github.com/jaegertracing/jaeger-client-go#zipkin-http-b3-compatible-header-propagation

Do you still need this PR?

tcolgate:
We're keeping this out of tree at present. If there's a more appropriate contrib location, it can live there once I've hammered any remaining bugs out.
https://github.com/QubitProducts/qubit-opencensus

PR #223, opened 2018-07-09
Sanic middleware

tcolgate:
This is work in progress. Provides a middleware styled after the flask
middleware. Not testeed at all at preset, but some open issue I don't know
how to deal with exist already:

- some of the requirements of sanic don't seem to work on 2.7
- I'm geting a failure on protobuf when attempting to run this
  ouside of the tests.

Sanic doesn't have all the of the request testing options that flask has, so
I need work out what to do there, in addition, the mock tracer isn't getting any
spans (I'm not sure why at present).

### DIFF
--- a/opencensus/trace/asyncio_context.py
+++ b/opencensus/trace/asyncio_context.py
@@ -1,0 +1,57 @@
+# Copyright 2017, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import aiotask_context as context
+import uuid
+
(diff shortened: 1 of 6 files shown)


tcolgate:
The current approach here wont work as sanic is apparently async, so the thread local storage used in exectition_context wont work. I'm investigating further, may need to add async_context?

tcolgate:
I've made some progress, but it's not obvious to me what the correct API should be.
I think we need different implementations of context, the existing execution_contextion does threads, and the alternative if asyncio context. I think these should be selectable when instantiating the context_tracer, but that seems cumbersome. I'm not a day-to-day python developer, if anyone has any ideas, I'd appreciate them.

tcolgate:
I've got all the bits hooked up, but mostly a proof of concept at the moment (sanic -> context spans -> aiohttp works). I'm really only just getting my head around the internal structure of the code,

tcolgate:
I'll close this for now since some asyncio support is needed to build this on. If I can get something sane looking, I'll raise a separate PR for that.

AyWa:
@tcolgate your code was helpful (and https://github.com/QubitProducts/qubit-opencensus) 

From my understanding `asyncio` is now supported right ? might be a good time to bring back `aiohttp` support and `sanic` support ?

tcolgate:
I didn't have a great experience with a few other OC contributions, and what I have works for me right now, so I'm in no hurry on this.

AyWa:
> I didn't have a great experience with a few other OC contributions, and what I have works for me right now, so I'm in no hurry on this.

Okay thx for your reply. I might try to do a contrib if I have time

PR #232, opened 2018-07-12
[WIP] feature: asyncio support

tcolgate:
This branch attempt to implement some asyncio support. The existing thread based execution context fails in programs using asyncio (per thread storage is shared amount asyncio context spans are mis-associated based on the thread code happens to run in)

This branch:
- adds basic support for asyncio
- adds an example middleware for sanic
- add example "integrations" to monkeypatch aiohttp and aioredis
- adds a simple example to show how to set things up and play with the overall API

I should probably confess that this is my first exposure to asyncio in python, and I'm not python programmer to begin with. The basics of this do seem to work for now (error handling needs tidying up).

Current concerns. it feels wrong that the sanic middleware setups up the initial tracer, rather than have it passed in, but passing it in doesn't work well either.

I've raised this mostly to get API feedback. It may be that you don't want the sanic/aiohttp/aioredis parts, but having a standard way of setting you tracing in asyncio probably does make sense to have in the core.

### DIFF
--- a/examples/trace/helloworld-async/main.py
+++ b/examples/trace/helloworld-async/main.py
@@ -1,0 +1,70 @@
+# This is just here for me to test
+
+from sanic import Sanic
+from sanic.response import json
+from opencensus.trace.exporters import jaeger_exporter
+from opencensus.trace.ext.aiohttp.trace import trace_integration as aiohttp_integration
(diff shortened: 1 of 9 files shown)


tcolgate:
I'll continue working on this out of tree:
https://github.com/QubitProducts/qubit-opencensus

PR #236, opened 2018-07-17
fix: jaeger trace_id conversion only uses 64 bits

tcolgate:


tcolgate:
The existing code splits the incoming trace id into two 32 bit chunks, rather than two 64 bit chunks.

PR #238, opened 2018-07-20
implement W3C tracestate (part)

reyang:
Add W3C tracestate class and test cases.
Inline documents will be added in subsequent commits.

### DIFF
--- a/examples/stats/helloworld/main.py
+++ b/examples/stats/helloworld/main.py
@@ -60,7 +60,10 @@ def main():
 
     # Get aggregated stats and print it to console.
     view_data = view_manager.get_view(VIDEO_SIZE_VIEW_NAME)
-    pprint(vars(view_data))   # TODO: print more meaningful info for view_data
+    pprint(vars(view_data))
+    for k, v in view_data._tag_value_aggregation_map.items():
+      pprint(vars(v))
+      pprint(vars(v.aggregation_data))
 
 
 if __name__ == '__main__':
--- a/opencensus/stats/aggregation.py
+++ b/opencensus/stats/aggregation.py
@@ -75,6 +75,7 @@ class SumAggregation(BaseAggregation):
         super(SumAggregation, self).__init__(aggregation_type=aggregation_type)
         self._sum = aggregation_data.SumAggregationDataFloat(
             sum_data=float(sum or 0))
+        self.aggregation_data = self._sum
 
     @property
     def sum(self):
@@ -97,6 +98,7 @@ class CountAggregation(BaseAggregation):
         super(CountAggregation, self).__init__(
             aggregation_type=aggregation_type)
         self._count = aggregation_data.CountAggregationData(count)
+        self.aggregation_data = self._count
 
     @property
     def count(self):
@@ -128,6 +130,8 @@ class DistributionAggregation(BaseAggregation):
             buckets=boundaries, aggregation_type=aggregation_type)
         self._boundaries = bucket_boundaries.BucketBoundaries(boundaries)
         self._distribution = distribution or {}
+        self.aggregation_data = aggregation_data.DistributionAggregationData(
+            0, 0, 0, 0, 0, None, boundaries)
 
     @property
     def boundaries(self):
--- a/opencensus/stats/aggregation_data.py
+++ b/opencensus/stats/aggregation_data.py
@@ -107,14 +107,22 @@ class DistributionAggregationData(BaseAggregationData):
                  min_,
                  max_,
                  sum_of_sqd_deviations,
-                 counts_per_bucket,
-                 bounds):
+                 counts_per_bucket=None,
+                 bounds=None):
         super(DistributionAggregationData, self).__init__(mean_data)
         self._mean_data = mean_data
         self._count_data = count_data
         self._min = min_
         self._max = max_
         self._sum_of_sqd_deviations = sum_of_sqd_deviations
+        if bounds is None:
+            bounds = []
+
+        if counts_per_bucket is None:
+            counts_per_bucket = []
+            bucket_size = len(bounds) + 1
+            for i in range(bucket_size):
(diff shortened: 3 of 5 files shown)


reyang:
Please help to trigger another round of CI, looks like a transient failure in system test.

bogdandrutu:
Done. Also added @SergeyKanzhelev to the organization members so he can do the same. Also filed an issue about this https://github.com/census-instrumentation/opencensus-python/issues/239

opencensus/trace/trace_state.py
```diff
@@ -0,0 +1,98 @@
...
+    # CAN decide to forward the tracestate
+    def is_valid(self):
+        if len(self) is 0:
+            return False
+        # combined header length MUST be less than or equal to 512 bytes
+        if len(self.to_string()) > 512:
+            return False
+        # there can be a maximum of 128 list-members in a list
+        if len(self) > 128:  # pragma: NO COVER
+            # it is impossible to pass the 512 length test and reach here ...
```
reyang: I'll change 128 to 32 in the next PR (according to Bogdan's change: https://github.com/w3c/distributed-tracing/commit/17026e7836c6a08d537a2ebe65b22c290ec1e32f)

reyang: @bogdandrutu would you grant me permission to trigger CI? Looks like the transient failure happened again. I can also look into the CI improvement if there is a need.

SergeyKanzhelev: @reyang I triggered it for you

SergeyKanzhelev: yes, I think there is a need =)

bogdandrutu:
I have a general comment that may be useful. In OpenCensus we try to separate the propagation format from the in-process representation in order to allow possibly multiple serialization formats (see propagation directory).

@SergeyKanzhelev @reyang do you think this is useful to do? In Java I tried to do that as well here https://github.com/census-instrumentation/opencensus-java/pull/1300

opencensus/trace/trace_state.py
```diff
@@ -0,0 +1,61 @@
...
+            raise ValueError('key must be an instance of str')
+        if not re.match(_KEY_VALIDATION_RE, key):
+            raise ValueError('illegal key provided')
+        if not isinstance(value, str):
+            raise ValueError('value must be an instance of str')
+        if not re.match(_VALUE_VALIDATION_RE, value):
+            raise ValueError('illegal value provided')
+        super(TraceState, self).__setitem__(key, value)
+
+    def append(self, key, value):
```
bogdandrutu: It looks like the Tracestate is mutable and this may cause problems (not sure in python with their threading schema): if possible I would prefer to have the Tracestate immutable because once we create it and put in the the SpanContext we don't want to be changed. Scenarios like: create a span with a tracestate then modify the tracestate should not happen in my opinion.

opencensus/trace/trace_state.py
```diff
@@ -0,0 +1,61 @@
...
+import re
+
+_KEY_FORMAT = r'[a-z][_0-9a-z\-\*\/]{0,255}'
+_VALUE_FORMAT = r'[\x20-\x2b\x2d-\x3c\x3e-\x7e]{1,256}'
+
+_KEY_VALIDATION_RE = re.compile('^' + _KEY_FORMAT + '$')
+_VALUE_VALIDATION_RE = re.compile('^' + _VALUE_FORMAT + '$')
+
+
+class TraceState(OrderedDict):
```
bogdandrutu: @adriancole suggested to name this Tracestate to be consistent with the w3c standard. I don't have any strong opinion on this but we should be consistent between libraries (so if we use one name in OpenCensus we should be consistent in all languages, I have not submitted my PR so happy to change if this is the decision).

@SergeyKanzhelev any preferences here between Tracestate vs TraceState?

opencensus/trace/trace_state.py
reyang: This is contrary to my understanding, I was thinking tracestate should allow changes at any point. Would you elaborate?

For example:

with tracer.span('A'):
    if foo():
        current_span.span_context.tracestate['system_x_tracking_id'] = '123'
        invoke_system_x()
    else:
        current_span.span_context.tracestate['system_y_tracking_id'] = '456'
        invoke_system_y()

opencensus/trace/trace_state.py
reyang: Either one works for me, as long as it is **consistent** across specs and different language implementations.

SergeyKanzhelev: `Tracestate` is preferable

opencensus/trace/trace_state.py
SergeyKanzhelev: You should be able to scope `tracestate` properties downstream. Like for hierarchical identifiers in MS and skywalking. But modifying upstream sounds like a bad idea. So I think you both are correct. (note, I haven't looked into PR itself yet)

opencensus/trace/trace_state.py
reyang: Changed to Tracestate.
I've also renamed the file (e.g. trace_state.py to tracestate.py) as we're treating tracestate as a single word.

opencensus/trace/propagation/tracestate_string_format.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2017, OpenCensus Authors
```
bogdandrutu: use year 2018

bogdandrutu:
@reyang is it possible to make the Tracestate immutable?

nikmd23:
@bogdandrutu, do you have a moment to fill in

> Would you elaborate why you prefer immutable? I was originally considering making tracestate mutable + copy-on-write to balance the flexibility vs. performance.

It'd be super helpful to drive this conversation forward.

Issue #210, opened 2018-06-26
wrapt dependency still missing


MikeRoetgers:
Hello,

I know this was already discussed in https://github.com/census-instrumentation/opencensus-python/issues/142. Could someone add the `wrapt` dependency to the next release? :slightly_smiling_face: 
Just picked up opencensus and stumbled over this instantly. The warning unfortunately isn't really prepared to handle this case either (`Failed to integrate module: requests, supported integrations are httplib, mysql, postgresql, pymysql, requests, sqlalchemy, google_cloud_clientlibs`).

Thank you!
Mike

olance:
Allow me to bump this – the original issue (#142) is going to be a year old very soon and is just about correctly including a dependency with the package.

Meanwhile, It's biting 100% of your lib's users on first use

c24t:
I added this in #432. You have to install the package as `opencensus[requests]` if you want the optional dependencies for requests since we don't want users that aren't using requests to install dependencies they don't need.

We may move each integration into its own distribution (see #445), which would solve the optional dependency problem; you'd install `opencensus-requests` separately instead.

olance:
@c24t thanks and sorry for the noise!
Having separate packages for the different intégrations would make total sense :)

reyang:
Now we have the wrapt dependency here https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-requests/setup.py#L43.

The requests module is in its own distribution, which can be installed using `pip install opencensus-ext-requests`.

PR #242, opened 2018-07-30
Log a more useful message when an integration fails to load

benjaminjkraft:
The error was being swallowed, which made it more difficult than
necessary to understand what was wrong when I ran into #210.

### DIFF
--- a/opencensus/trace/config_integration.py
+++ b/opencensus/trace/config_integration.py
@@ -37,11 +37,15 @@ def trace_integrations(integrations, tracer=None):
             module = importlib.import_module(path_to_module)
             module.trace_integration(tracer=tracer)
             integrated.append(item)
-        except Exception:
-            log.warning(
-                'Failed to integrate module: {}, supported integrations are {}'
-                .format(
-                    item,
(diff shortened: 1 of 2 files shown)


reyang:
@benjaminjkraft `config_integration` has been refactored in https://github.com/census-instrumentation/opencensus-python/commit/26fd836a8ae4e38493ef70862ab2baee6ced2339. I believe by logging the module namespace and the actual exception/callstack, we should have sufficient information for troubleshooting.

I would like to get your feedback, do you think we should provide more info?
I will close this PR for now, please feel free to reopen and work on the improvements. I can help to review and merge it.

@c24t for awareness.

benjaminjkraft:
Oops, I never realized this hadn't been merged.  Anything that logs the error is good enough for us!

reyang:
@benjaminjkraft would you look at https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/config_integration.py#L36 and confirm if this is good for your scenario?

benjaminjkraft:
Yeah that seems fine!

reyang:
Thanks @benjaminjkraft!

PR #247, opened 2018-08-02
fix ZipkinExporter when using BackgroundThreadTransport

felippe-mendonca:
I was using ZipkingExporter with the default transport (SyncTransport) having no problem, but, when I started to use the BackgroundThreadTransport I'd notice that groups of traces were exported with same trace_id. The code below creates this behaviour:

```python
import time
from opencensus.trace.span_context import SpanContext
from opencensus.trace.exporters.zipkin_exporter import ZipkinExporter

exporter = ZipkinExporter(
  host_name='localhost',
  port=9411,
  transport=BackgroundThreadTransport
)

for i in range(3):
  tracer = Tracer(exporter=exporter)
  name = 'my_span_{}'.format(i)
  with tracer.span(name=name) as span:
    time.sleep(0.5)
```
Two traces were generate (see below). This happened because the `_WAIT_PERIOD` is equal 1 second, thus two traces were accumulated before being exported. (since the trace duration is 500ms).

1)
```json
[
  {
    "traceId": "b6a2ac1d66734bd78f74d083a68e290f",
    "id": "6741ecffda664a47",
    "name": "my_span_0",
    "timestamp": 1533202058197415,
    "duration": 500138,
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "",
        "endpoint": {
          "serviceName": "my_service",
          "port": 9411
        }
      }
    ]
  }
]
```
2)
```json
[
  {
    "traceId": "65a2c37b014348dcaa4673c137f0ce0e",
    "id": "79e1d239ee7f4650",
    "name": "my_span_1",
    "timestamp": 1533202058698123,
    "duration": 500340,
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "",
        "endpoint": {
          "serviceName": "my_service",
          "port": 9411
        }
      }
    ]
  },
  {
    "traceId": "65a2c37b014348dcaa4673c137f0ce0e",
    "id": "d41fe77ffe704436",
    "name": "my_span_2",
    "timestamp": 1533202059198777,
    "duration": 500611,
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "",
        "endpoint": {
          "serviceName": "my_service",
          "port": 9411
        }
      }
    ]
  }
]
```

After this modification, I got three separate traces as it should be.
So, I don't know if this was the expected behaviour, but I guess that isn't :)


adriancole:
note:

usually best to use v2 data to discuss change since that is what this
exporter sends. ex if you have latest zipkin server json button can show v2
json but it can also be retrieved at /api/v2/trace/your_id

PR #248, opened 2018-08-02
[WIP] Tags refactoring

zeako:
Fix tag spec requirements, also refactor some behaviors.

TagMap expects tag related objects

### DIFF
--- a/opencensus/tags/tag.py
+++ b/opencensus/tags/tag.py
@@ -26,16 +26,7 @@ class Tag(object):
     :param value: Value of the key in the tag.
 
     """
+
     def __init__(self, key, value):
-        self._key = tag_key.TagKey(key)
-        self._value = tag_value.TagValue(value)
-
-    @property
-    def key(self):
-        """The key of the current tag"""
-        return self._key
-
-    @property
-    def value(self):
-        """The value of the key in the current tag"""
-        return self._value
+        self.key = tag_key.TagKey(key)
+        self.value = tag_value.TagValue(value)
--- a/opencensus/tags/tag_key.py
+++ b/opencensus/tags/tag_key.py
@@ -20,15 +20,12 @@ class TagKey(object):
     :param name: The name of the key
 
     """
+
     def __init__(self, name):
-        self._name = name
+        self.name = name
 
-    @property
-    def name(self):
-        """The name of the current key"""
-        return self._name
-
-    def is_valid_name(self, name):
+    @staticmethod
+    def is_valid_name(name):
         """Checks if the name of the key is valid
 
         :type name: str
@@ -37,11 +34,8 @@ class TagKey(object):
         :rtype: bool
         :returns: True if it valid, else returns False
         """
-        if (len(name) > 0) and (len(name) <= 255):
-            if (all(ord(char) < 126 for char in name) and
-                    all(ord(char) > 32 for char in name)):
-                return True
-            else:
-                return False
+
+        if 0 < len(name) < 256:
+            return all(32 <= ord(char) <= 126 for char in name)
         else:
             return False
--- a/opencensus/tags/tag_key.py
+++ b/opencensus/tags/tag_key.py
@@ -11,6 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from opencensus.tags.validation import is_legal_chars
+
+_TAG_NAME_ERROR = 'tag name must not be empty, no longer than 255 characters and of ascii values between 32 - 126'
 
 
 class TagKey(object):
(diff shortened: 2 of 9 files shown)


zeako:
Take note that there are some breaking changes regarding how `TagMap` works, trying to align its consistency.

haven't modified yet anything that depends on tags.

opencensus/tags/tag_key.py
```diff
@@ -12,36 +12,23 @@
...
-    :param name: The name of the key
 
-    """
-    def __init__(self, name):
-        self._name = name
+class TagKey(str):
+    """A tag key with a property name"""
 
-    @property
-    def name(self):
```
songy23: Why remove this?

opencensus/tags/tag_map.py
```diff
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import OrderedDict
```
songy23: `tag_map` doesn't need to be ordered.

opencensus/tags/tag_map.py
```diff
@@ -20,71 +25,69 @@ class TagMap(object):
...
-        return self._map
+    def __iter__(self):
+        return self.map.items().__iter__()
 
     def insert(self, key, value):
         """Inserts a key and value in the map if the map does not already
         contain the key.
 
-        :type key: :class: '~opencensus.tags.tag_key.TagKey'
+        :type key: Union[str, :class: '~opencensus.tags.tag_key.TagKey']
```
songy23: I think it's better to keep `key` as `TagKey` and `value` as `TagValue`, to be consistent with other methods in this class.

opencensus/tags/tag_value.py
```diff
@@ -12,37 +12,23 @@
...
-    :param value: A string representing the value of a key in a tag
 
-    """
-    def __init__(self, value):
-        self._value = value
+class TagValue(str):
+    """The value of a tag"""
 
-    @property
-    def value(self):
```
songy23: Same, why remove this?

opencensus/tags/tag.py
```diff
@@ -12,30 +12,19 @@
...
+    :type value: '~opencensus.tags.tag_key.TagValue'
     :param value: Value of the key in the tag.
 
     """
-    def __init__(self, key, value):
-        self._key = tag_key.TagKey(key)
-        self._value = tag_value.TagValue(value)
-
-    @property
-    def key(self):
```
songy23: Same here.

opencensus/tags/tag.py
```diff
@@ -12,30 +12,19 @@
...
-        """The key of the current tag"""
-        return self._key
-
-    @property
-    def value(self):
```
songy23: ditto

tests/unit/tags/propagation/test_binary_serializer.py
```diff
@@ -43,12 +44,9 @@ def test_from_byte_array(self):
 
     def test_to_byte_array(self):
         from opencensus.tags.tag_map import TagMap
-        from collections import OrderedDict
 
-        tag_map = OrderedDict(
-            [('key1', 'val1'), ('key2', 'val2'),
-             ('key3', 'val3'), ('key4', 'val4')])
-        tags = [tag_map]
+        tags = [Tag('key1', 'val1'), Tag('key2', 'val2'),
```
songy23: I think `key` and `value` should be `TagKey` and `TagValue`, rather than strings?

tests/unit/tags/test_tag.py
```diff
@@ -13,15 +13,15 @@
...
+from opencensus.tags import Tag
 
 
 class TestTag(unittest.TestCase):
 
     def test_constructor(self):
         key = 'key1'
         value = 'value1'
-        tag = tag_module.Tag(key=key, value=value)
+        tag = Tag(key=key, value=value)
```
songy23: ditto

opencensus/tags/tag_key.py
zeako: Seems like an overkill.
Using `property` was unnecessary, since we can just use `name` by itself.

also, in other implementations eventually `TagKey` is an object with string value, so inheriting from `str` enables us all its benefits while keeping it simple.

can keep the `name` field for consistency across languages thought it's not specified in the specs

opencensus/tags/tag_map.py
zeako: `dict` broke `test_to_byte_array` which relied on specific order.
though from python 3.6 dict is ordered by design.
https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6

thoughts?

opencensus/tags/tag_value.py
zeako: Same comment for `TagKey`

opencensus/tags/tag.py
zeako: Simplifies implementation

opencensus/tags/tag.py
zeako: Same for previous comment

tests/unit/tags/propagation/test_binary_serializer.py
zeako: Can do to keep it consistent across other implementations

opencensus/tags/tag_map.py
songy23: I see, then I think it's fine to leave it as a OrderedDict to be consistent with the language spec.

songy23:
Any update on this PR?

PR #250, opened 2018-08-03
Stackdriver stats exporter - WIP

MarceloAquino7:
- Stackdriver stats exporter methods
- Stackdriver stats exporter tests
- Base class for future stats exporters
- few changes inside stats core - WIP


songy23:
Please sign the CLA.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,415 @@
...
+		self._trace_client_options = trace_client_opt
+		self._bundle_delay_threshold = bundle_delay_thrs
+		self._bundle_count_threshold = bundle_count_thrs
+		self._resource = resource
+		self._metric_prefix = metric_prefix
+		self._default_trace_attributes = default_trace_attributes
+		self._default_monitoring_labels = default_monitoring_labels
+
+	@property
+	def project_id(self):
```
zeako: just note that there's no need for `@property` decorator if you don't add any extra logic when setting/getting it.
it isn't really pythonic to do so in this situation.
also you should just name the attributes without the underscore.
for more info:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/implementing_java-style_getters_and_setters.html?highlight=java

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,415 @@
...
+	def default_labels(self):
+		return self._default_labels
+
+	def set_default_labels(self, value):
+		self._default_labels = value
+
+	def on_register_view(self, view):
+		""" create metric descriptor for the registered view  """
+		if view:
+			err = self.create_measure(view)
```
zeako: in python doing a `try except` clause is the way to go.
it isn't idiomatic to return errors such as it is in `Golang`

also, semi-related:
https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,415 @@
...
+				description = view.description,
+				unit = unit,
+				labels = new_label_descriptors(self.default_labels, view.columns),
+				name = "projects/%s/metricDescriptors/%s" % (self.options.project_id, metric_type),
+				display_name = "%s/%s" % (display_name_prefix, view_name),
+			)
+			project_name = self.client.project_path(self.options.project_id)
+			descriptor = self.client.create_metric_descriptor(project_name, descriptor)
+			self.created_views[view_name] = descriptor.name
+		except Exception as identifier:
```
zeako: try to narrow to specific exceptions

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,415 @@
...
+from opencensus.stats import measure
+
+MAX_TIME_SERIES_PER_UPLOAD = 200
+OPENCENSUS_TASK_DESCRIPTION = "Opencensus task identifier"
+DEFAULT_DISPLAY_NAME_PREFIX = "OpenCensus"
+ERROR_BLANK_PROJECT_ID = "expecting a non-blank ProjectID"
+CONS_NAME = "name"
+CONS_TIME_SERIES = "timeseries"
+
+class Options(object):
```
zeako: if `Options` is to be an immutable class you can just define it as a `namedtuple`.
it acts is a data-class without the boilerplate.
https://pymotw.com/2/collections/namedtuple.html

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: Hi @zeako , thanks for your suggestions.

About the attributes names, what do you suggest ?

I created them with underscore because this is the way that some attributes were created inside view_data, for example.

zeako: just use their names as-is, no underscore needed or `@property` declarations.

if you also think `namedtuple` can be used instead of just a class, then go for it (though keep in mind that it'll make `Options` immutable, which isn't a bad thing by itself).
and if you want to add functions to it you can always extend it with a class

MarceloAquino7:
CLA signed

MarceloAquino7:
CLA issue test

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		project_name = self.client.project_path(self.options.project_id)
+		descriptor = self.client.create_metric_descriptor(project_name, descriptor)
+
+def new_stats_exporter(options):
+	""" new_stats_exporter returns an exporter that uploads stats data to Stackdriver Monitoring.
+	 Only one Stackdriver exporter should be created per ProjectID per process, any subsequent
+	 invocations of NewExporter with the same ProjectID will return an error.
+	"""
+	if str(options.project_id).strip() == "":
+		return None, ERROR_BLANK_PROJECT_ID
```
zeako: your return pattern is inconsistent, at the end you don't return `None` for error.
and as stated before, this pattern of returning errors isn't the way to go in python, please use exceptions.

MarceloAquino7: Thank you for your suggestion. Fixed !

songy23:
Please follow the same steps in https://github.com/census-instrumentation/opencensus-python/pull/256#issuecomment-412948886 to fix the CLA.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		as well as error handling. 
+		"""
+		if view_datas:
+			self.upload_stats(view_datas)
+		
+	def upload_stats(self, view_datas):
+		""" It receives an array of view_data object 
+			and create time series for each value
+		"""
+		for request in self.make_request(view_datas,
```
odeke-em: This is a bit of a weird indentation, how about?
```python
requests = self.make_request(view_data, MAX_TIME_SERIES_PER_UPLOAD)
for request in requests:
    self.client.create_time_series(request[CONS_NAME], request[CONS_TIME_SERIES])
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+	def make_request(self, view_datas, limit):
+		""" Create the data structure that will be 
+			sent to Stackdriver Monitoring
+		"""
+		requests = []
+		time_series = []
+
+		resource = self.options.resource
+		for view_data in view_datas:
```
odeke-em: s/datas/data/g as datas isn't a plural for data :)

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+					point.value.distribution_value.bucket_options.explicit_buckets.bounds.extend(list(map(float,agg_data.bounds)))
+					point.value.distribution_value.bucket_counts.extend(list(map(int,agg_data.counts_per_bucket)))
+				elif type(tag_value.value) is int:
+					point.value.int64_value = int(tag_value.value)
+				elif type(tag_value.value) is float:
+					point.value.float_value = float(tag_value.value)
+				elif type(tag_value.value) is str:
+					point.value.string_value = str(tag_value.value)
+
+				start = datetime.datetime.strptime(view_data.start_time,"%Y-%m-%dT%H:%M:%S.%fZ")
```
odeke-em: I wonder why this conversion is necessary, from the definition of `view_data.start_time`, `view_data.end_time`
https://github.com/census-instrumentation/opencensus-python/blob/2bc3745ef61719db90f091a62897c802b5762331/opencensus/stats/view_data.py#L25-L29

both seem to be a datetime.datetime object.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+				end = datetime.datetime.strptime(view_data.end_time,"%Y-%m-%dT%H:%M:%S.%fZ")
+
+				timestamp_start = (start - datetime.datetime(1970, 1, 1)).total_seconds()
```
odeke-em: I think we could create a static variable at the beginning of the class definition, say
```python
EPOCH_DATETIME = datetime.datetime(1970, 1, 1)

def ...:
    timestamp_star = (view_data.end_time - EPOCH_DATETIME).total_seconds()
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+	@property
+	def client(self):
+		return self._client
+
+	@property
+	def default_labels(self):
+		return self._default_labels
+
+	def set_default_labels(self, value):
+			self._default_labels = value
```
liyanhui1228: Shall we change it to be adding a key/value pair to the labels dict instead of updating the whole labels?

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+	def on_register_view(self, view):
+		""" create metric descriptor for the registered view  """
+		if view:
```
liyanhui1228: Use `if view is not None` to be explicit.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+			self.create_metric_descriptor(view)
+
+	def emit(self, view_datas):
+		""" export data to Stackdriver Monitoring """
+		if view_datas:
```
liyanhui1228: Same here.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+				point.interval.end_time.nanos = int((timestamp_end - point.interval.end_time.seconds) * 10**9)
+
+				#
+				# Uncomment this when LastValue gets supported
+				#
+				#if not agg.aggregation_type is aggregation.Type.LASTVALUE:
+				if timestamp_start == timestamp_end: # avoiding start_time and end_time to be equal
+					timestamp_start = timestamp_start - 1
+				point.interval.start_time.seconds = int(timestamp_start)
+				point.interval.start_time.nanos = int((timestamp_start - point.interval.end_time.seconds) * 10**9)
```
odeke-em: This looks wrong and will result in a negative number, as `point.interval.end_time.seconds` will ALWAYS be greater than `timestamp_start`. Perhaps you meant to say
```python
ns = int((timestart - point.interval.start_time.seconds) * 1e9)
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+		resource = self.options.resource
+		for view_data in view_datas:
+			series = monitoring_v3.types.TimeSeries()
+			series.metric.type = namespaced_view_name(view_data.view.name)
+
+			if not resource:
+    				series.resource.type = 'global'
+			else:
+				series.resource = resource
```
liyanhui1228: Move this part to a separate function called `def create_time_series(...args needed here)` would be better.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+				point.interval.start_time.nanos = int((timestamp_start - point.interval.end_time.seconds) * 10**9)
+
+			time_series.append(series)
+
+			request = {}
+			request[CONS_NAME] = self.client.project_path(self.options.project_id)
+			request[CONS_TIME_SERIES] = time_series
+			requests.append(request)
+
+			if len(time_series) == int(limit):
```
odeke-em:
```python
if len(time_series) >= int(limit):
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+    				series.resource.type = 'global'
+			else:
+				series.resource = resource
+
+			for tag_value, agg in view_data.tag_value_aggregation_map.items():
+				point = series.points.add()
+				if agg.aggregation_type is aggregation.Type.DISTRIBUTION:
+					agg_data = view_data.tag_value_aggregation_map.get(tag_value).aggregation_data
+					
+					point.value.distribution_value.count = agg_data.count_data
```
liyanhui1228: Line is too long here, probably could:
```
dist_value = point.value.distribution_value
dist_value.count = agg_data.count_data
...
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		value_type = None
+		unit = view_measure.unit
+		# Default metric Kind
+		metric_kind = monitoring_v3.enums.MetricDescriptor.MetricKind.CUMULATIVE
+
+		if view_aggregation.aggregation_type is aggregation.Type.COUNT:
+			value_type = monitoring_v3.enums.MetricDescriptor.ValueType.INT64
+			# If the aggregation type is count, which counts the number of recorded measurements, 
+			# the unit must be "1", because this view does not apply to the recorded values.
+			unit = 1
```
odeke-em: Isn't this supposed to be the string "1" instead of the int 1? The comment above it, also seems to allude to "1"

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		elif view_aggregation.aggregation_type is aggregation.Type.SUM:
+			if view_measure is measure.MeasureInt:
+				value_type = monitoring_v3.enums.MetricDescriptor.ValueType.INT64
+			elif view_measure is measure.MeasureFloat:
+				value_type = monitoring_v3.enums.MetricDescriptor.ValueType.DOUBLE
+		elif view_aggregation.aggregation_type is aggregation.Type.DISTRIBUTION:
+			value_type = monitoring_v3.enums.MetricDescriptor.ValueType.DISTRIBUTION
+		#
+		# Aggreagation type last value is not currently supported by opencesus python stats
```
odeke-em: s/ggreagation/ggregation/g

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+			name = "projects/%s/metricDescriptors/%s" % (self.options.project_id, metric_type),
+			display_name = "%s/%s" % (display_name_prefix, view_name),
+		)
+		project_name = self.client.project_path(self.options.project_id)
+		descriptor = self.client.create_metric_descriptor(project_name, descriptor)
+
+def new_stats_exporter(options):
+	""" new_stats_exporter returns an exporter that uploads stats data to Stackdriver Monitoring.
+	 Only one Stackdriver exporter should be created per ProjectID per process, any subsequent
+	 invocations of NewExporter with the same ProjectID will return an error.
```
odeke-em: I don't see that enforced anywhere.

opencensus/stats/exporters/stackdriver_exporter.py
liyanhui1228: Do we really need to minus `datetime.datetime(1970, 1, 1)` to get the timestamp?

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		# Default metric Kind
+		metric_kind = monitoring_v3.enums.MetricDescriptor.MetricKind.CUMULATIVE
+
+		if view_aggregation.aggregation_type is aggregation.Type.COUNT:
+			value_type = monitoring_v3.enums.MetricDescriptor.ValueType.INT64
+			# If the aggregation type is count, which counts the number of recorded measurements, 
+			# the unit must be "1", because this view does not apply to the recorded values.
+			unit = 1
+		elif view_aggregation.aggregation_type is aggregation.Type.SUM:
+			if view_measure is measure.MeasureInt:
```
liyanhui1228: `isinstance(view_measure, measure.MeasureInt)`

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+	return exporter
+
+def get_task_value():
+	""" getTaskValue returns a task label value in the format of
+	 "py-<pid>@<hostname>". 
+	"""
+	hostname = platform.uname()[1]
+	if not hostname:
+		hostname = "localhost"
+	return "py" + str(os.getpid()) + hostname
```
odeke-em: Missing a dash, this can be made clearer and easier to spot if we did
```python
return "py-{pid}@{hostname}".format(pid=os.getpid(), hostname=hostname)
```

opencensus/stats/measure_to_view_map.py
```diff
@@ -33,11 +33,18 @@ def __init__(self):
         self._registered_measures = {}
         # stores the set of the exported views
         self._exported_views = set()
+        # stores the array of the registered exporters
```
odeke-em:
```python
# Stores the registered exporters
```

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+		label_descriptors.append(label)
+	
+	for tag_key in keys:
+		label = {}
+		label["key"] = remove_non_alphanumeric(tag_key.name)
+		label_descriptors.append(label)
+	return label_descriptors
+
+def remove_non_alphanumeric(text):
+	""" Remove characters not accepted in labels key
```
liyanhui1228: Is it documented somewhere for the label key format? Maybe put into the doc of this function is helpful.

opencensus/stats/exporters/stackdriver_exporter.py
odeke-em: @liyanhui1228 yeah, Python's datetime.datetime is one hell of a beast unfortunately. Some folks recommend using the '%s' specifier with strftime but that produces a string which then has to be converted into a float, PTAL at https://stackoverflow.com/questions/11743019/convert-python-datetime-to-epoch-with-strftime

README.rst
```diff
@@ -371,6 +375,67 @@ You can enable Requests integration by specifying ``'requests'`` to ``trace_inte
...
+
+        from opencensus.stats.exporters import stackdriver_exporter as stackdriver
+        from opencensus.stats import stats as stats_module
+
+Prerequisites
+*************
+
+- OpenCensus Python libraries require Python 2.7 or later.
+- Google Cloud Platform account and project.
+- Google Stackdriver Tracing enabled on your project (Need help? `Click here`_).
```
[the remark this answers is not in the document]
ocervell: Replace 'Stackdriver Tracing' by 'Stackdriver Monitoring' since that's the API we're talking about here.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+import datetime
+import platform
+from . import base
+from google.cloud import monitoring_v3
+from opencensus.stats import view_data
+from opencensus.stats import aggregation
+from opencensus.stats import aggregation_data
+from opencensus.stats import stats
+from opencensus.stats import measure
+# Add here the import for transport class
```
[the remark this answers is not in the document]
MarceloAquino7: Yes, that is correct.

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: Hello Emmanuel, thanks for you reply.

During the development of this exporter, I tried to use these two attributes but the Stackdriver Monitoring API raised an error saying that both values were in wrong type format. Do you think that there is a better approach ?

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: I think that you are correct, however I did not found any example using the start() from view_data, because only start() and end() will create two different timestamps. If the user do not use them, start_time and end_time will be the same value. I'm going to fix this, thank you.

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: I don't think that is possible to reach above the limit, because when the array reach the limit value, we set an empty array.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+	def emit(self, view_datas):
+		""" export data to Stackdriver Monitoring """
+		if view_datas:
+			self.handle_upload(view_datas)
+
+	def export(self, view_datas):
+		""" export data to transport class """
+		if view_datas:
+			pass #self.transport.export(view_datas) || Uncomment this when transport class gets merged to this branch
```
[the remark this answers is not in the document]
MarceloAquino7: We are waiting for the stats core PR to be merged, because this commented code needs the new transport class

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: We are using methods to fill this field, so I don't know if does make sense to create a def to add one by one.

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: if we do not remove invalid characters, the API returns something like:
"google.api_core.exceptions.InvalidArgument: 400 Field metricDescriptor.labels[0].key had an invalid value of "py@62916GARCIA-3.local": Label key contains invalid characters."

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: I think it is just a comment error, because it really should be a number.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,357 @@
...
+        label = {}
+        label["key"] = remove_non_alphanumeric(tag_key.name)
+        label_descriptors.append(label)
+    return label_descriptors
+
+
+def remove_non_alphanumeric(text):
+    """ Remove characters not accepted in labels key
+    """
+    return str(re.sub('[^0-9a-zA-Z ]+', '', text)).replace(" ","")
```
zeako: I think that this is redundant, tags refactoring was merged to master and legal checks are being done at creation

songy23:
Please fix the CLA, rebase against `census-instrumentation:master` and resolve the merge conflicts.

opencensus/stats/exporters/stackdriver_exporter.py
vcasadei: Done

README.rst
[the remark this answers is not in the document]
vcasadei: Done

README.rst
```diff
@@ -371,6 +375,67 @@ You can enable Requests integration by specifying ``'requests'`` to ``trace_inte
...
 .. _Requests package: https://pypi.python.org/pypi/requests
 
+------
+ Stats
+------
+
+Stackdriver
+-----------
+
+The OpenCensus Stackdriver Stats Exporter for Python is a package that allows the user exports data to `Stackdriver`_. The API of this project is still evolving. The use of vendoring or a dependency management tool is recommended.
```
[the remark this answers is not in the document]
vcasadei: Done

opencensus/stats/exporters/stackdriver_exporter.py
vcasadei: Done

opencensus/stats/measure_to_view_map.py
vcasadei: Done

vcasadei:
I signed it!

songy23:
Ping @googlebot

opencensus/stats/exporters/stackdriver_exporter.py
liyanhui1228: @MarceloAquino7 Please take a look at here, we should clean up the code if the check is redundant.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+    return label_descriptors
+
+
+def remove_non_alphanumeric(text):
+    """ Remove characters not accepted in labels key
+    """
+    return str(re.sub('[^0-9a-zA-Z ]+', '', text)).replace(" ", "")
+
+
+def is_float(value):
```
odeke-em: This is a weird helper whose usage is quite awkward but also the body and the return arguments. `is_float` implies just checking if a value is a float not conversion as well. Let's also change the order of arguments returned

I think we want to make this something like
```python
def as_float(value):
     """ Converts a value to a float if possible
          On success, it returns (converted_value, True)
          On failure, it returns (None, False)
     """
     try:
          return float(value), True
      except: # Catch all exception including ValueError
          return None, False
```

and then we can use it like this
```python
convFloat, isFloat = as_float(tag_value[0])
if isFloat:
   point.value.double_value = convFloat
```

instead of
```python
 if is_float(tag_value[0])[0]:
    point.value.double_value = float(tag_value[0])
```
which unfortunately again runs the conversion but also for which the if condition isn't readable.

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: @zeako and @liyanhui1228 thanks for your reply. I think that we should not remove this function because in tag_value and tag_key creation it checks for ASCII values between 32 and 126. This range contains a few characters not accepted by Stackdriver Monitoring API for label's key.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+            start = datetime.strptime(v_data.start_time, EPOCH_PATTERN)
+            end = datetime.strptime(v_data.end_time, EPOCH_PATTERN)
+
+            timestamp_start = (start - EPOCH_DATETIME).total_seconds()
+            timestamp_end = (end - EPOCH_DATETIME).total_seconds()
+
+            point.interval.end_time.seconds = int(timestamp_end)
+
+            secs = point.interval.end_time.seconds
+            point.interval.end_time.nanos = int((timestamp_end-secs)*10**9)
```
odeke-em: Minor nit, we can express `10**9` as `1e9` which will always keep it as a constant.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+class StackdriverStatsExporter(base.StatsExporter):
+    """ StackdriverStatsExporter exports stats
+    to the Stackdriver Monitoring."""
+    def __init__(self,
+                 options=Options(),
+                 client=None,
+                 default_labels={},
+                 transport=sync.SyncTransport):
```
odeke-em: Could we please make the default we want is an AsyncTransport? Currently we have SyncTransport as the default
and it has caused a bit of unexpectedness where people perceive the libraries as being slow yet it is just the SyncTransport.

MarceloAquino7: Thanks for your reply. Done !

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: Thanks for your reply. Done !

opencensus/stats/exporters/stackdriver_exporter.py
MarceloAquino7: Because of the nox lint, I was not allowed to use just except:, but instead I created using except Exception as exception:

Issue #231, opened 2018-07-12
Stats: ViewData.tag_value_aggregation_map is not updated correctly


songy23:
The key for `ViewData.tag_value_aggregation_map` should be tuples of `TagValue`s, rather than individual `TagValue`: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/stats/view_data.py#L104-L108.

PR #253, opened 2018-08-09
Fix bug in updating aggregation map.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/231.

1. Remove `_tag_map` and `get_tag_map` from `view_data`. `view_data.record()` should always use the given context.
2. The key for `ViewData.tag_value_aggregation_map` should be tuples of `TagValue`s, rather than individual `TagValue`.
3. When creating a new entry in `ViewData.tag_value_aggregation_map`, use `copy.deepcopy()` to create a blank `AggregationData`.
4. Use tag_value_aggregation_data_map instead of tag_value_aggregation_map.


songy23:
Thanks!

Issue #254, opened 2018-08-09
Stats: Add an API to record measurements against implicit context.


songy23:
See [OpenCensus-Specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Record.md#recording-stats):
> Users should record Measurements against a context, either an explicit context or the implicit current context. 

Currently there's only an API that records measurements against explicit context: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/stats/measurement_map.py#L50-L57.

PR #258, opened 2018-08-14
Stats: Allow recording against implicit tag_map.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/254.


songy23:
Thanks!

PR #265, opened 2018-08-20
Implement B3 propagation

kornholi:
Supersedes #168.

This works for our usecase (in-app spans for requests in an Istio mesh, exported directly to Stackdriver), but it's not correct since ParentSpanId is not propagated correctly.

See https://github.com/openzipkin/b3-propagation/blob/682d1fc9caa31fa25678674a3a7a1ecd4e9f7813/README.md#why-is-parentspanid-propagated

Should `to_headers` take a parent `SpanContext`, or should the parent be included in `SpanContext`?

### DIFF
--- a/opencensus/trace/propagation/b3_format.py
+++ b/opencensus/trace/propagation/b3_format.py
@@ -16,6 +16,7 @@
 from opencensus.trace.span_context import SpanContext, INVALID_SPAN_ID
 from opencensus.trace.trace_options import TraceOptions
 
+_STATE_HEADER_KEY = 'b3'
 _TRACE_ID_KEY = 'x-b3-traceid'
 _SPAN_ID_KEY = 'x-b3-spanid'
 _SAMPLED_KEY = 'x-b3-sampled'
@@ -39,20 +40,42 @@ class B3FormatPropagator(object):
         if headers is None:
             return SpanContext(from_header=False)
 
-        trace_id = headers.get(_TRACE_ID_KEY)
-        span_id = headers.get(_SPAN_ID_KEY)
-        sampled = headers.get(_SAMPLED_KEY) == '1'
+        trace_id, span_id, sampled = None, None, None
+        trace_options = TraceOptions()
 
-        # Trace Id and Span Id headers both have to exist
+        state = headers.get(_STATE_HEADER_KEY)
+        if state:
+            fields = state.split('-', 4)
+
+            if len(fields) == 1:
+                sampled = fields[0]
+            elif len(fields) == 2:
+                trace_id, span_id = fields
+            elif len(fields) == 3:
+                trace_id, span_id, sampled = fields
+            elif len(fields) == 4:
+                trace_id, span_id, sampled, _parent_span_id = fields
+            else:
+                return SpanContext(from_header=False)
+        else:
+            trace_id = headers.get(_TRACE_ID_KEY)
+            span_id = headers.get(_SPAN_ID_KEY)
+            sampled = headers.get(_SAMPLED_KEY)
+
+        if sampled is not None:
+            if len(sampled) != 1:
+                return SpanContext(from_header=False)
+
+            sampled = sampled in ('1', 'd')
+            trace_options.set_enabled(sampled)
+
+        # TraceId and SpanId headers both have to exist
         if not trace_id or not span_id:
-            return SpanContext()
+            return SpanContext(trace_options=trace_options)
 
         # Convert 64-bit trace ids to 128-bit
         if len(trace_id) == 16:
             trace_id = '0'*16 + trace_id
-
-        trace_options = TraceOptions()
-        trace_options.set_enabled(sampled)
 
         span_context = SpanContext(
             trace_id=trace_id,
--- a/opencensus/trace/propagation/b3_format.py
+++ b/opencensus/trace/propagation/b3_format.py
@@ -1,4 +1,4 @@
-# Copyright 2017, OpenCensus Authors
+# Copyright 2018, OpenCensus Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/opencensus/trace/propagation/b3_format.py
+++ b/opencensus/trace/propagation/b3_format.py
@@ -41,7 +41,6 @@ class B3FormatPropagator(object):
             return SpanContext(from_header=False)
 
         trace_id, span_id, sampled = None, None, None
-        trace_options = TraceOptions()
 
         state = headers.get(_STATE_HEADER_KEY)
         if state:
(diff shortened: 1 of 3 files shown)


kornholi:
@songy23 @liyanhui1228 ptal when you have a chance

liyanhui1228:
Can we also combine the code in #168 to this PR to let the flask, django, pyramid integrations be able to recognize the B3 format?

songy23:
> Should to_headers take a parent SpanContext, or should the parent be included in SpanContext?

Just took a look at the Java implementation, looks like ParentSpanId should not be propagated: https://github.com/census-instrumentation/opencensus-java/pull/889/files#r155878293.

/cc @adriancole

adriancole:
parent should be propagated definitely if RPC. We have done some work to
lower impact when not set but that doesn't mean it should not be attempted.

messaging can safely leave out parent.

x-b3-flags means force trace and really only ever set to 1 (header left out
otherwise). if there is no force/debug trace feature main thing is just
propagating what if any was incoming

opencensus/trace/propagation/b3_format.py
```diff
@@ -0,0 +1,89 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from opencensus.trace.span_context import SpanContext, INVALID_SPAN_ID
+from opencensus.trace.trace_options import TraceOptions
+
+_TRACE_ID_KEY = 'x-b3-traceid'
```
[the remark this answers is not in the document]
kornholi: Header names are case-insensitive (and forced lower case in HTTP/2)

kornholi:
> Can we also combine the code in #168 to this PR to let the flask, django, pyramid integrations be able to recognize the B3 format? Or we can put that part in a separate PR. This PR LGTM.

It already works because of @reyang's refactoring. Just have to set `PROPAGATOR` to `B3Propagator` in `OPENCENSUS_TRACE` config.

> x-b3-flags means force trace and really only ever set to 1 (header left out
otherwise). if there is no force/debug trace feature main thing is just
propagating what if any was incoming

Right, the issue here is that our propagators basically act as (de)serializers for `SpanContext`, and `SpanContext` currently does not encode that information. I'm trying to figure out the best way to refactor this.

reyang:
@kornholi regarding "SpanContext currently does not encode that information", would you try putting the info in SpanContext.tracestate (which is an ordered-key-value-pair intended for [extensibility](https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#tracestate-field))?

adriancole:
the b3 tracestate entry is imho done, but waiting a few more nods
https://github.com/openzipkin/b3-propagation/issues/21

kornholi:
I think this is ready to be merged.

Regarding ParentSpanId and Debug headers, the other implementations strip them away so we should do the same here. See https://github.com/census-instrumentation/opencensus-go/blob/264a2a48d94c062252389fffbc308ba555e35166/plugin/ochttp/propagation/b3/b3.go#L35-L43 etc. It's still unclear to me why we're piggybacking on the B3 name if it's not actually B3 (split client/server spans).

opencensus/trace/propagation/b3_format.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class B3FormatPropagator(object):
+    """Propagator for processing the B3 HTTP header format.
+
+    See: https://github.com/openzipkin/b3-propagation
+    """
+
+    def from_headers(self, headers):
+        """Generate a SpanContext object using the trace context headers.
```
reyang: The comment has "trace context headers" while the code is using Zipkin headers.
Perhaps we should remove these excessive comments? I don't find them quite useful as I can tell what the function is doing from its name "from_headers".

kornholi: I agree. However they are short enough so I just clarified the B3 part. We need to set some guidelines as right now the comment quality/quantity varies greatly.

opencensus/trace/propagation/b3_format.py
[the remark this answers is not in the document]
adriancole: most send X-B3 in the case format you mention even in case sensitive contexts. the new "b3" header should always be lowercase.

opencensus/trace/propagation/b3_format.py
```diff
@@ -0,0 +1,88 @@
...
+
+        :rtype: :class:`~opencensus.trace.span_context.SpanContext`
+        :returns: SpanContext generated from B3 propagation headers.
+        """
+        if headers is None:
+            return SpanContext(from_header=False)
+
+        trace_id = headers.get(_TRACE_ID_KEY)
+        span_id = headers.get(_SPAN_ID_KEY)
+        sampled = headers.get(_SAMPLED_KEY) == '1'
```
adriancole: the header can be absent. when it is it should be sampled using whatever your heuristic is not set to unsampled. make sense?

opencensus/trace/propagation/b3_format.py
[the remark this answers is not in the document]
zoidbergwill: Maybe it'd be worth adding support for the new [`b3` single header](https://github.com/openzipkin/b3-propagation/blob/master/README.md#single-header) in this PR too?

adriancole:
if it is still open (looks the case), then yeah probably worth adding b3 single indeed

akoskaaa:
Hey, I'd love this to get merged. What is required exactly?

opencensus/trace/propagation/b3_format.py
zoidbergwill: From what I understand this will work if it's missing, because later we call `.enabled or .should_sample` [here](https://sourcegraph.com/github.com/census-instrumentation/opencensus-python@c84e87f/-/blob/opencensus/trace/tracer.py#L76-77)

kornholi:
@liyanhui1228 how should we move forward?

c24t:
Hey @kornholi, I'm working with @liyanhui1228 and can take a look at this PR this week.

kornholi:
@c24t ping

c24t:
Sorry to hold you up here, will take a look soon.

reyang:
@kornholi would you help to update the branch? Let's get this merged given the major part looks good. We can track in issues in case there are open questions, we shouldn't get blocked here entirely. (once the branch got updated, we will wait for 24 hours to see if people have strong objections before I go ahead and merge)
Adding @akoskaaa for awareness.

Once this got this PR merged, we can close #168 since it is a duplicated effort. @justinwp

kornholi:
Thanks @reyang, I rebased the branch. I agree on fixing the rest of issues separately, this has been sitting here too long :)

PR #266, opened 2018-08-20
Initial OpenCensusD trace exporter implementation

lmolkova:
This change implements the first version of [OCD exporter](https://github.com/census-instrumentation/opencensus-service) for python.

Open questions (could be addressed in other PRs):

1. Where this code should live. Assuming this exporter will become default one, does it make sense to keep it in separate repo/package?

2. Keeping proto-generated code separately how it's done for Go: https://github.com/census-instrumentation/opencensus-proto/gen-go could be an option, but it seems to complicate packaging and build.

3. Python project structure separates trace and metrics exporter, but exporter proto is the same: https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/exporter/exporter.proto
so it violates python project structure a bit.

4. grpc endpoint is hardcoded for now and it's assumed that the agent is started already

@rakyll @ramonza @bogdandrutu @SergeyKanzhelev 

I'm looking for your early feedback. Tests will follow.


opencensus/trace/exporters/ocd_exporter.py
```diff
@@ -0,0 +1,212 @@
...
+        :type span_datas: list of :class:`~opencensus.trace.span_data.SpanData`
+        :param span_datas: SpanData tuples to convert to protocuf spans and send to OCD agent
+
+        :rtype: list of opencensus.proto.exporter.exporter_pb2.ExportSpanRequests
+        :returns: List of span export requests.
+        """
+
+        for span_data in span_datas:
+            pb_span = self.translate_to_ocd(span_data)
+            yield exporter_pb2.ExportSpanRequest(spans = [pb_span])
```
lmolkova: please advise if there is a better/more efficient way to stream spans

ramonza:
I can take a look through it but I haven't really got a lot of Python experience.

lmolkova:
@ramonza I'd be happy to discuss open question with you (in the description)

songy23:
Please hold off this PR for the moment. We recently made some changes to the OC-D protocol.

lmolkova:
@songy23 could you please share what changes and for how long it should be held off? 
We've (Microsoft) been targeting this feature for the end of Sep.

BTW, what is the OC python release cadence and procedure?

lmolkova:
@songy23 do you mean https://github.com/census-instrumentation/opencensus-proto/pull/74 ?
I'll update everything once the proto change PR is merged, do you think it blocks the review for now?

@bogdandrutu do we have any timeline for  https://github.com/census-instrumentation/opencensus-proto/pull/74 to be merged?

songy23:
I'll share the update design soon. We can move forward with this PR once we agreed upon the design.

Currently @liyanhui1228 is in charge of OC Python releasing.

opencensus/trace/exporters/ocd_exporter.py
```diff
@@ -0,0 +1,217 @@
...
+from google.protobuf.wrappers_pb2 import BoolValue, UInt32Value
+
+from opencensus.trace.exporters import base
+from opencensus.trace.exporters.transports import sync
+from opencensus.trace.exporters.gen.ocd.trace import trace_pb2
+from opencensus.trace.exporters.gen.ocd.exporter import exporter_pb2, exporter_pb2_grpc
+
+DEFAULT_ENDPOINT = 'localhost:50051'
+
+class OcdExporter(base.Exporter):
```
ramonza: I don't think we should use "OCD" as a name anywhere. It seems inappropriate since it's a psychiatric disorder any may cause offense or be hurtful.

lmolkova: did you have any thoughts on how to name it? I think we should be consistent across languages.

adriancole:
```proto3
service Export {
  rpc ExportSpan (stream ExportSpanRequest) returns (stream ExportSpanResponse) {}
  rpc ExportMetrics (stream ExportMetricsRequest) returns (stream ExportMetricsResponse) {}
}
```

the proto rpc is ExportSpan so SpanProtoExporter is an option (to avoid stuttering)

lmolkova:
@songy23 @liyanhui1228 @bogdandrutu @reyang 

Please review. This includes latest proto changes (https://github.com/census-instrumentation/opencensus-proto/pull/79, https://github.com/census-instrumentation/opencensus-proto/pull/84).

Could you please also help with following issue?

At this point, lint fails on auto-generated proto code (opencensus/trace/exporters/gen/opencensusd/*). Is there a way to exclude it from checks? I believe jaeger auto-gen files are also excluded from lint checks.

lmolkova:
@songy23 @liyanhui1228 @bogdandrutu @reyang 

ping

lmolkova:
@songy23 thanks for the review, I've addressed your comments.

opencensus/trace/exporters/span_proto_exporter/span_proto_exporter.py
```diff
@@ -0,0 +1,196 @@
...
+    def emit(self, span_datas):
+        """
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
+        """
+
+        try:
+            responses = self.client.Export(
```
bogdandrutu: I think we want to keep the RPC (stream) alive for the entire time, not just the channel and stub. And the protocol says to send the node once per RPC stream not per channel, I need to investigate how to do this in python, but if we cannot easily do this we should always send the Node.

lmolkova: My understanding was that until grpc is terminated because of error, it is kept alive. But I don't have enough grpc knowledge and can't find detailed documentation for python. I'd appreciate any help with it.

Let's send Node with every request for now and optimize later if possible.

lmolkova:
@bogdandrutu thanks for the review, I addressed all comments.

@liyanhui1228 could you please help with lint and proto-generated code?

opencensus/trace/exporters/ocagent/trace_exporter.py
```diff
@@ -0,0 +1,189 @@
...
+    def emit(self, span_datas):
+        """
+        :type span_datas: list of :class:
+            `~opencensus.trace.span_data.SpanData`
+        :param list of opencensus.trace.span_data.SpanData span_datas:
+            SpanData tuples to emit
+        """
+
+        try:
+            responses = self.client.Export(
```
bogdandrutu: Add a todo to add the optimization to keep the RPC alive. I got the answer from grpc team:
We need to use an iterator (an object that yields messages) similar to this example:
https://github.com/grpc/grpc/blob/master/examples/python/route_guide/route_guide_client.py#L65

No need to do it right now but good to have it documented.

songy23:
Another possible way to pass the lint check is to format the proto generated files, though later we'll need to do it manually every time when those files are re-generated.

reyang:
Can we make a change to https://github.com/census-instrumentation/opencensus-python/blob/master/nox.py#L86 to ignore the generated folder entirely?

FYI - http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html

lmolkova:
@reyang thanks, it worked!

so all green now, @liyanhui1228 @songy23  please merge

PR #271, opened 2018-08-22
Set SpanKind in extensions

lmolkova:
Span kind is optional, but highly recommended flag that denotes the type of span: Server or Client.

It's the only reliable way to determine whether it is an incoming or outgoing request. Some backends like  Azure ApplicationInsights rely on such into to build the UX.


lmolkova:
@songy23 please review

Issue #268, opened 2018-08-21
Http attribute names are not consistent with the spec and some other OC implementations


lmolkova:
Here are attributes that are used in python:
https://github.com/census-instrumentation/opencensus-python/blob/ae727954bd03af9db768ba44721e7436be0b768f/opencensus/trace/attributes_helper.py

```
COMMON_ATTRIBUTES = {
    'AGENT': 'g.co/agent',
    'COMPONENT': '/component',
    'ERROR_MESSAGE': '/error/message',
    'ERROR_NAME': '/error/name',
    'HTTP_CLIENT_CITY': '/http/client_city',
    'HTTP_CLIENT_COUNTRY': '/http/client_country',
    'HTTP_CLIENT_PROTOCOL': '/http/client_protocol',
    'HTTP_CLIENT_REGION': '/http/client_region',
    'HTTP_HOST': '/http/host',
    'HTTP_METHOD': '/http/method',
    'HTTP_REDIRECTED_URL': '/http/redirected_url',
    'HTTP_REQUEST_SIZE': '/http/request/size',
    'HTTP_RESPONSE_SIZE': '/http/response/size',
    'HTTP_STATUS_CODE': '/http/status_code',
    'HTTP_URL': '/http/url',
    'HTTP_USER_AGENT': '/http/user_agent',
    'PID': '/pid',
    'STACKTRACE': '/stacktrace',
    'TID': '/tid',
}
```

https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes

suggests that http attributes should look like `http.host` rather than '/http/host' and exporter is responsible to translate them to the particular backend format.

PR #272, opened 2018-08-23
Rename attributes to the default naming schema and use standard http attributes

### LINKED ISSUE #269: Requests instrumentation does not use standard http attributes
https://github.com/census-instrumentation/opencensus-python/blob/ae727954bd03af9db768ba44721e7436be0b768f/opencensus/trace/ext/requests/trace.py

it uses `requests/url` and `requests/status_code`  attributes.

Here is the list of standard attributes:
https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes

lmolkova:
Fixes #269, #268


opencensus/trace/exporters/stackdriver_exporter.py
```diff
@@ -204,3 +205,47 @@ def translate_to_stackdriver(self, trace):
 
         spans = {'spans': spans_list}
         return spans
+
+    def map_attributes(self, span_data_attributes):
+        if span_data_attributes is None:
+            return span_data_attributes
+
+        for (key, value) in span_data_attributes.items():
```
liyanhui1228: We can update the attribute keys like this:
```
attribute_map = span_data_attributes.get('attributeMap')
for key, value in list(attribute_map.items()):
    new_key = ATTRIBUTE_MAPPING.get(key)
    attribute_map[new_key] = attribute_map.pop(key)
```

liyanhui1228:
@songy23 We have system tests running for django and dbapi integration, so it should be fine after this change as the CI is green. It should also be fine with Pyramid.

PR #274, opened 2018-08-23
Handle unicode and str in exporters for py2

### LINKED ISSUE #273: Postgresql exporter serialization with Python2
Using Python2.7, opencensus installed from GitHub on 3b42664.

When running tracing with the postgresql extension and the jaeger exporter, I get such logs:

```
2018-08-22 14:59:37,872 1 WARNING dev opencensus.trace.exporters.jaeger_exporter: Could not serialize attribute             postgresql/query:SELECT "crm_lead".id FROM "crm_lead" WHERE (("crm_lead"."active" = %s)  AND  (("crm_lead"."type" = %s)  OR  "crm_lead"."type" IS NULL )) ORDER BY "crm_lead"."priority" DESC,"crm_lead"."date_action" ,"crm_lead"."id" DESC  limit 80 to tag
```

A part of the queries is correctly serialized though. The application (odoo) has both sql queries written as py2 str (`''`) and as py2 unicode (`u''`).

guewen:
Before this commit, in py2 only bytes strings were exported and in py3
only unicode strings were exported.

I'm not sure I'm doing it right, that's at least an opening for a discussion.

Fixes #273

### DIFF
--- a/opencensus/trace/exporters/jaeger_exporter.py
+++ b/opencensus/trace/exporters/jaeger_exporter.py
@@ -19,6 +19,7 @@ import datetime
 import logging
 import socket
 
+from six import text_type, string_types
 from thrift.protocol import TBinaryProtocol, TCompactProtocol
 from thrift.transport import THttpClient, TTransport
 
@@ -307,12 +308,14 @@ def _extract_tags(attr):
 
 def _convert_attribute_to_tag(key, attr):
     """Convert the attributes to jaeger tags."""
+    if isinstance(attr, string_types):
+        attr = attr.decode('utf8')
     if isinstance(attr, bool):
         return jaeger.Tag(
             key=key,
             vBool=attr,
             vType=jaeger.TagType.BOOL)
-    if isinstance(attr, str):
+    if isinstance(attr, text_type):
         return jaeger.Tag(
             key=key,
             vStr=attr,
--- a/opencensus/trace/exporters/zipkin_exporter.py
+++ b/opencensus/trace/exporters/zipkin_exporter.py
@@ -21,6 +21,8 @@ import logging
 
 import requests
 
+from six import text_type, string_types
+
 from opencensus.trace.exporters import base
 from opencensus.trace.exporters.transports import sync
 from opencensus.trace.utils import check_str_length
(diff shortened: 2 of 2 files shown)


guewen:
> I signed it!

Hey, I signed the CLA :)

guewen:
I know the tests don't pass on py3 but before refining I'd like a validation that I understood correctly the goal and I'm not heading in the wrong direction.

c24t:
Thanks for the PR @guewen. Using the `six` types to solve the problem of exporting `unicode` attribute values looks right to me.

In general though this problem is a big can of worms, and this PR makes it clear that we need to be more careful about internal use of the `str` type.

If I understand correctly: it looks like the library assumes string-valued attribute values are always `str`s. This is usually a safe assumption, but it means that we can't store non-ASCII characters. This is a problem for code that naively uses `unicode`s in place of `str`s since we'll silently fail to export these attributes.

So in python 2.x, `str`s are ASCII-encoded byte strings. Decoding a byte string with any valid encoding gets you a `unicode`... which is not a `str`:

```python
>>> type(b'abc')
str

>>> type(b'abc'.decode('utf-8'))
unicode

>>> isinstance(b'abc'.decode('utf-8'), str)
False
```

And in python 3.x `str`s are effectively python 2.x's `unicode`s, and byte strings are demoted to `bytes` with no implicit encoding:

```python
>>> type(b'abc')
bytes

>>> type(b'abc'.decode('utf-8'))
str

```

We have to support both versions of python, and have to support non-ASCII characters in attribute values. But the spec also says to [truncate these strings to 256 bytes](https://github.com/census-instrumentation/opencensus-proto/blob/6d79e46232e026136ae013489ef611ca721cb839/src/opencensus/proto/trace/v1/trace.proto#L323-L324) without specifying an encoding.

In 2/3 decoding a byte string with any valid encoding gets you a `unicode`/`str`, which is itself stored internally as unicode, using [up to 4 bytes per character depending on the python implementation](https://stackoverflow.com/questions/26079392/how-is-unicode-represented-internally-in-python#comment40863091_26079426). Among other problems, this means that we might truncate a 265 character string down to 64 characters even if it's possible to encode it with ASCII. This is a moot point now since it doesn't look like we're actually truncating these strings, but does suggest we have to be careful making changes like this that add `decode` calls where byte strings would otherwise stay byte strings.

c24t:
Which is all to say: the direction looks good, but there may be some unintended consequences.

guewen:
Thanks for your detailed answer, particularly, I wasn't aware of the 256 bytes truncation (new to the subject).

Issue #135, opened 2018-03-08
Error when using Zipkin Exporter


bshaffer:
I get the following error when using the Zipkin exporter:

```
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded in __instancecheck__
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded in cmp
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded while calling a Python object
ERROR:root:maximum recursion depth exceeded
ERROR:root:maximum recursion depth exceeded
```

Here is my setup code:

```py
    zipkin_host = os.environ.get('ZIPKIN_PORT_9411_TCP_ADDR')
    exporter = zipkin_exporter.ZipkinExporter(
        host_name=zipkin_host)
    spanCtx = tracer_module.SpanContext(trace_id=str(traceId))
    tracer = tracer_module.Tracer(exporter=exporter, span_context=spanCtx)
    config_integration.trace_integrations(['google_cloud_clientlibs'])
```

bshaffer:
The output continues like this indefinitely until I force quit the process. The result is 10,000+ spans.

liyanhui1228:
This is fixed in PR #133.

nlamirault:
@liyanhui1228 it is possible to make a new release with this bugfix ?

SEJeff:
[...]

PR #289, opened 2018-08-29
Don't trace if it's a http request made by the exporter

geobeau:
So the issue is that when we trace httplib we also trace calls made by the exporter which make an infinite loop. The issue was reported in #135. This fix should prevent tracing when we see the same host for the request.


opencensus/trace/ext/httplib/trace.py
```diff
@@ -61,6 +61,13 @@ def wrap_httplib_request(request_func):
 
     def call(self, method, url, body, headers, *args, **kwargs):
         _tracer = execution_context.get_opencensus_tracer()
+        try:  # pragma: NO COVER
+            # Don't trace if it's a http request made by the exporter
+            if self._dns_host == _tracer.exporter.host_name:
```
bogdandrutu: Probably we can make the mechanism more generic and allow httplib to accept a list of paths + hosts to not be traced. I think this is reasonable request.

@lmolkova @SergeyKanzhelev what do you think?

geobeau: Good idea. To note: the request integration has the same issues, so it would need something similar (I can implement it as well in the same PR as soon as we aggree on this part)

geobeau: Any leads on how to implement that? I don't think integrations are currently using options/parameters, so I'm not sure what would be the best way to do it.

liyanhui1228: We have a blacklist url check method here: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/utils.py#L40

And in Flask and Django integration, we allow users to provide a list of blacklisted URLs, and we will disable tracing for the blacklisted ones. Probably in here we could reuse that method.

In the exporters we actually didn't define `exporter.host_name` in the base class, could you add the param if we need it for every exporter?

lmolkova: host_name is 'localhost' for several exporters by default: zipkin, jaeger, proto. Tracing calls to localhost could be legit and useful and we should prevent tracking based on something more unique like 'host:port'.

lmolkova: I also wonder how this list should be shared between different  instrumentations: requests, httplib, grpc and who is responsible for configuring it? 

My assumption: each exporter should register endpoints in this list. Otherwise users are responsible for finding out all endpoints for exporters they use and registering them in all relevant instrumentations.

geobeau:
Open for another review. I changed it to be more configurable and expanded it to `requests`.

opencensus/trace/ext/utils.py
```diff
@@ -63,3 +65,31 @@ def disable_tracing_url(url, blacklist_paths=None):
             return True
 
     return False
+
+
+def disable_tracing_hostname(url, blacklist_hostnames=None):
```
liyanhui1228: Can this be combined with the `disable_tracing_url` function in this file?

geobeau: I am not sure. They both disable tracing but one is working on paths (blacklist path starting with an element from the list) and the other one with hostname (blacklist hostname/port exactly matching) with different default value. I think it's nice to have them separated for readability.

geobeau: @liyanhui1228 Do you stand on your position on this?

liyanhui1228: It's fine to separate them :)

geobeau:
Anything blocking merge ? @liyanhui1228

c24t:
Hey @geobeau, I'm working with @liyanhui1228. Unless @bogdandrutu or @lmolkova have any more comments on the diff it LGTM (pending picking up recent changes on master), and I can merge it.

geobeau:
Cool, thanks @c24t !

c24t:
@geobeau there was a failing test after picking up master. Let me know if the fix in 0a30edd looks kosher to you and I'll merge it if so.

geobeau:
Looks good to me

c24t:
See #382 to fix CI timeouts.

c24t:
It looks like async transport worker threads created in (at least) [test_stackdriver_stats.py](https://github.com/census-instrumentation/opencensus-python/blob/dfb039e2b61bfc2ff564b04930b7fea6bfb3340a/tests/unit/stats/exporter/test_stackdriver_stats.py) are each blocking for a few seconds as the test process exits, causing it to hang for long enough for CI to fail.

We're likely seeing this now because I increased the [export interval](https://github.com/census-instrumentation/opencensus-python/blob/dfb039e2b61bfc2ff564b04930b7fea6bfb3340a/opencensus/common/transports/async_.py#L27) in #354 -- lowering the export interval or [grace period](https://github.com/census-instrumentation/opencensus-python/blob/dfb039e2b61bfc2ff564b04930b7fea6bfb3340a/opencensus/common/transports/async_.py#L25) fixes the long wait time for the threads to exit. I don't know why we're only timing out on _this_ PR, and not the others since #354.

Ideally we should be cleaning these threads up as the tests that create them exit, and I think the transport should flush immediately on exit instead of waiting (up to the duration of the grace period) for the next scheduled export. But this is a problem for another PR.

geobeau:
Thanks for the fix!

PR #299, opened 2018-09-06
Prometheus system tests

vcasadei:


tests/system/stats/prometheus/prometheus_stats_test.py
```diff
@@ -0,0 +1,72 @@
...
+        tag_value = tag_value_module.TagValue(str(random.randint(1, 10000)))
+        tag_map = tag_map_module.TagMap()
+        tag_map.insert(FRONTEND_KEY, tag_value)
+        measure_map = stats_recorder.new_measurement_map()
+        measure_map.measure_int_put(VIDEO_SIZE_MEASURE, 25 * MiB)
+        measure_map.record(tag_map)
+
+        if sys.version_info > (3, 0):
+            import urllib.request
+            contents = urllib.request.urlopen("http://localhost:9202/metrics").read()
```
liyanhui1228: We will need to run the prometheus server on Circle CI environment in order to run the system tests. Probably use the `prom/prometheus` docker image. The config for pulling and running the image will be similar to [here](https://github.com/census-instrumentation/opencensus-python/blob/master/.circleci/config.yml#L9).

vcasadei: @liyanhui1228 there is no need to run prometheus docker server, because the prometheus client creates an http server. Therefore, we need to validate if the stats information is being correctly passed to the client only, and not if it is being served to the prometheus server.

The server is kind of a web scrawler that searches for the http server that's created by the client. As the client is developed by Prometheus, we are assured that the data will be available to the server.

That's what we talked about on the last meeting about this. Do you agree?

PR #300, opened 2018-09-07
Enable agent TraceExporter configuration in django and flask

lmolkova:
This change allows to use TraceExporter for ocagent with django and flask.

SERVICE_NAME setting is added with the intention to deprecate ZIPKIN_EXPORTER_SERVICE_NAME: if SERVICE_NAME is not present, we check for ZIPKIN_EXPORTER_SERVICE_NAME

### DIFF
--- a/opencensus/trace/ext/django/middleware.py
+++ b/opencensus/trace/ext/django/middleware.py
@@ -38,11 +38,12 @@ BLACKLIST_PATHS = 'BLACKLIST_PATHS'
 GCP_EXPORTER_PROJECT = 'GCP_EXPORTER_PROJECT'
 SAMPLING_RATE = 'SAMPLING_RATE'
 TRANSPORT = 'TRANSPORT'
+SERVICE_NAME = 'SERVICE_NAME'
 ZIPKIN_EXPORTER_SERVICE_NAME = 'ZIPKIN_EXPORTER_SERVICE_NAME'
 ZIPKIN_EXPORTER_HOST_NAME = 'ZIPKIN_EXPORTER_HOST_NAME'
 ZIPKIN_EXPORTER_PORT = 'ZIPKIN_EXPORTER_PORT'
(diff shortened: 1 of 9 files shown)


opencensus/trace/ext/flask/flask_middleware.py
```diff
@@ -237,3 +247,13 @@ def _teardown_request(self, exception):
...
         except Exception:  # pragma: NO COVER
             log.error('Failed to trace request', exc_info=True)
+
+    def _get_service_name(self, params):
+        _service_name = params.get(
+            SERVICE_NAME, None)
+
+        if _service_name is None:
+            _service_name = params.get(
+                ZIPKIN_EXPORTER_SERVICE_NAME, 'my_service')
```
songy23: ditto

opencensus/trace/ext/django/middleware.py
```diff
@@ -216,3 +224,13 @@ def process_response(self, request, response):
...
         finally:
             return response
+
+    def _get_service_name(self, params):
+        _service_name = params.get(
+            SERVICE_NAME, None)
+
+        if _service_name is None:
+            _service_name = params.get(
+                ZIPKIN_EXPORTER_SERVICE_NAME, 'my_service')
```
songy23: If we want to send Zipkin spans to OC-Agent, they need to be converted to OC span proto first. Is it the case in Django?

If that's not the case, please do not add Zipkin as a service endpoint for now. We plan to add an interceptor to accept Zipkin spans in OC-Agent.

lmolkova: this is not about sending the spans, it's just a django/flask app setting for the service name on the exporter.
There was a setting ZIPKIN_EXPORTER_SERVICE_NAME, but I want to generalize it and rename to SERVICE_NAME

songy23: Got it, thanks for the explaination.

PR #303, opened 2018-09-07
jaeger exporter: check whether time_event.annotation is None

meridional:
When I tried to use grpc intercepter and jaeger exporter, I encountered this error:

`AttributeError: 'NoneType' object has no attribute 'attributes'`

This can be reproduced by simply substituting the stackdriver exporter of the `examples/trace/grpc/hello_world_server.py` to jaeger exporter.


meridional:
cla check.

PR #305, opened 2018-09-10
Stackdriver end to end tests

vcasadei:


tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,118 @@
...
+                                        [FRONTEND_KEY],
+                                        VIDEO_SIZE_MEASURE,
+                                        VIDEO_SIZE_DISTRIBUTION)
+
+        stats = stats_module.Stats()
+        view_manager = stats.view_manager
+        stats_recorder = stats.stats_recorder
+
+        client = monitoring_v3.MetricServiceClient()
+        exporter = stackdriver.StackdriverStatsExporter(options=stackdriver.Options(project_id="opencenus-node"),
```
liyanhui1228: Why using `opencensus-node` project id?

liyanhui1228:
@vcasadei Thanks for this PR! I have just configured the credentials to give it permission to write to stackdriver monitoring, the CI is running and the system tests for stackdriver stats exporter is passed. And I also patched the monitoring client in the stackdriver unit test to prevent it from talking to the real monitoring client. Ping me if you need anything:)

tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,121 @@
...
+        # Process video.
+        # Record the processed video size.
+        tag_value = tag_value_module.TagValue("1200")
+        tag_map = tag_map_module.TagMap()
+        tag_map.insert(FRONTEND_KEY_ASYNC, tag_value)
+        measure_map = stats_recorder.new_measurement_map()
+        measure_map.measure_int_put(VIDEO_SIZE_MEASURE_ASYNC, 25 * MiB)
+
+        measure_map.record(tag_map)
+        self.assertTrue(True)
```
liyanhui1228: This isn't test anything, maybe assert something meaningful here?

tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,121 @@
...
+        measure_map.record(tag_map)
+
+
+        # Sleep for [0, 10] milliseconds to fake wait.
+        time.sleep(random.randint(1, 10) / 1000.0)
+
+        name = exporter.client.project_path(PROJECT)
+        list_metrics_descriptors = exporter.client.list_metric_descriptors(name)
+        element = [element for element in list_metrics_descriptors if element.description == "SampleViewDescriptionSync"]
+        self.assertTrue(any(element))
```
liyanhui1228: Printing out the element it looks like
```
[name: "projects/[PROJECT_ID]/metricDescriptors/custom.googleapis.com/opencensus/SampleViewNameSync"
labels {
  key: "SampleKeySync"
}
metric_kind: CUMULATIVE
value_type: DISTRIBUTION
unit: "By"
description: "SampleViewDescriptionSync"
display_name: "OpenCensus/SampleViewNameSync"
type: "custom.googleapis.com/opencensus/SampleViewNameSync"
]
```
Can we assert each of the attributes equals what we expected?

And we will also need to remove the exported metrics after running the test, because next time even if the exporter fails, the test will still pass as it is using the previous data. Or we can randomly generate some unique name for the metric to make sure the one we assert is the one we just exported.

tests/system/stats/stackdriver/stackdriver_stats_test.py
MarceloAquino7: In this case, there is an issue because we are testing in an async way. So, due to our async transport ( that wait for something like 5 seconds to send data ) we could not read it from Stackdriver api to verify if data was sent in a correct way.

tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+        # Sleep for [0, 10] milliseconds to fake wait.
+        time.sleep(random.randint(1, 10) / 1000.0)
+
+        name = exporter.client.project_path(PROJECT)
+        list_metrics_descriptors = exporter.client.list_metric_descriptors(name)
+        element = next((element for element in list_metrics_descriptors if element.description == view_description), None)
+        
+        self.assertNotEqual(element, None)
```
liyanhui1228: `self.assertIsNotNone(element)`

tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,153 @@
...
+        measure_map = stats_recorder.new_measurement_map()
+        measure_map.measure_int_put(VIDEO_SIZE_MEASURE_ASYNC, 25 * MiB)
+
+        measure_map.record(tag_map)
+        
+        name = exporter.client.project_path(PROJECT)
+        list_metrics_descriptors = exporter.client.list_metric_descriptors(name)
+        element = next((element for element in list_metrics_descriptors if element.description == view_description), None)
+        
+        self.assertNotEqual(element, None)
```
liyanhui1228: Same here.

tests/system/stats/stackdriver/stackdriver_stats_test.py
```diff
@@ -0,0 +1,153 @@
...
+        measure_map.measure_int_put(VIDEO_SIZE_MEASURE, 25 * MiB)
+
+        measure_map.record(tag_map)
+
+
+        # Sleep for [0, 10] milliseconds to fake wait.
+        time.sleep(random.randint(1, 10) / 1000.0)
+
+        name = exporter.client.project_path(PROJECT)
+        list_metrics_descriptors = exporter.client.list_metric_descriptors(name)
```
liyanhui1228: Retry logic should be added here instead of adding to the whole test, we want to wait for the metric creation and we don't want to create the metric again when failing.

liyanhui1228:
@songy23 Could you help us fix the CLA?

songy23:
Sure, should be good to merge now.

PR #309, opened 2018-09-13
update the doc with TraceContextPropagator

reyang:


README.rst
```diff
@@ -169,22 +169,26 @@ Propagators
 ~~~~~~~~~~~
 
 You can specify the propagator type for serializing and deserializing the
-``SpanContext`` and its headers. There are currently two built in propagators:
-``GoogleCloudFormatPropagator`` and ``TextFormatPropagator``.
+``SpanContext`` and its headers. There are currently three built in propagators:
+``GoogleCloudFormatPropagator``, ``TextFormatPropagator`` and ``TraceContextPropagator``.
 
-This example shows how to use the ``GoogleCloudFormatPropagator``:
+This example shows how to use the ``TraceContextPropagator``:
```
songy23: Instead of removing the example for `GoogleCloudFormatPropagator`, can you leave the original example as-is and add another example on `TraceContextPropagator` below?

reyang: Sure I'll do that.

Issue #312, opened 2018-09-17
AttributeError: __exit__ and __enter__ are missing in BlankSpan class. 


mayurkale22:
The magic methods ```__exit__``` and ```__enter__``` are missing in BlankSpan class. 
```
from opencensus.trace import Span

with Span(name="foo") as s:
    print(s.name)

foo
```
BUT

```
from opencensus.trace.blank_span import BlankSpan 

with BlankSpan(name="foo") as s:
    print(s.name)

AttributeError: __exit__
```

One way to guarantee that both Span and BlankSpan are compatible is to create a common interface that they both implement. This is how it is done for tracers for instance.

mayurkale22:
@tgermain Thanks for informing us.

mayurkale22:
Closed in #313

PR #313, opened 2018-09-18
Add BaseSpan interface to make Span and BlankSpan compatible

mayurkale22:
issue #312

### DIFF
--- a/opencensus/trace/blank_span.py
+++ b/opencensus/trace/blank_span.py
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from opencensus.trace import base_span
 from opencensus.trace import time_event as time_event_module
 from opencensus.trace.span_context import generate_span_id
 from opencensus.trace.tracers import base
-from opencensus.trace.base_span import BaseSpan
 
 
-class BlankSpan(BaseSpan):
+class BlankSpan(base_span.BaseSpan):
     """A BlankSpan is an individual timed event which forms a node of the trace
     tree. All operations are no-op.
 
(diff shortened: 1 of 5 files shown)


tests/unit/trace/test_base_span.py
```diff
@@ -0,0 +1,82 @@
...
+        span = BaseSpan()
+
+        with self.assertRaises(NotImplementedError):
+            span.add_link(None)
+
+    def test_iter_abstract(self):
+        span = BaseSpan()
+
+        with self.assertRaises(NotImplementedError):
+            list(iter(span))
```
tgermain: a test for `__enter__` and `__exit__` might be missing. Something like that for instance :
```python
from unittest import mock

@mock.patch(BaseSpan, "__enter__")
@mock.patch(BaseSpan, "__exit__")
def test_context_manager(self, m_exit, m_enter)
    span = BaseSpan()
    with span:
        pass
    self.assertTrue(m_enter.called)
    self.assertTrue(m_exit.called)
```

opencensus/trace/blank_span.py
```diff
@@ -15,9 +15,10 @@
 from opencensus.trace import time_event as time_event_module
 from opencensus.trace.span_context import generate_span_id
 from opencensus.trace.tracers import base
+from opencensus.trace.base_span import BaseSpan
```
liyanhui1228: Sort the imports, and import module instead of class.

tests/unit/trace/test_base_span.py
mayurkale22: Done

opencensus/trace/blank_span.py
mayurkale22: Done

mayurkale22:
Fixes #312

davewalkexpel:
@liyanhui1228 Hi, can a release be cut that included this? It would really help me. Thanks!

mayurkale22:
@davewalkexpel Unfortunately no. But we might do a patch release soon.

PR #315, opened 2018-09-18
Allow arbitrary metric prefixes

### LINKED ISSUE #311: Allow arbitrary metric prefixes
Today OpenCensus Python only supports custom.googleapis.com domain prefix but Stackdriver supports more prefixes (and it's expanding). OC should not enforce metric prefix validation (There will never be a way to statically know the set of prefixes), we should allow users to report metrics for their registered domain.

https://github.com/census-instrumentation/opencensus-python/blob/dfae33d14d48a7a23ef25b8ac886acfcccfe3ee7/opencensus/stats/exporters/stackdriver_exporter.py#L352

mayurkale22:
Fixes #311 

Counterpart in Java : https://github.com/census-instrumentation/opencensus-java/pull/1449


songy23:
@liyanhui1228 Could you help restart the CI? Looks like there's some error with the CI host.

PR #325, opened 2018-09-24
Only use default Jaeger endpoint when no endpoint is defined

achandras:
I think that the default endpoint should only be used when the custom endpoint is empty. Currently, we get something like this:
```
http://.../custom_endpoint/api/traces
```
which seems a bit confusing.


achandras:
Thanks for the approval @reyang

reyang:
@achandras thanks for the contribution!

PR #327, opened 2018-09-24
Exporter/Zipkin: Add annotations

### LINKED ISSUE #323: trace/exporter/{Jaeger, Zipkin}: no annotations showing up in UI
The following piece of code adds annotation to the span, but none are shown in the Zipkin UI. 

```
with tracer.span(name="repl") as span:
        line = sys.stdin.readline()
        span.add_annotation("Invoking processLine", len_=len(line))
        out = processInput(tracer, line)
```

screen shot 2018-09-23 at 23 46 07

Could this be a linking problem on the backend @odeke-em ?  Also, #314 (Jaeger UI) might be similar to this.

mayurkale22:
Fixes #323

### DIFF
--- a/opencensus/trace/exporters/zipkin_exporter.py
+++ b/opencensus/trace/exporters/zipkin_exporter.py
@@ -176,6 +176,7 @@ class ZipkinExporter(base.Exporter):
                 'duration': int(round(duration_ms)),
                 'localEndpoint': local_endpoint,
                 'tags': _extract_tags_from_span(span.attributes),
+                'annotations': _extract_annotations_from_span(span),
             }
 
             span_kind = span.span_kind
@@ -211,3 +212,24 @@ def _extract_tags_from_span(attr):
             continue
         tags[attribute_key] = value
     return tags
+
+
+def _extract_annotations_from_span(span):
+    """Extract and convert time event annotations to zipkin annotations"""
+    if span.time_events is None:
+        return []
+
+    annotations = []
+    for time_event in span.time_events:
+        annotation = time_event.annotation
+        if not annotation:
+            continue
+
+        event_time = datetime.datetime.strptime(time_event.timestamp,
+                                                ISO_DATETIME_REGEX)
+        epoch_time = calendar.timegm(
+            event_time.timetuple()) * 1e6 + event_time.microsecond
+        annotations.append({'timestamp': int(round(epoch_time)),
+                            'value': annotation.description})
+
+    return annotations
--- a/tests/unit/trace/exporters/test_zipkin_exporter.py
+++ b/tests/unit/trace/exporters/test_zipkin_exporter.py
@@ -15,7 +15,7 @@
 import unittest
 
 import mock
-
+from datetime import datetime
 from opencensus.trace import span_context
 from opencensus.trace import span_data as span_data_module
 from opencensus.trace import time_event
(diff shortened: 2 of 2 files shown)


tests/unit/trace/exporters/test_zipkin_exporter.py
```diff
@@ -202,6 +205,178 @@ def test_translate_to_zipkin_span_kind_none(self):
...
+    def test_translate_to_zipkin_with_annotations(self):
+        trace_id = '6e0c63257de34c92bf9efcd03927272e'
+
+        annotation_attributes = {
+            'annotation_bool': True,
+            'annotation_string': 'annotation_test',
+            'key_float': .3
+        }
+
+        import datetime
```
odeke-em: Please put imports at the top and not unexpected in the body of a test.

tests/unit/trace/exporters/test_zipkin_exporter.py
```diff
@@ -202,6 +205,178 @@ def test_translate_to_zipkin_span_kind_none(self):
...
+                stack_trace=None,
+                time_events=time_events,
+                links=None,
+                status=None,
+                same_process_as_parent_span=None,
+                span_kind=None,
+            ),
+        ]
+
+        trace_id = '6e0c63257de34c92bf9efcd03927272e'
```
odeke-em: Why are we redefining this `trace_id`? Is this a copy and paste mistake?

tests/unit/trace/exporters/test_zipkin_exporter.py
mayurkale22: Done

tests/unit/trace/exporters/test_zipkin_exporter.py
mayurkale22: Same problem with other test cases. Corrected.

opencensus/trace/exporters/zipkin_exporter.py
```diff
@@ -211,3 +212,24 @@ def _extract_tags_from_span(attr):
...
+
+    annotations = []
+    for time_event in span.time_events:
+        annotation = time_event.annotation
+        if not annotation:
+            continue
+
+        event_time = datetime.datetime.strptime(time_event.timestamp,
+                                                ISO_DATETIME_REGEX)
+        epoch_time = calendar.timegm(
```
odeke-em: it would be nice in the future for us to unify time conversions for Zipkin to a helper function say
`datetimeStrToMicroseconds`
```python
def datetimeStrToMicroseconds(timestamp):
   """
  Convert a timestamp string into a microsecond value as required by Zipkin.
  It returns an int
   """
   dt = datetime.datetime.strptime(timestamp, ISO_DATETIME_REGEX)
   mUs = (calendar.timegm(dt.timetuple()) * 1e6) + dt.microsecond
   return int(round(mUs))
```
To avoid repeating this pattern. I'll file an issue to follow-up with this.

odeke-em: I've filed followup issue https://github.com/census-instrumentation/opencensus-python/issues/328 PTAL.

opencensus/trace/exporters/zipkin_exporter.py
```diff
@@ -211,3 +212,24 @@ def _extract_tags_from_span(attr):
...
+            event_time.timetuple()) * 1e6 + event_time.microsecond
+        annotations.append({'timestamp': int(round(epoch_time)),
```
odeke-em: Actually hold up, Zipkin's data model says microseconds but these are millisecond conversions unless am confused
datacenterknowledge.com/archives/2017/03/16/google-data-center-faq
screen shot 2018-09-24 at 5 48 34 pm

mayurkale22:
@odeke-em Thanks for approval

PR #332, opened 2018-09-27
Refactor code timestampStr to microseconds

### LINKED ISSUE #328: exporter/zipkin: refactor by extraction code for conversion from timestampStr to microseconds
I noticed while reviewing https://github.com/census-instrumentation/opencensus-python/pull/327#discussion_r220030262 that there is a legacy duplication of code to convert between timestamp strings to microseconds that is
https://github.com/census-instrumentation/opencensus-python/blob/c96083589792a0c278193af1a227c21f9c6c6e97/opencensus/trace/exporters/zipkin_exporter.py#L154-L167

and with that PR there will be three sites with the same code pattern.

I propose that we refactor that code by extracting it to a small helper function
```python
def datetimeStrToMilliseconds(timestamp):
   """
   Convert a timestamp string into a millisecond value as required by Zipkin.
   It returns an int
   """
   dt = datetime.datetime.strptime(timestamp, ISO_DATETIME_REGEX)
   mUs = (calendar.timegm(dt.timetuple()) * 1e6) + dt.microsecond
   return int(round(mUs))
``` 
which will then reduce all those call sites to simply
```python
start_timestamp_ms = datetimeStrToMilliseconds(span.start_time)
end_timestamp_ms = datetimeStrToMilliseconds(span.end_time)
epoch_time_ms = datetimeStrToMilliseconds(time_event.timestamp)

mayurkale22:
Fixes #328


songy23:
Please rebase.

Issue #335, opened 2018-10-04
Add Metrics APIs as per "metrics.proto"


mayurkale22:
Add Metrics package and below classes based on [metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto). Although Metrics proto is still being finalized, We think it is safe to start working on value classes like LabelKey, TimeSeries, Point etc.

- [x] Metrics: Add LabelKey and LabelValue.
- [x] Metrics: Add Value and Point.
- [x] Metrics: Add TimeSeries.
- [x] Metrics: Add MetricDescriptor.
- [x] Metrics: Add Metric.
- [x] Metrics: Add Summary and Distribution.
- [x] Metrics: Produce and store Metrics in Stats impl.

Counterpart classes available in Java for reference : https://github.com/census-instrumentation/opencensus-java/tree/master/api/src/main/java/io/opencensus/metrics

PR #336, opened 2018-10-04
Add Metrics: LabelKey and LabelValue

mayurkale22:
This PR is based on #335


tests/unit/metrics/test_label_key.py
```diff
@@ -35,12 +35,3 @@ def test_constructor_Empty(self):
         self.assertIsNotNone(label_key)
         self.assertEqual(label_key.key, '')
         self.assertEqual(label_key.description, '')
-
-    def test_constructor_WithNonAsciiChars(self):
```
songy23: Why remove these tests? I think we should allow non-ascii characters here.

PR #338, opened 2018-10-07
Propagate tracing headers for requests module

irekatroszko:
This includes the header generated by current trace propagator's method
when wrapping requests.Session. It fixes the cross request propagation
for requests module.

The fix targets wrap on requests.Session object, because it covers all common
requests module use cases:

- Calling requests.{get,post,put,...} instantiates requests.Session,
so the header would end up properly applied from the Session instance.

- Instantiating requests.Session for caching purposes would directly
apply the header.


reyang:
@irekatroszko this branch is out-of-date, would you help to update it?

irekatroszko:
Should be up to date now @reyang

reyang:
@irekatroszko thanks for the contribution!

PR #337, opened 2018-10-05
Metrics: Add Value, Point and Summary

mayurkale22:

PR #342, opened 2018-10-09
Add MetricDescriptor class, tests

c24t:
This is a largely mechanical translation of [the Java client's `MetricDescriptor` class](https://github.com/census-instrumentation/opencensus-java/blob/08d09c48a40b7dc6b5dba8603795084fc6ea9695/api/src/main/java/io/opencensus/metrics/export/MetricDescriptor.java), in the style of the other metrics classes from #337.

Addresses #335.


opencensus/metrics/export/metric_descriptor.py
```diff
@@ -0,0 +1,140 @@
...
+                        MetricDescriptorType.CUMULATIVE_DISTRIBUTION}
+
+
+class MetricDescriptorType(object):
+    """The kind of metric. It describes how the data is reported.
+
+    MetricDescriptorType is an enum of valid MetricDescriptor type values. See
+    opencensus-proto for details:
+
+        https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L73  # noqa
```
c24t: I wasn't sure whether to include this, but since we're copying these wholesale from the spec I thought it'd be helpful to have a link. We might also consider generating code for this kind of thing.

opencensus/metrics/export/metric_descriptor.py
```diff
@@ -0,0 +1,140 @@
...
+
+    A gauge is an instantaneous measurement of a value.
+
+    A cumulative measurement is a value accumulated over a time interval. In a
+    time series, cumulative measurements should have the same start time and
+    increasing end times, until an event resets the cumulative value to zero
+    and sets a new start time for the following points.
+
+    """
+    __metaclass__ = _MetricDescriptorTypeMeta
```
c24t: I'm happy to change this if you've got a different preferred way of handling enums.

c24t: Changed to use `six.add_metaclass` for python 3.

songy23:
@c24t Do you have access to merge commits now?

PR #347, opened 2018-10-12
Add Metric and supporting classes

c24t:
Like #342, this diff adds a few metrics classes following the [proto spec](https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L33) and [java implementation](https://github.com/census-instrumentation/opencensus-java/tree/08d09c48a40b7dc6b5dba8603795084fc6ea9695/api/src/main/java/io/opencensus/metrics/export). The new classes are `Metric`, `TimeSeries`, and the value class `ValueDistribution`.

This diff is written in the style of #337, following the java package structure. I included preconditions to match the comments in the metrics proto file, but this client's behavior isn't guaranteed to match the java client's.

Addresses #335.

### DIFF
--- a/opencensus/metrics/export/metric.py
+++ b/opencensus/metrics/export/metric.py
@@ -22,17 +22,17 @@ class Metric(object):
     This class implements the spec for v1 Metrics as of opencensus-proto
     release v0.0.2. See opencensus-proto for details:
 
-        https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#33  # noqa
+    https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#33
 
     Defines a Metric which has one or more timeseries.
 
-    :type descriptor: class: '~opencensus.metrics.export.metric_descriptor.MetricDescriptor'  # noqa
+    :type descriptor: class: '~opencensus.metrics.export.metric_descriptor.MetricDescriptor'
     :param descriptor: The metric's descriptor.
 
     :type timeseries: list(:class: '~opencensus.metrics.export.time_series.TimeSeries')
     :param timeseries: One or more timeseries for a single metric, where each
     timeseries has one or more points.
-    """
+    """  # noqa
 
     def __init__(self, descriptor, time_series):
         if not time_series:
--- a/opencensus/metrics/export/metric.py
+++ b/opencensus/metrics/export/metric.py
@@ -16,6 +16,24 @@ from opencensus.metrics.export import metric_descriptor
 from opencensus.metrics.export import value
 
 
+DESCRIPTOR_VALUE = {
+    metric_descriptor.MetricDescriptorType.GAUGE_INT64:
+    value.ValueLong,
+    metric_descriptor.MetricDescriptorType.CUMULATIVE_INT64:
+    value.ValueLong,
+    metric_descriptor.MetricDescriptorType.GAUGE_DOUBLE:
+    value.ValueDouble,
+    metric_descriptor.MetricDescriptorType.CUMULATIVE_DOUBLE:
+    value.ValueDouble,
+    metric_descriptor.MetricDescriptorType.GAUGE_DISTRIBUTION:
+    value.ValueDistribution,
+    metric_descriptor.MetricDescriptorType.CUMULATIVE_DISTRIBUTION:
+    value.ValueDistribution,
+    metric_descriptor.MetricDescriptorType.SUMMARY:
+    value.ValueSummary,
+}
+
+
 class Metric(object):
     """A collection of time series data and label metadata.
 
@@ -53,24 +71,8 @@ class Metric(object):
 
     def _check_type(self):
         """Check that point value types match the descriptor type."""
-        check_type = None
-        if self.descriptor.type in (
-                metric_descriptor.MetricDescriptorType.GAUGE_INT64,
-                metric_descriptor.MetricDescriptorType.CUMULATIVE_INT64):
-            check_type = value.ValueLong
-        elif self.descriptor.type in (
-                metric_descriptor.MetricDescriptorType.GAUGE_DOUBLE,
-                metric_descriptor.MetricDescriptorType.CUMULATIVE_DOUBLE):
-            check_type = value.ValueDouble
-        elif self.descriptor.type in (
-                metric_descriptor.MetricDescriptorType.GAUGE_DISTRIBUTION,
-                metric_descriptor.MetricDescriptorType.CUMULATIVE_DISTRIBUTION
-        ):
-            check_type = value.ValueDistribution
-        elif self.descriptor.type == (
-                metric_descriptor.MetricDescriptorType.SUMMARY):
-            check_type = value.ValueSummary
-        else:
+        check_type = DESCRIPTOR_VALUE.get(self.descriptor.type)
+        if check_type is None:
             raise ValueError("Unknown metric descriptor type")
         for ts in self.time_series:
             if not ts.check_points_type(check_type):
--- a/opencensus/metrics/export/metric.py
+++ b/opencensus/metrics/export/metric.py
@@ -12,12 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from opencensus.metrics.export.metric_descriptor import MetricDescriptorType
-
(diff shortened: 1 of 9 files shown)


opencensus/metrics/export/metric.py
```diff
@@ -0,0 +1,76 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from opencensus.metrics.export.metric_descriptor import MetricDescriptorType
+
+from opencensus.metrics.export.value import ValueLong
+from opencensus.metrics.export.value import ValueDouble
```
mayurkale22: Sort the imports, and import module instead of class.

opencensus/metrics/export/metric.py
```diff
@@ -0,0 +1,76 @@
...
+
+    :type descriptor: class: '~opencensus.metrics.export.metric_descriptor.MetricDescriptor'  # noqa
+    :param descriptor: The metric's descriptor.
+
+    :type timeseries: list(:class: '~opencensus.metrics.export.time_series.TimeSeries')
+    :param timeseries: One or more timeseries for a single metric, where each
+    timeseries has one or more points.
+    """
+
+    def __init__(self, time_series, descriptor):
```
mayurkale22: nit: s/descriptor/metric_descriptor, also first param should be ```descriptor```.

opencensus/metrics/export/metric.py
```diff
@@ -0,0 +1,76 @@
...
+        if not time_series:
+            raise ValueError
```
mayurkale22: error message is missing, something like ```raise ValueError("time_series must not be null")```

opencensus/metrics/export/metric.py
```diff
@@ -0,0 +1,76 @@
...
+    :type descriptor: class: '~opencensus.metrics.export.metric_descriptor.MetricDescriptor'  # noqa
+    :param descriptor: The metric's descriptor.
+
+    :type timeseries: list(:class: '~opencensus.metrics.export.time_series.TimeSeries')
+    :param timeseries: One or more timeseries for a single metric, where each
+    timeseries has one or more points.
+    """
+
+    def __init__(self, time_series, descriptor):
+        if not time_series:
```
mayurkale22: Check time_series element not null

opencensus/metrics/export/time_series.py
```diff
@@ -0,0 +1,75 @@
...
+    '~opencensus.metrics.label_value.LabelValue')
+    :param label_values: The set of label values that uniquely identify this
+    timeseries.
+
+    :type points: list(:class: '~opencensus.metrics.export.point.Point')
+    :param points: The data points of this timeseries.
+    """
+
+    def __init__(self, start_timestamp, label_values, points):
+        if start_timestamp is None:
```
mayurkale22: Param should be like label_values, points, start_timestamp

opencensus/metrics/export/time_series.py
```diff
@@ -0,0 +1,75 @@
...
+
+    This class implements the spec for v1 TimeSeries structs as of
+    opencensus-proto release v0.0.2. See opencensus-proto for details:
+
+        https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L112  # noqa
+
+    A TimeSeries is a collection of data points that describes the time-varying
+    values of a metric.
+
+    :type start_timestamp: int
```
mayurkale22: type ```time```?

opencensus/metrics/export/time_series.py
```diff
@@ -0,0 +1,75 @@
...
+    :param start_timestamp: The time when the cumulative value was reset to
+    zero.
```
mayurkale22: Please mention ```Must be non-null for cumulative points```

opencensus/metrics/export/value.py
```diff
@@ -86,5 +89,148 @@ class ValueSummary(Value):
...
+
+    Exemplars are example points that may be used to annotate aggregated
+    Distribution values. They are metadata that gives information about a
+    particular value added to a Distribution bucket.
+
+    :type value: double
+    :param value: Value of the exemplar point, determines which bucket the
+    exemplar belongs to.
+
+    :type timestamp: str
```
mayurkale22: type ```time```

opencensus/metrics/export/value.py
```diff
@@ -86,5 +89,148 @@ class ValueSummary(Value):
...
+    :param count: The number of values in the population.
+
+    :type sum_: float
+    :param sum_: The sum of the values in the population.
+
+    :type sum_of_squared_deviation: float
+    :param sum_of_squared_deviation: The sum of squared deviations from the
+    mean of the values in the population.
+
+    :type bucket_bounds: list(float)
```
mayurkale22: Please add TODO here to change this to ```BucketOptions``` as per new [metric.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto#L204).

opencensus/metrics/export/metric.py
c24t: Ah I see the note in the style guide about class imports now, done.

opencensus/metrics/export/metric.py
c24t: It's called `descriptor` in the proto def: https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L37.

opencensus/metrics/export/metric.py
c24t: This is checking for null and empty series, I'm assuming that's the behavior we want based on ["one or more timeseries for a single metric"](https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L47), but there's no such check in the java client.

opencensus/metrics/export/time_series.py
c24t: What's a `time`? Changed to `str` to match the existing tests for other ts fields, but we should consider using datetimes (or epoch ms/ns) everywhere instead.

opencensus/metrics/export/time_series.py
c24t: How come? I'm following the order in the proto def: https://github.com/census-instrumentation/opencensus-proto/blob/24333298e36590ea0716598caacc8959fc393c48/src/opencensus/proto/metrics/v1/metrics.proto#L112.

opencensus/metrics/export/value.py
c24t: See above, unless time has some special meaning I'm missing here.

opencensus/metrics/export/time_series.py
mayurkale22: This is to keep things consistent across libraries. Couterapart in Java : https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/metrics/export/TimeSeries.java#L53. Also start_timestamp is optional field, but must be non-null for cumulative Points.

c24t: OK, I changed the field order and added a check in f17f1f1.

opencensus/metrics/export/metric.py
```diff
@@ -0,0 +1,89 @@
...
+        elif self.descriptor.type in (
+                metric_descriptor.MetricDescriptorType.GAUGE_DISTRIBUTION,
+                metric_descriptor.MetricDescriptorType.CUMULATIVE_DISTRIBUTION
+        ):
+            check_type = value.ValueDistribution
+        elif self.descriptor.type == (
+                metric_descriptor.MetricDescriptorType.SUMMARY):
+            check_type = value.ValueSummary
+        else:
+            raise ValueError("Unknown metric descriptor type")
```
liyanhui1228:
```
# This relation mapping could be put at the top of this file
DESCRIPTOR_VALUE = {
    metric_descriptor.MetricDescriptorType.GAUGE_INT64: value.ValueLong,
    metric_descriptor.MetricDescriptorType.CUMULATIVE_INT64: value.ValueLong,
    ...
}

check_type = DESCRIPTOR_VALUE.get(self.descriptor.type)
if check_type is None:
    raise ValueError("Unknown metric descriptor type")
```

PR #361, opened 2018-10-25
Fix unit test failures on Windows

reyang:
The current implementation fails the unit test on Windows, due to the backslashes introduced by os.path.join.
The fix is to replace backslash with forward slash before return.

### DIFF
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -427,9 +427,8 @@ def get_task_value():
 def namespaced_view_name(view_name, metric_prefix):
     """ create string to be used as metric type
     """
-    if metric_prefix == "":
-        return os.path.join("custom.googleapis.com", "opencensus", view_name)
-    return os.path.join(metric_prefix, view_name)
+    metric_prefix = metric_prefix or "custom.googleapis.com/opencensus"
+    return os.path.join(metric_prefix, view_name).replace('\\', '/')
 
 
 def new_label_descriptors(defaults, keys):


reyang:
Looks like the CI environment has been updated which caused the following error:

opencensus/stats/exporters/stackdriver_exporter.py:23:42: W606 'async' and 'await' are reserved keywords starting with Python 3.7
opencensus/stats/exporters/stackdriver_exporter.py:112:28: W606 'async' and 'await' are reserved keywords starting with Python 3.7
opencensus/stats/exporters/stackdriver_exporter.py:450:5: F841 local variable 'exception' is assigned to but never used
opencensus/trace/propagation/google_cloud_format.py:22:17: W605 invalid escape sequence '\/'
opencensus/trace/propagation/google_cloud_format.py:22:40: W605 invalid escape sequence '\d'

I'll try to fix them.

reyang:
I've submitted a separate PR #362 to unblock the build.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -412,9 +412,8 @@ def get_task_value():
 def namespaced_view_name(view_name, metric_prefix):
     """ create string to be used as metric type
     """
-    if metric_prefix == "":
-        return os.path.join("custom.googleapis.com", "opencensus", view_name)
-    return os.path.join(metric_prefix, view_name)
+    metric_prefix = metric_prefix or "custom.googleapis.com/opencensus"
+    return os.path.join(metric_prefix, view_name).replace('\\', '/')
```
SergeyKanzhelev: using os.path.join here may not be ideal. Perhaps use `join` with the specified separator instead?

reyang: os.path.join will take care of duplicated/trailing separators, for example, join("a/b/", "c") will yield the same result as join("a/b", "c"). It'll be better to reuse such logic from os.path.join than manually doing path normalization and joining.

reyang:
Thanks @bogdandrutu and @SergeyKanzhelev.

PR #362, opened 2018-10-25
Fix build errors

reyang:
I'm getting build errors which seems to be related to the recent py3.6 upgrade in CI:
opencensus/stats/exporters/stackdriver_exporter.py:23:42: W606 'async' and 'await' are reserved keywords starting with Python 3.7
opencensus/stats/exporters/stackdriver_exporter.py:112:28: W606 'async' and 'await' are reserved keywords starting with Python 3.7
opencensus/stats/exporters/stackdriver_exporter.py:450:5: F841 local variable 'exception' is assigned to but never used
opencensus/trace/propagation/google_cloud_format.py:22:17: W605 invalid escape sequence '/'
opencensus/trace/propagation/google_cloud_format.py:22:40: W605 invalid escape sequence '\d'

async is becoming a keyword in python3.7, this blocks the current CI (which is using the latest py3.6, and failed due to warnings).


reyang:
@liyanhui1228 do we want to combine

opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -19,7 +19,7 @@
 from opencensus.trace.trace_options import TraceOptions
 
 _TRACE_CONTEXT_HEADER_NAME = 'X-Cloud-Trace-Context'
-_TRACE_CONTEXT_HEADER_FORMAT = '([0-9a-f]{32})(\/([0-9a-f]{16}))?(;o=(\d+))?'
+_TRACE_CONTEXT_HEADER_FORMAT = r'([0-9a-f]{32})(\/([0-9a-f]{16}))?(;o=(\d+))?'
```
reyang: This is a separate issue which was caught by the latest pylint.
The string literal used for regular expression was not properly escaped.

zoidbergwill:
I'm not a fan of `async_.py`, but we can fix it in another PR.

reyang:
> I'm not a fan of async_.py, but we can fix it in another PR.

Me too! Let's see if we can get rid of it in #642.

PR #363, opened 2018-10-26
Thread pools integration

geobeau:
Today, Opencensus wraps Threading, we propagate the current tracer to the child thread by wrapping Thread.run() and Thread.start() (https://github.com/python/cpython/blob/3.6/Lib/threading.py). 

Unfortunately it doesn't work well when we use a pool of reusable worker thread. Multiprocessing.Pool (https://github.com/python/cpython/blob/3.6/Lib/multiprocessing/pool.py) lazy create a pool of thread the first time we _init_ the Pool object, propagating its tracer to them. When it gets reused later by another request, the task sent to the workers is traced using the tracer from the original request breaking both of the trace.

To trace tasks sent to thread pool, we wrap apply_async() to wrap func and add the tracer in the args. futures.ThreadPoolExecutor is also tracer by wrapping submit()


reyang:
@geobeau build failed due to a recent upgrade in circleci, which is unrelated to your change. I have it fixed in #362, would you update the branch again? Thank you!

geobeau:
Done @reyang :)

tests/unit/trace/test_execution_context.py
```diff
@@ -38,3 +40,60 @@ def test_has_attrs(self):
...
+
+        thread_local = threading.local()
+        setattr(thread_local, 'random_non_oc_attr', some_value)
+
+        execution_context.clean()
+
+        self.assertNotEqual(mock_span, execution_context.get_current_span())
+        self.assertEqual(some_value, getattr(thread_local, 'random_non_oc_attr'))
+
+
```
reyang: remove the extra empty lines.

reyang:
@geobeau great! LGTM besides the extra empty lines.

ocervell:
Does this mean we now support passing `trace_id` from the main thread to children threads ?

geobeau:
Yes. The behavior for thread pools is a bit more complicated but the normal thread creation and passing should work well.

PR #369, opened 2018-10-26
Replace exporter version constants with package version

### LINKED ISSUE #296: Single definition for OC version across all components
The oc-python version is defined separately in setup.py, stackdriver exporters, ocagent exporter, so version update is error prone.

c24t:
Fixes #296 by adding a top-level version file and removing hardcoded version numbers in the stackdriver and ocagent exporters.

I'm `exec`ing the version file in `setup.py` to avoid importing the opencensus package during setup. This isn't strictly necessary now, but I think this solution is simpler and likely to save us some headaches in the future.

cc @lmolkova

### DIFF
--- a/opencensus/__version__.py
+++ b/opencensus/__version__.py
@@ -1,0 +1,15 @@
+# Copyright 2018, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


c24t:
Thanks @lmolkova, I updated the test.

PR #370, opened 2018-10-26
Set stackdriver stats exporter user agent

c24t:
This diff sets the stackdriver exporter's GC client's `client_library_version` to this library's version, to be included in the user agent that the GC client [attaches to API calls](https://github.com/googleapis/google-cloud-python/blob/0c13d62e807502b6392511108d46bed3e5634ffa/api_core/google/api_core/gapic_v1/method.py#L226-L229).

The form of the user agent before this change:

    gl-python/2.7.10 grpc/1.8.3 gax/1.4.1

And after:

     gl-python/2.7.10 grpc/1.8.3 gax/1.4.1 gccl/opencensus-1.0.8

### DIFF
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -12,17 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 import os
 import platform
-from . import base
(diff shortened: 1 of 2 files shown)


c24t:
I changed the UA in 5f3bb69 to include "opencensus" since the version number alone doesn't tell us which client library the caller is using.

PR #375, opened 2018-10-31
DistributionAggregation fixes

### LINKED ISSUE #373: Multiple apparent bugs in DistributionAggregation implementation
It looks like there were a few bugs added in #149 and #287:

Exemplars are always attached to the last bucket:

```python
from opencensus.stats.aggregation import DistributionAggregation

da = DistributionAggregation(list(map(float, range(0, 11))))
ad = da.aggregation_data

for dp in range(10):
    ad.add_sample(dp, dt.now(), {"trace_id": "dead", "span_id": "beef"})
    
ad.exemplars
```
```python
{11: <opencensus.stats.aggregation_data.Exemplar at 0x10ccf6390>}
```

Off-by-one error adding a value greater than the last bucket boundary:

```python
from datetime import datetime as dt

da = DistributionAggregation([1, 3, 5])
ad = da.aggregation_data

ad.add_sample(float('-inf'), dt.now(), {"trace_id": "dead", "span_id": "beef"})
ad.add_sample(2, dt.now(), {"trace_id": "dead", "span_id": "beef"})
ad.add_sample(4, dt.now(), {"trace_id": "dead", "span_id": "beef"})
ad.add_sample(float('inf'), dt.now(), {"trace_id": "dead", "span_id": "beef"})
  
ad.counts_per_bucket
```
```python
[1, 1, 2, 0]
```

Min/max are 0 by default, which doesn't work for e.g. histograms without 0-valued points:

```python
da = DistributionAggregation(list(map(float, range(0, 11))))
ad = da.aggregation_data

for dp in range(10, 20):
    ad.add_sample(dp, dt.now(), {"trace_id": "dead", "span_id": "beef"})
    
ad.min  # should be 10
```
```python
0
```
...or histograms with all negative points:

```python
da = DistributionAggregation(list(map(float, range(0, 11))))
ad = da.aggregation_data

for dp in range(-20, -10):
    ad.add_sample(dp, dt.now(), {"trace_id": "dead", "span_id": "beef"})
    
ad.max  # should be -10
```
```python
0
```

c24t:
This diff addresses a few bugs in `DistributionAggregation` and `DistributionAggregationData` from #373 that needed to be fixed before converting these aggregations to metrics for #335.

This is a noisy diff since I formatted and fixed old lint errors in the files I touched. I'll comment on the important changes.

Fixes #373.

### DIFF
--- a/opencensus/stats/aggregation_data.py
+++ b/opencensus/stats/aggregation_data.py
@@ -217,7 +217,6 @@ class DistributionAggregationData(BaseAggregationData):
         self._sum_of_sqd_deviations = self._sum_of_sqd_deviations + (
             (value - old_mean) * (value - self._mean_data))
 
-
     def increment_bucket_count(self, value):
         """Increment the bucket count based on a given value from the user"""
         if len(self._bounds) == 0:
--- a/opencensus/stats/aggregation_data.py
+++ b/opencensus/stats/aggregation_data.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 from opencensus.stats import bucket_boundaries
 
 
@@ -21,6 +22,7 @@ class BaseAggregationData(object):
     :param aggregation_data: represents the aggregated value from a collection
 
     """
+
     def __init__(self, aggregation_data):
         self._aggregation_data = aggregation_data
 
@@ -37,6 +39,7 @@ class SumAggregationDataFloat(BaseAggregationData):
     :param sum_data: represents the aggregated sum
 
     """
+
     def __init__(self, sum_data):
         super(SumAggregationDataFloat, self).__init__(sum_data)
         self._sum_data = sum_data
@@ -60,6 +63,7 @@ class CountAggregationData(BaseAggregationData):
     :param count_data: represents the aggregated count
 
     """
+
     def __init__(self, count_data):
         super(CountAggregationData, self).__init__(count_data)
         self._count_data = count_data
@@ -104,6 +108,7 @@ class DistributionAggregationData(BaseAggregationData):
     :param bounds: the histogram distribution of the values
 
     """
+
     def __init__(self,
                  mean_data,
                  count_data,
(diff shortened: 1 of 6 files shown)


songy23:
Please fix the lint.
```bash
opencensus/stats/aggregation_data.py:221:5: E303 too many blank lines (2)
nox > Command flake8 --exclude=opencensus/trace/exporters/gen/opencensus/ opencensus/ failed with exit code 1
nox > Session lint failed. :(
```

tests/unit/stats/test_measure_to_view_map.py
```diff
@@ -229,12 +221,11 @@ def test_record(self):
         view_columns = ["testTag1", "testColumn2"]
         view_measure = measure
         view_aggregation = mock.Mock()
-        view = View(
-            name=view_name,
-            description=view_description,
-            columns=view_columns,
-            measure=view_measure,
-            aggregation=view_aggregation)
+        View(name=view_name,
```
songy23: Why is this created but never used?

tests/unit/stats/test_measure_to_view_map.py
```diff
@@ -343,7 +315,7 @@ def test_record_with_exporter(self):
         view_columns = ["testTag1", "testColumn2"]
         view_measure = measure
         view_aggregation = mock.Mock()
-        view = View(
+        View(
```
songy23: ditto

tests/unit/stats/test_measure_to_view_map.py
```diff
@@ -378,18 +350,13 @@ def test_export(self):
...
```
songy23: ditto

opencensus/stats/aggregation_data.py
```diff
@@ -123,13 +128,14 @@ def __init__(self,
             bounds = []
 
         if counts_per_bucket is None:
-            counts_per_bucket = []
-            bucket_size = len(bounds) + 1
-            for i in range(bucket_size):
-                counts_per_bucket.append(0)
+            counts_per_bucket = [0 for ii in range(len(bounds) + 1)]
+        elif len(counts_per_bucket) != len(bounds) + 1:
+            raise ValueError("counts_per_bucket length does not match bounds "
```
mayurkale22: I think we can change to ```"counts_per_bucket has X fields, which is more than the Y buckets allowed."``` where, X:len(counts_per_bucket) and Y: len(bounds) + 1. What do you think?

tests/unit/stats/test_measure_to_view_map.py
c24t: I didn't start the fire here, I just removed the unused `view` var to fix the [F841](https://lintlyci.github.io/Flake8Rules/rules/F841.html) lint error, and same for the changes below.

Many of the tests in this module seem half-baked, but I think this is a problem for another PR.

opencensus/stats/aggregation_data.py
c24t: One benefit of fixed error messages is that you can grep for the exact text, but I'm happy with either here.

tests/unit/stats/test_measure_to_view_map.py
songy23: SGTM

PR #379, opened 2018-11-01
Turn prints into log statements

ocervell:
We had some hanging 'print' statements in the BackgroundThread: let's turn them into `logging` statements so that an end-user can control what logs they want to see.


c24t:
Thanks @ocervell! If you're ok with my change I'll merge it.

Issue #249, opened 2018-08-03
custom parameters to BackgroundThreadTransport


felippe-mendonca:
When I started to use ZipkingExporter with the BackgroundThreadTransport, I'd noticed some delay to export *(as well as the increase of queue of spans, driving an increase of used memory from service)* in services that generate spans with a high rate. Then I realised that the [BackgroundThreadTransport](https://github.com/census-instrumentation/opencensus-python/blob/2f501202965c6bb704d61fcc7a055c82b0645aa4/opencensus/trace/exporters/transports/background_thread.py#L204) constructor has two parameters, but it's not possible to set these parameters since the ZipkinExporter constructor receives a type of transport instead of an instance. For me, it seems that isn't possible to change the constructor parameters of this transport type, but in some cases it could be interesting.

For now, to cover my issue I change the constructor of ZipkingExporter (see below) as well as implementation of BackgroundThreadExporter (now it has a method to set the exporter instead of receive this on constructor).

```python
exporter = ZipkinExporter(
  service_name='MyService',
  host_name='localhost',
  port=9411,
  transport=BackgroundThreadTransport(max_batch_size=100)
)
```

I didn't open a pull request with this patch because I was in doubt if it's possible to change the transport parameters without the patch. Moreover, I just made the changes for ZipkinExporter. If this change makes sense, I can make the necessary modifications to cover all the exporter and then open a pull request.

Another thing is: for the BackgroundThreadTransport, beyond the `max_batch_size`, the `_WAIT_PERIOD` parameter may be interesting the possibility to change it.

ocervell:
+1 for this, biting me as well. I think the `grace_period` parameter and `max_batch_size` really depend on the use case, and we should be able to specify those.

reyang:
#642 should address this.

PR #380, opened 2018-11-01
Support transport options in exporter

ocervell:
This PR adds support to pass options to the transport initialized in our exporters.

This will allow, for instance, to set the grace_period and max_batch_size for the BackgroundThread transport when initializing the exporter:
```
from opencensus.trace.exporters.stackdriver_exporter import StackdriverExporter

exporter = StackdriverExporter(
    transport=BackgroundThreadTransport, 
    transport_config={'grace_period': 1, 'max_batch_size': 30})
```

### DIFF
--- a/opencensus/trace/exporters/file_exporter.py
+++ b/opencensus/trace/exporters/file_exporter.py
@@ -31,14 +31,14 @@ class FileExporter(base.Exporter):
     :type file_mode: str
     :param file_mode: The file mode to open the output file with.
                       Defaults to w+
-                      
+
     :type transport: :class:`type`
     :param transport: Class for creating new transport objects. It should
                       extend from the base :class:`.Transport` type and
                       implement :meth:`.Transport.export`. Defaults to
                       :class:`.SyncTransport`. The other option is
                       :class:`.BackgroundThreadTransport`.
-                      
+
     :type transport_config: :class:`dict`
     :param transport_config: Transport configuration dictionary.
     """
--- a/opencensus/trace/exporters/file_exporter.py
+++ b/opencensus/trace/exporters/file_exporter.py
@@ -41,12 +41,11 @@ class FileExporter(base.Exporter):
 
     """
 
-    def __init__(self, file_name=DEFAULT_FILENAME,
-                 transport=sync.SyncTransport,
-                 file_mode='w+'):
+    def __init__(self, file_name=DEFAULT_FILENAME, file_mode='w+',
+                 transport=sync.SyncTransport, transport_config=None):
         self.file_name = file_name
-        self.transport = transport(self)
         self.file_mode = file_mode
+        self.transport = self.__init_transport(transport, transport_config)
 
     def emit(self, span_datas):
         """
--- a/opencensus/trace/exporters/file_exporter.py
+++ b/opencensus/trace/exporters/file_exporter.py
@@ -45,7 +45,7 @@ class FileExporter(base.Exporter):
                  transport=sync.SyncTransport, transport_config=None):
         self.file_name = file_name
         self.file_mode = file_mode
-        self.transport = self.__init_transport(transport, transport_config)
+        self.transport = base.init_transport(self, transport, transport_config)
 
     def emit(self, span_datas):
         """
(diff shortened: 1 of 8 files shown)


opencensus/trace/exporters/file_exporter.py
```diff
@@ -41,12 +41,11 @@ class FileExporter(base.Exporter):
 
     """
 
-    def __init__(self, file_name=DEFAULT_FILENAME,
-                 transport=sync.SyncTransport,
-                 file_mode='w+'):
+    def __init__(self, file_name=DEFAULT_FILENAME, file_mode='w+',
+                 transport=sync.SyncTransport, transport_config=None):
         self.file_name = file_name
-        self.transport = transport(self)
```
c24t: I wonder why we're initializing the transport here instead of passing the exporter constructor a preconfigured transport. This may be a simpler solution to your problem.

felippe-mendonca: @c24t, I suggested this in #249 (pass an object instead), but add a configuration parameter could be interesting for backwards compatibility.

ocervell: I agree with @felippe-mendonca, this is mostly to keep backwards compatibility while giving more options. Maybe we can think of a more global reformat later on.

c24t:
@ocervell it looks like this branch still has some lint and coverage failures.

To check for new lint errors quickly you can do:
```sh
flake8 $(git diff --name-only upstream/master)
```
And to see the detailed coverage report:
```sh
% python -m pytest --cov=opencensus --cov-config=.coveragerc
% python -m coverage html
% open htmlcov/index.html
```

ocervell:
Need to add a test for the `init_transport` function.

felippe-mendonca:
> Need to add a test for the `init_transport` function to get full coverage.

@ocervell , I'd suggest only the modification below on `test_base_exporter.py` file to test the `init_transport` function, since it's already tested indirectly on all exporters' tests.

```python
class TestBaseExporter(unittest.TestCase):
    def test_export_abstract(self):
        from opencensus.trace.exporters import base

        exporter = base.Exporter()
        trace = {}

        with self.assertRaises(NotImplementedError):
            exporter.export(trace)

    def test_emit_abstract(self):
        from opencensus.trace.exporters import base

        exporter = base.Exporter()
        trace = {}

        with self.assertRaises(NotImplementedError):
            exporter.emit(trace)

    def test_init_transport(self):
        from opencensus.trace.exporters import base

        with self.assertRaises(TypeError):
            transport = base.init_transport(
                exporter=base.Exporter(),
                transport=MockTransport,
                config={
                    "par2": 0,
                })


class MockTransport(object):
    def __init__(self, exporter=None, par1=0):
        self.export_called = False
        self.exporter = exporter
        self.par1 = par1

    def export(self, trace):
        self.export_called = True
```

davewalkexpel:
Hi, this is cool but would it be possible to make `_WAIT_PERIOD` configurable via the new `transport_config` dict? Or would that be out of scope and should I start a new PR?

My use case is I want to send less write requests to Stackdriver so that we don't exceed our write quota. I can control the max batch size of the requests now, but the app is still going to send a write request every second even if there's only on span to emit (traces are getting created continuously). If could I up the wait period to something like five seconds I would really be in business. Thanks!

c24t:
> would it be possible to make `_WAIT_PERIOD` configurable via the new `transport_config` dict?

It'd mean adding an arg to the `AsyncTransport` constructor, but there's at least as good an argument for making this configurable as the grace period (which already has an arg).

felippe-mendonca:
@davewalkexpel, Seems good to be possible to also configure _WAIT_TIME in your use case. I had a similar problem but I'd solved only changing the batch size. Besides, adjust _WAIT_TIME could also be interesting in cases where you have multiples components sending traces, which mean multiple requests to the collector. In that case, increase interval and batch size might reduce collector's pressure. However, I think that this feature fits to another PR, and maybe change the name from 'wait_time' to 'interval'. Or, even more, open an discussion about the AsyncTransport export strategy.

reyang:
This has been addressed by the configuration refactor work in #394.
The following mechanism has been added in #621:
https://github.com/census-instrumentation/opencensus-python/tree/master/opencensus/common/configuration

reyang:
There is plan to refactor exporter, worker, queue and transport, refer to the conversations in #642.

PR #354, opened 2018-10-17
Bump async transport _WAIT_PERIOD to 60s

c24t:
Fixes #353.

PR #382, opened 2018-11-02
Fix hanging async tests

c24t:
Since merging #354 a few tests including [`Test_Worker.test__thread_main`](https://github.com/census-instrumentation/opencensus-python/blob/d3b289748c9b10798c27dc6be59053cc09567f3f/tests/unit/common/transports/test_async.py#L115-L169) have hung while waiting out the 60s exporter delay. This is causing CI tests to time out, as in https://circleci.com/gh/geobeau/opencensus-python/46.

This diff removes the delay for async transport workers during testing and fixes a few existing lint errors.

### DIFF
--- a/tests/unit/common/transports/test_async.py
+++ b/tests/unit/common/transports/test_async.py
@@ -18,6 +18,12 @@ import mock
 
 from opencensus.common.transports import async_
 
+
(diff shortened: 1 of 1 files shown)


c24t:
Thanks @reyang!

PR #389, opened 2018-11-08
Drop negative measurements on record

### LINKED ISSUE #376: Drop negative bucket bounds and negative values in record API
Negative values are currently not supported by any of the backends that we support and have histograms (Stackdriver and Prometheus). We have decided to implement a backwards incompatible change and drop negative values in the Record API (We will start with logging an error for few versions then throw an exception).

This could be a breaking change if users are recording negative value for any `measure`. We planned to implement this change asap in all the libraries and explain this behaviour to the current users very clearly.

Counterpart in Java: https://github.com/census-instrumentation/opencensus-java/pull/1499 and https://github.com/census-instrumentation/opencensus-java/pull/1525

c24t:
This diff prevents sets of measurements from being recorded to view data objects if any measurement in the set is negative, prevents `MeasurementMaps` from being recorded if they've ever contained negative value measurements, and removes 0- and negative-valued bucket boundaries from `DistributionAggregation`s.

See the [java implementation](https://github.com/census-instrumentation/opencensus-java/pull/1499/files) for the justification for these changes.

Fixes #376

### DIFF
--- a/opencensus/stats/aggregation.py
+++ b/opencensus/stats/aggregation.py
@@ -123,11 +123,17 @@ class DistributionAggregation(BaseAggregation):
     :param aggregation_type: represents the type of this aggregation
 
     """
-    def __init__(
-            self,
-            boundaries=None,
-            distribution=None,
-            aggregation_type=Type.DISTRIBUTION):
+
+    def __init__(self,
+                 boundaries=None,
+                 distribution=None,
+                 aggregation_type=Type.DISTRIBUTION):
+        if boundaries:
+            if not all(boundaries[ii] < boundaries[ii + 1]
+                       for ii in range(len(boundaries) - 1)):
+                raise ValueError("bounds must be sorted in increasing order")
+            boundaries = [bb for bb in boundaries if bb > 0]
+
         super(DistributionAggregation, self).__init__(
             buckets=boundaries, aggregation_type=aggregation_type)
         self._boundaries = bucket_boundaries.BucketBoundaries(boundaries)
--- a/opencensus/stats/aggregation.py
+++ b/opencensus/stats/aggregation.py
@@ -12,8 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
 from opencensus.stats import bucket_boundaries
 from opencensus.stats import aggregation_data
+
+
+logger = logging.getLogger(__name__)
 
 
 class Type(object):
@@ -132,7 +137,16 @@ class DistributionAggregation(BaseAggregation):
             if not all(boundaries[ii] < boundaries[ii + 1]
                        for ii in range(len(boundaries) - 1)):
                 raise ValueError("bounds must be sorted in increasing order")
-            boundaries = [bb for bb in boundaries if bb > 0]
+            for ii, bb in enumerate(boundaries):
+                if bb > 0:
+                    break
+            else:
+                ii += 1
+            if ii:
+                logger.warning("Dropping {} negative bucket boundaries, the "
+                               "values must be strictly > 0"
+                               .format(ii))
+            boundaries = boundaries[ii:]
 
         super(DistributionAggregation, self).__init__(
             buckets=boundaries, aggregation_type=aggregation_type)
--- a/opencensus/stats/aggregation_data.py
+++ b/opencensus/stats/aggregation_data.py
@@ -131,29 +131,17 @@ class DistributionAggregationData(BaseAggregationData):
         self._sum_of_sqd_deviations = sum_of_sqd_deviations
         if bounds is None:
             bounds = []
+        else:
+            assert bounds == list(sorted(set(bounds)))
 
         if counts_per_bucket is None:
             counts_per_bucket = [0 for ii in range(len(bounds) + 1)]
-        elif any(cc < 0 for cc in counts_per_bucket):
-            raise ValueError("Bucket counts may not be negative")
-        elif len(counts_per_bucket) != len(bounds) + 1:
-            raise ValueError("counts_per_bucket length does not match bounds "
-                             "length")
+        else:
+            assert all(cc >= 0 for cc in counts_per_bucket)
+            assert len(counts_per_bucket) == len(bounds) + 1
 
-        if bounds:
-            if not all(bounds[ii] < bounds[ii + 1]
-                       for ii in range(len(bounds) - 1)):
-                raise ValueError("bounds must be sorted in increasing order")
-
-            dropped = [bb for bb in bounds if bb <= 0]
-            if dropped:
-                logger.warn("Bucket boundaries must be positive. Dropping "
-                            "boundaries {}"
-                            .format(dropped))
-                bounds = bounds[len(dropped):]
-                dropped_counts = counts_per_bucket[:len(dropped):]
-                counts_per_bucket = counts_per_bucket[len(dropped):]
-                counts_per_bucket[0] += sum(dropped_counts)
+        assert bounds == sorted(bounds)
+        assert all(bb > 0 for bb in bounds)
 
         self._counts_per_bucket = counts_per_bucket
         self._bounds = bucket_boundaries.BucketBoundaries(
--- a/opencensus/stats/aggregation_data.py
+++ b/opencensus/stats/aggregation_data.py
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
 from opencensus.stats import bucket_boundaries
+
+
+logger = logging.getLogger(__name__)
 
 
 class BaseAggregationData(object):
@@ -129,9 +134,26 @@ class DistributionAggregationData(BaseAggregationData):
 
         if counts_per_bucket is None:
             counts_per_bucket = [0 for ii in range(len(bounds) + 1)]
+        elif any(cc < 0 for cc in counts_per_bucket):
+            raise ValueError("Bucket counts may not be negative")
         elif len(counts_per_bucket) != len(bounds) + 1:
             raise ValueError("counts_per_bucket length does not match bounds "
                              "length")
+
+        if bounds:
+            if not all(bounds[ii] < bounds[ii + 1]
+                       for ii in range(len(bounds) - 1)):
+                raise ValueError("bounds must be sorted in increasing order")
+
+            dropped = [bb for bb in bounds if bb <= 0]
+            if dropped:
+                logger.warn("Bucket boundaries must be positive. Dropping "
+                            "boundaries {}"
+                            .format(dropped))
+                bounds = bounds[len(dropped):]
+                dropped_counts = counts_per_bucket[:len(dropped):]
+                counts_per_bucket = counts_per_bucket[len(dropped):]
+                counts_per_bucket[0] += sum(dropped_counts)
 
         self._counts_per_bucket = counts_per_bucket
         self._bounds = bucket_boundaries.BucketBoundaries(
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -164,31 +164,13 @@ class Collector(object):
                                        labels=labels)
         elif isinstance(agg_data,
                         aggregation_data_module.DistributionAggregationData):
+
+            assert(agg_data.bounds == sorted(agg_data.bounds))
             points = {}
-            # Histograms are cumulative in Prometheus.
-            # 1. Sort buckets in ascending order but, retain
-            # their indices for reverse lookup later on.
-            # TODO: If there is a guarantee that distribution elements
-            # are always sorted, then skip the sorting.
-            indices_map = {}
-            buckets = []
-            i = 0
-            for boundarie in view.aggregation.boundaries.boundaries:
-                if boundarie not in indices_map \
-                        or indices_map == {}:  # pragma: NO COVER
-                    indices_map[str(boundarie)] = i
-                    buckets.append(str(boundarie))
-                i += 1
-
-            buckets.sort()
-
-            # 2. Now that the buckets are sorted by magnitude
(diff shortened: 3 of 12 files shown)


opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -164,31 +164,13 @@ def to_metric(self, desc, view):
...
-            # we can create cumulative indicesmap them back by reverse index
             cum_count = 0
-            for bucket in buckets:
-                i = indices_map[bucket]
-                cum_count += int(agg_data.counts_per_bucket[i])
-                points[bucket] = cum_count
+            for ii, bound in enumerate(agg_data.bounds):
+                cum_count += agg_data.counts_per_bucket[ii]
+                points[str(bound)] = cum_count
             labels = desc['labels'] if points is None else None
```
c24t: `points` is never null here, but I didn't touch it because I don't know what prometheus expects. @mayurkale22 let me know if `labels` should be null when there are no points in the distribution and I'll change it.

tests/unit/stats/exporter/test_prometheus_stats.py
```diff
@@ -37,44 +37,47 @@
...
-REGISTERED_VIEW = {'test1_my.org/views/video_size_test2-my.org/keys/frontend':
-                       {'documentation': 'processed video size over time',
-                        'labels': ['my.org/keys/frontend'],
-                        'name': 'test1_my.org/views/video_size_test2'}}
-
-REGISTERED_VIEW2 = {'opencensus_my.org/views/video_size_test2-my.org/keys/frontend':
-                        {'documentation': 'processed video size over time',
-                         'labels': ['my.org/keys/frontend'],
-                         'name': 'opencensus_my.org/views/video_size_test2'}}
+    [16.0 * MiB, 256.0 * MiB])
```
c24t: This is the important change, most of the rest below is lint.

mayurkale22: I would do this (lint) changes in separate PR. If you dont mind can you send another PR for lint styling?

mayurkale22:
On a side note, I think we should add **CHANGELOG.md** now, to document all the notable changes about the versions. Same as: https://github.com/census-instrumentation/opencensus-java/blob/master/CHANGELOG.md. This PR could be a breaking change if users are recording negative value for any measure. What do you think?

tests/unit/stats/exporter/test_prometheus_stats.py
c24t: I moved the lint changes to https://github.com/census-instrumentation/opencensus-python/pull/390, should clean this diff up once it's merged.

opencensus/stats/measure_to_view_map.py
```diff
@@ -91,10 +95,17 @@ def register_view(self, view, timestamp):
         if registered_measure is None:
             self._registered_measures[measure.name] = measure
         self._measure_to_view_data_list_map[view.measure.name].append(
-            ViewData(view=view, start_time=timestamp, end_time=timestamp))
+            view_data_module.ViewData(view=view, start_time=timestamp,
+                                      end_time=timestamp))
 
     def record(self, tags, measurement_map, timestamp, attachments=None):
         """records stats with a set of tags"""
+        for measure, value in measurement_map.items():
```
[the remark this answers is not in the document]
c24t: `MeasureToViewMap.record` is downstream of `MeasurementMap.record`; if you make the change in `MeasurementMap` instead calls to `MeasureToViewMap.record` won't drop negative values, as it is it should work for both classes.

tests/unit/stats/test_aggregation_data.py
```diff
@@ -88,8 +88,8 @@ def test_constructor(self):
         _min = 0
         _max = 1
         sum_of_sqd_deviations = mock.Mock()
-        counts_per_bucket = [1, 1, 1, 1]
-        bounds = [0, 1.0 / 2.0, 1]
```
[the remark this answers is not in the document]
c24t: Added another check in 1a4756e.

c24t:
> On a side note, I think we should add **CHANGELOG.md** now...

I think adding a changelog is a great idea, and we should push a version update once these changes are merged. We have some work to do to check that the code added since the last release is ready to ship though.

opencensus/stats/aggregation_data.py
```diff
@@ -129,10 +134,27 @@ def __init__(self,
...
                              "length")
 
+        if bounds:
+            if not all(bounds[ii] < bounds[ii + 1]
+                       for ii in range(len(bounds) - 1)):
+                raise ValueError("bounds must be sorted in increasing order")
+
+            dropped = [bb for bb in bounds if bb <= 0]
+            if dropped:
+                logger.warn("Bucket boundaries must be positive. Dropping "
```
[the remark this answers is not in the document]
c24t: I don't know that it's worth trying to keep error messages in sync across clients unless they're e.g. in the spec, but updated in bce908c.

opencensus/stats/measure_to_view_map.py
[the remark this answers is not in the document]
c24t: Here's a quick example, it works as is but should fail if the check is moved to `MeasurementMap.record`.

```python
from datetime import datetime as dt
from collections import OrderedDict

from opencensus.stats.measure import MeasureFloat
from opencensus.stats.aggregation import DistributionAggregation
from opencensus.tags.tag_map import TagMap
from opencensus.tags.tag_map import TagValue
from opencensus.tags.tag_map import TagKey
from opencensus.stats.view import View
from opencensus.stats.measurement_map import MeasurementMap

def get_tagmap(dd):
    tm = TagMap()
    for kk, vv in dd.items():
        tm.insert(TagKey(kk), TagValue(vv))
    return tm

tm_dict = OrderedDict((('os', 'android'), ('lang', 'en_us')))
tm = get_tagmap(tm_dict)
mf = MeasureFloat('joules_in', 'Run the joules', 'kW/s')
vv = View('view_name', 'view_desription', (TagKey('os'), TagKey('lang')), mf, DistributionAggregation([1, 10, 100]))

try:
    from opencensus.stats.stats import Stats
    stats
except NameError:
    stats = Stats()
    stats.view_manager.register_view(vv)

stats.view_manager.measure_to_view_map.record(
    tags=tm,
    measurement_map={mf: -12.34},
    timestamp=dt.now(),
)
(vd,) = stats.view_manager.measure_to_view_map._measure_to_view_data_list_map[mf.name]

assert not vd.tag_value_aggregation_data_map

mm = MeasurementMap(stats.view_manager.measure_to_view_map)
mm.measure_float_put(mm, -12.34)
mm.record(tm)
(vd,) = stats.view_manager.measure_to_view_map._measure_to_view_data_list_map[mf.name]

assert not vd.tag_value_aggregation_data_map
```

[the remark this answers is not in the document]
mayurkale22: Correct me if I am wrong. MeasureToViewMap is an internal class, users shouldn't access that directly. As per your example, it is accessible by ```view_manager.measure_to_view_map```, but I think view_manager should be only resposible for registerView and getView.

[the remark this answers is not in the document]
c24t: @mayurkale22 and I talked a bit offline: the tl;dr is that `MeasurementMap` should fail to record if a user has *ever* tried to use it to record a negative value.

opencensus/stats/measurement_map.py
```diff
@@ -75,11 +90,27 @@ def measure_put_attachment(self, key, value):
...
         self._attachments[key] = value
 
-    def record(self, tag_map_tags=execution_context.get_current_tag_map()):
+    def record(self, tag_map_tags=None):
         """records all the measures at the same time with a tag_map.
         tag_map could either be explicitly passed to the method, or implicitly
         read from current execution context.
         """
+        if tag_map_tags is None:
+            tag_map_tags = execution_context.get_current_tag_map()
```
c24t: Unrelated to this PR, I'm assuming we want to get this at call time and not module load time.

opencensus/stats/measurement_map.py
```diff
@@ -75,11 +90,27 @@ def measure_put_attachment(self, key, value):
...
+                           "measurements, refusing to record")
+            return
+        for measure, value in self.measurement_map.items():
+            if value < 0:
+                self._invalid = True
+                logger.warning("Dropping values, value to record must be "
+                               "non-negative")
+                logger.info("Measure '{}' has negative value ({}), refusing "
+                            "to record measurements from {}"
+                            .format(measure.name, value, self))
```
c24t: Not that we're doing it other places in the library, but I think it's a good idea to include more context at a more verbose logging level so users can turn up specific loggers to see what's going wrong in detail. Reviewers: let me know what you think.

opencensus/stats/measurement_map.py
```diff
@@ -51,10 +60,16 @@ def attachments(self):
 
     def measure_int_put(self, measure, value):
         """associates the measure of type Int with the given value"""
+        if value < 0:
+            # Should be an error in a later release.
+            logger.warning("Cannot record negative values")
```
c24t: Note that we're breaking with the java impl here in that we don't mark the map invalid until record time.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -164,31 +164,13 @@ def to_metric(self, desc, view):
                                        labels=labels)
         elif isinstance(agg_data,
                         aggregation_data_module.DistributionAggregationData):
+
+            assert(agg_data.bounds == sorted(agg_data.bounds))
```
mayurkale22: We should do same in the stats/stackdriver exporter?

c24t: [Here](https://github.com/census-instrumentation/opencensus-python/blob/19f89e19f167e2f0c55fbb288b228a9e4053f020/opencensus/stats/exporters/stackdriver_exporter.py#L221-L222)? I don't think we need this assertion everywhere we use distribution aggregations, I've only got it in the prometheus exporter since guaranteeing the bounds were sorted let me simplify the code a bit.

opencensus/stats/measurement_map.py
```diff
@@ -75,11 +90,27 @@ def measure_put_attachment(self, key, value):
...
         """
+        if tag_map_tags is None:
+            tag_map_tags = execution_context.get_current_tag_map()
+        if self._invalid:
+            logger.warning("Measurement map has included negative value "
+                           "measurements, refusing to record")
+            return
+        for measure, value in self.measurement_map.items():
+            if value < 0:
+                self._invalid = True
```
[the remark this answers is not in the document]
c24t: I don't have a strong preference, but this is intentional since I don't think it makes sense to raise until the user tries to record.

c24t:
Something to consider for later when we're making other breaking changes: we might want to remove the `MeasurementMap` class altogether and replace it with a `record(tags, measurement, timestamp, attachments)` API method where `measurement` is a d `Measure`-keyed dict.

opencensus/stats/measurement_map.py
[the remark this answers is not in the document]
mayurkale22: > I don't think it makes sense to raise until the user tries to record.

Agreed. But my point was if you set ```self._invalid = True``` in the put method, during record time you just have to check ```if (self._invalid == True)``` instead of iterating on map.

opencensus/stats/aggregation_data.py
```diff
@@ -126,12 +131,17 @@ def __init__(self,
...
+            assert bounds == list(sorted(set(bounds)))
 
         if counts_per_bucket is None:
             counts_per_bucket = [0 for ii in range(len(bounds) + 1)]
-        elif len(counts_per_bucket) != len(bounds) + 1:
-            raise ValueError("counts_per_bucket length does not match bounds "
-                             "length")
+        else:
+            assert all(cc >= 0 for cc in counts_per_bucket)
+            assert len(counts_per_bucket) == len(bounds) + 1
```
zoidbergwill: Asserts aren't guaranteed to run in production, since some people might run Python without assertions, but we always raise ValueErrors or AssertionErrors for these, I think.

zoidbergwill: Sorry, I should have elaborated. Caveat #1 from [this blog post](https://dbader.org/blog/python-assert-tutorial) mentions it. The [`assert`](https://docs.python.org/3/reference/simple_stmts.html#assert) docs also mention it, but not necessarily emphasising how they shouldn't be used.

c24t: I tried to use asserts in a way consistent with the blog. An assertion failure here should mean a _programming error_ since we only ever expect to call `DistributionAggregationData.__init__` from `DistributionAggregation.__init__`, where we actually do raise a `ValueError` if the user gives us unsorted bounds.

I.e. precondition checks for API methods should raise `Type`/`ValueError`s (and etc.), but precondition checks in internal-only code can assert instead since bad user input should never cause them to fail. The logic doesn't change if the assertions are removed, and they should only fail if there's a bug in the library.

That said, there's no bright line separating the API from the internals here, and we don't lose anything by changing these to `ValueErrors` (other than signaling the difference between expected and unexpected errors). If you've got a strong preference I don't mind changing it.

opencensus/stats/aggregation.py
```diff
@@ -123,11 +128,26 @@ class DistributionAggregation(BaseAggregation):
...
+
+    def __init__(self,
+                 boundaries=None,
+                 distribution=None,
+                 aggregation_type=Type.DISTRIBUTION):
+        if boundaries:
+            if not all(boundaries[ii] < boundaries[ii + 1]
+                       for ii in range(len(boundaries) - 1)):
+                raise ValueError("bounds must be sorted in increasing order")
+            for ii, bb in enumerate(boundaries):
```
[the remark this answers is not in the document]
c24t: This loop should normally break on the first item, I think the slight readability improvement is worth it here.

Issue #181, opened 2018-05-24
tracer: option to use multiple exporters instead of just one


odeke-em:
I am just wrapping up some presentations and noticed that currently all the signatures require just one exporter, unless am mistaken. I'd like to use Zipkin, Stackdriver Tracing and Jaeger at the same time.

I believe I should be able to register infinitely many exporters and have the same data exported to each one of them -- as this accomplishes one of the benefits of OpenCensus i.e. multiplicity of backends and no single vendor lock-in.

Thank you.

chingor13:
I would suggest creating a MultiExporter that implements the base exporter interface and accepts many exporters in the constructor.

liyanhui1228:
@chingor13 Thanks for the suggestion!

odeke-em:
Great suggestion @chingor13, thank you! Thank you @liyanhui1228 too for taking this on!

flands:
@liyanhui1228 any update on this? very much needed. /cc @c24t

c24t:
@flands I put up an exploratory PR at #397, if this approach works for you we can get (something like) it merged in later.

flands:
@c24t thanks! I will try it out and let you know.

PR #397, opened 2018-11-14
Send to multiple exporters from tracer

c24t:
This is an experimental branch to address #181. It changes `Tracer` such that it can use multiple exporters at once.

### DIFF
--- a/opencensus/trace/tracer.py
+++ b/opencensus/trace/tracer.py
@@ -33,34 +33,48 @@ class Tracer(object):
                     :class:`.AlwaysOnSampler`. The rest options are
                     :class:`.AlwaysOffSampler`, :class:`.FixedRateSampler`.
 
-    :type exporter: :class:`~opencensus.trace.exporters.base.exporter`
-    :param exporter: Instances of exporter objects. Default to
-                     :class:`.Printexporter`. The rest options are
-                     :class:`.Fileexporter`, :class:`.Printexporter`,
-                     :class:`.Loggingexporter`, :class:`.Zipkinexporter`,
-                     :class:`.GoogleCloudexporter`
-    """
+    :type exporters: class:`~opencensus.trace.exporters.base.exporter`
+    :param exporters: Provided for backwards compatibility, new callers should
+                      use the `exporters` arg instead.
+
+    :type propagator: :class:`object`
+    :param propagator: An optional propagator, defaults to a
+                       :class:`.GoogleCloudFormatPropagator.
+
+    :type exporters: list(:class:`~opencensus.trace.exporters.base.exporter`)
+    :param exporters: A list of instances of exporter objects. Default to
+                      :class:`.Printexporter`. The rest options are
+                      :class:`.Fileexporter`, :class:`.Printexporter`,
+                      :class:`.Loggingexporter`, :class:`.Zipkinexporter`,
+                      :class:`.GoogleCloudexporter`
+    """  # noqa
     def __init__(
             self,
             span_context=None,
             sampler=None,
             exporter=None,
-            propagator=None):
+            propagator=None,
+            exporters=None):
         if span_context is None:
             span_context = SpanContext()
 
         if sampler is None:
             sampler = always_on.AlwaysOnSampler()
 
-        if exporter is None:
-            exporter = print_exporter.PrintExporter()
+        if exporters is None:
+            if exporter is None:
+                exporters = [print_exporter.PrintExporter()]
+            else:
+                exporters = [exporter]
+        elif exporter is not None:
+            raise ValueError("One of 'exporter' or 'exporters' must be null")
 
         if propagator is None:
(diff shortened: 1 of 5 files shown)


opencensus/trace/tracer.py
```diff
@@ -33,34 +33,48 @@ class Tracer(object):
...
                     :class:`.AlwaysOffSampler`, :class:`.FixedRateSampler`.
 
-    :type exporter: :class:`~opencensus.trace.exporters.base.exporter`
-    :param exporter: Instances of exporter objects. Default to
-                     :class:`.Printexporter`. The rest options are
-                     :class:`.Fileexporter`, :class:`.Printexporter`,
-                     :class:`.Loggingexporter`, :class:`.Zipkinexporter`,
-                     :class:`.GoogleCloudexporter`
-    """
+    :type exporters: class:`~opencensus.trace.exporters.base.exporter`
```
zoidbergwill: This should be `type exporter` and `param exporter` still, right? Judging from the contents.

opencensus/trace/tracers/context_tracer.py
```diff
@@ -120,7 +120,8 @@ def end_span(self, *args, **kwargs):
         with self._spans_list_condition:
             if cur_span in self._spans_list:
                 span_datas = self.get_span_datas(cur_span)
-                self.exporter.export(span_datas)
+                for exporter in self.exporters:
+                    exporter.export(span_datas)
```
zoidbergwill: If someone tries to send something to multiple exporters, but one is incorrectly configured, we should try export to the others still, right, are we just gonna hope that each `Exporter.emit` swallows the exceptions and doesn't raise any?

flands:
@c24t this appears to work, thanks! note, there are other parts of the library that also need to be updated. For example, opencensus/trace/ext/flask/flask_middleware.py

ocervell:
@c24t looks cool. Does that mean we'll have one background thread / exporter here ?

opencensus/trace/tracers/context_tracer.py
reyang: I remember we had discussion about supporting multiple exporters and propagators, and people in that meeting generally agreed not to expose the interface for multiple exporters/propagators, but let people implement something like AggregatedExporter/Propagator. The idea is that you can implement a wrapper exporter that calls into multiple exporters, and you can decide what's the expected behavior - either abort the others if one failed, or continue.

Can we have a clarification here - do we want to have multiple exports/propagators interface for all languages? Or we want to stick with the aggregated approach?

flands: @reyang both Java and Go libraries support multiple destinations today. From a testing perspective or migration perspective this is a good feature to support. Of course, long term this could be handed off to the agent/collector, but requiring the agent/collector limits flexibility. I would propose supporting multiple destinations.

reyang: @flands thank you! What about multiple propagators?

flands: IMO this is less of a concern -- switching and testing is straightforward and migration path is cutover. Let me check OC service team and get back shortly.

c24t: @reyang I wasn't around for the discussion, can you link me? I don't think this is necessarily the best design, and I'm willing to believe we should add AggregatedExporter/Propagator classes instead.

@flands IMO if the API were closer to the [java client's](https://static.javadoc.io/io.opencensus/opencensus-api/0.17.0/io/opencensus/trace/export/SpanExporter.Handler.html) I think it would make sense to argue for a particular implementation, but as it is I think we have some flexibility here, and don't necessarily need to change the tracer.

opencensus/trace/tracer.py
c24t: It should! good catch.

opencensus/trace/tracers/context_tracer.py
reyang: @c24t I couldn't find the discussion, it was Jul. or Aug. this year.

I remember at that time we were talking about how flush API should be provided, whether the user would need to call flush for each individual exporter (if we support multi), or a top level flush at tracer level should be provided (and should we block the flush API until all exporters have finished flush, what do we do if one exporter failed).

c24t: Want to take the conversation around the design to #181?

c24t:
> Does that mean we'll have one background thread / exporter here ?

We will for each exporter that's using a background thread transporter, but we're blocking between calls to `export`. If we expect users to use many exporters with blocking transports at once we might want to make the calls from (e.g.) an async executor instead.

opencensus/trace/tracers/context_tracer.py
reyang: Yep! #181 seems to be a good place for discussion.

PR #406, opened 2018-11-21
Add stackdriver underflow bucket

### LINKED ISSUE #401: Add underflow bucket for stackdriver stats exporter
#389 removed 0- and negative-value bucket boundaries from distribution aggregations. The stackdriver exporter sends these aggregations to stackdriver as-is, but stackdriver expects all distributions to have a single "underflow" bucket for negative values. 

See https://github.com/census-instrumentation/opencensus-java/pull/1525 for a similar change in the java client.

c24t:
Fixes #401.

### DIFF
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -219,9 +219,14 @@ class StackdriverStatsExporter(base.StatsExporter):
                 # point.value.distribution_value.range.min = agg_data.min
                 # point.value.distribution_value.range.max = agg_data.max
                 bounds = dist_value.bucket_options.explicit_buckets.bounds
+                # Stackdriver expects an "underflow" bucket for samples in
+                # (-inf, 0), DistributionAggregationData has all positive
+                # values.
+                bounds.extend([0])
                 bounds.extend(list(map(float, agg_data.bounds)))
 
                 buckets = dist_value.bucket_counts
+                buckets.extend([0])
                 buckets.extend(list(map(int, agg_data.counts_per_bucket)))
             else:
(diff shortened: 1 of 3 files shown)


opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -219,9 +219,14 @@ def create_time_series_list(self, v_data, option_resource_type,
                 # point.value.distribution_value.range.min = agg_data.min
                 # point.value.distribution_value.range.max = agg_data.max
                 bounds = dist_value.bucket_options.explicit_buckets.bounds
+                # Stackdriver expects an "underflow" bucket for samples in
```
mayurkale22: "The first bucket bound should be 0.0 because the Metrics first bucket is [0, first_bound) but Stackdriver monitoring bucket bounds begin with -infinity (first bucket is (-infinity, 0))." I would use this.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -219,9 +219,14 @@ def create_time_series_list(self, v_data, option_resource_type,
...
+                # (-inf, 0), DistributionAggregationData has all positive
+                # values.
+                bounds.extend([0])
                 bounds.extend(list(map(float, agg_data.bounds)))
 
                 buckets = dist_value.bucket_counts
+                buckets.extend([0])
```
mayurkale22: Better to add comment here also, "The first bucket (underflow bucket) should always be 0 count because the Metrics first bucket is [0, first_bound) but StackDriver distribution consists of an underflow bucket (number 0)."

mayurkale22:
Please fix the build

Issue #403, opened 2018-11-20
Release 0.2.0


c24t:
cc @ocervell 

We've got several significant unreleased changes on master, including #365 which fixes a bug in 0.1.8 that breaks the stackdriver exporter.

At a minimum, before releasing, we need to:
- [x] Manually check integration examples
- [x] Show that this code works to instrument the [GCP microservices demo](https://github.com/GoogleCloudPlatform/microservices-demo)
- [x] Fix #401 

Currently, we tag specific revisions on master as versioned releases. This makes it difficult to apply bugfixes (like #365) to old releases. I think we should consider using [release branches](https://nvie.com/posts/a-successful-git-branching-model/#release-branches) instead, and push minor version updates to PyPI with each new commit to a release branch.

If we move to release branches, we can release the next version as 0.2.0 and plan to release bugfixes as minor version updates (e.g. 0.2.1) while continuing development on master. @reyang, @songy23, what do you think?

c24t:
I split the first release branch into a separate PR (#407), we can track the 0.2.0 release in this issue.

songy23:
I believe this is fixed with https://github.com/census-instrumentation/opencensus-python/releases/tag/v0.2.0.

PR #407, opened 2018-11-21
Separate 0.1.8.x release branch

c24t:
cc @ocervell, addresses #403 

This PR proposes to create a permanent `0.1.8.x-release` branch on this remote, onto which we can cherry-pick bugfixes from `master`. Each change to this branch should be released as a minor version package update on PyPI.

This branch was split from `master` at the `0.1.8` tag (3abb0b6). The commits I've cherry-picked here were chosen to fix Forseti's OC integration for @ocervell.

I'd like to move to release branches for all new releases, but this does mean some extra maintenance work since we'll be maintaining multiple package versions instead of forward-fixing everything. It also means making judgement calls about new changes; ideally it should always be safe to update to a newer minor version of the library, but each change risks introducing a regression.

Let me know what you think!


c24t:
> Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments

CLA passed for each of these commits on master, nothing new to see here.

bogdandrutu:
Please use `git cherry-pick` for individual PRs (commits).

bogdandrutu:
@c24t we don't necessary want the same version, but we want to use the same conventions - create release branch, do minor release (not patch release as we do today), etc.

PR #416, opened 2018-11-29
Format all code with Google style

bogdandrutu:
Run command:
$yapf -i -r --style google .

See https://github.com/google/yapf for details.

Next:
Enable formatter check in our build.

Other changes:
Allow 80 characters per line in flake8.

### DIFF
--- a/nox.py
+++ b/nox.py
@@ -32,17 +32,9 @@ def unit(session, py):
     session.install('-e', '.')
 
     # Run py.test against the unit tests.
-    session.run(
-        'py.test',
-        '--quiet',
-        '--cov=opencensus',
-        '--cov-append',
-        '--cov-config=.coveragerc',
-        '--cov-report=',
-        '--cov-fail-under=97',
-        'tests/unit/',
-        *session.posargs
-    )
+    session.run('py.test', '--quiet', '--cov=opencensus', '--cov-append',
+                '--cov-config=.coveragerc', '--cov-report=',
+                '--cov-fail-under=97', 'tests/unit/', *session.posargs)
 
 
 @nox.session
@@ -66,12 +58,7 @@ def system(session, py):
     session.install('.')
 
     # Run py.test against the system tests.
-    session.run(
-        'py.test',
-        '-s',
-        'tests/system/',
-        *session.posargs
-    )
+    session.run('py.test', '-s', 'tests/system/', *session.posargs)
 
 
 @nox.session
@@ -83,10 +70,9 @@ def lint(session):
     session.interpreter = 'python3.6'
     session.install('flake8')
     session.install('.')
-    session.run(
-        'flake8',
-        '--exclude=opencensus/trace/exporters/gen/opencensus/',
-        'opencensus/')
+    session.run('flake8', '--max-line-length=80',
+                '--exclude=opencensus/trace/exporters/gen/opencensus/',
+                'opencensus/')
 
 
 @nox.session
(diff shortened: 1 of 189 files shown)


nox.py
```diff
@@ -83,19 +70,17 @@ def lint(session):
     session.interpreter = 'python3.6'
     session.install('flake8')
     session.install('.')
-    session.run(
-        'flake8',
-        '--exclude=opencensus/trace/exporters/gen/opencensus/',
-        'opencensus/')
+    session.run('flake8', '--max-line-length=80',
```
c24t: I think it's better to set the line length to 79 in the formatter than force flake8 to match the formatter. This seems like putting the cart before the horse.

c24t:
I considered doing the same thing when I started working on the repo, and I generally run `yapf` on any file I touch in a diff that has preexisting lint errors. The less time spent on formatting, the better. And `yapf` generally produces good looking code, but I don't think we should _enforce_ the use of a formatter, and I don't know that it's a good idea to apply the formatter to the whole repo.

One big problem with using a formatter everywhere is that it tends to mangle generated code, which makes it difficult to get a useful diff when regenerating the files. Formatting the new generate file usually produces a nosier diff than comparing the new and old file both unformatted. See [`common_pb2.py`](https://github.com/census-instrumentation/opencensus-python/pull/416/files#diff-37b570a3b915b3a969b8c2cfa61e0bf9).

Diffs like this that mechanically change a lot of files also incur a maintenance cost since it means digging through another layer of `git blame` to find meaningful changes. I don't mean to say that it's never worth making changes like this, just that the benefit should justify the cost.

To the extent that the point of using a formatter is to prevent lint errors, I think it would be better to enforce flake8 checks more strictly. This PR touches a lot of files with lint errors, but doesn't fix them:

```sh
% python3 -m flake8 $(git diff --name-only master) | wc -l
203
```

`yapf` also doesn't always produce more readable code. I think it's a great idea to use `yapf` as a common formatter, and think we should include a `.yapf.conf` in the repo. But I think enforcing its use causes more harm than good.

c24t:
`yapf`'s predefined google style has some weird defaults, including the 80 char line length (which breaks with PEP-8's 79). 

Here's [my preferred `yapf` config](https://gist.github.com/c24t/ed0477e4cf97384ae565dfbc623a3013), and a [diff again the default google style](https://gist.github.com/c24t/06f53b41fcb48ef3e061a4da405d3105).

nox.py
bogdandrutu: We would like to use google style guide:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length

bogdandrutu:
> yapf's predefined google style has some weird defaults, including the 80 char line length (which breaks with PEP-8's 79).

I prefer to use Google style guide for this repo because we need to import this to google repo which will cause a lot of formatting errors.

> Diffs like this that mechanically change a lot of files also incur a maintenance cost since it means digging through another layer of git blame to find meaningful changes. I don't mean to say that it's never worth making changes like this, just that the benefit should justify the cost.

I agree, but I think we need to enforce a style for the repo at one moment. And if we defer that it will get harder.

> One big problem with using a formatter everywhere is that it tends to mangle generated code, which makes it difficult to get a useful diff when regenerating the files. Formatting the new generated file usually produces a nosier diff than comparing the new and old file both unformatted. See common_pb2.py.

We should make sure we do not enforce formatting on any auto-generated file. So if I missed any file/directory please tell me the list that I need to exclude.

> To the extent that the point of using a formatter is to prevent lint errors, I think it would be better to enforce flake8 checks more strictly. This PR touches a lot of files with lint errors, but doesn't fix them:

If you change the command to 

```bash
python3 -m flake8 --max-line-length=80 $(git diff --name-only master) | wc -l
118 - ~30% being long lines errors because yapf does not split long strings.
```

PR #421, opened 2018-12-01
Make user-agent consistent with Java.

bogdandrutu:


c24t:
The whole UA string after this change would be e.g. `gl-python/2.7.10 grpc/1.8.3 gax/1.4.1 gccl/opencensus-python/0.1.9`, so _more_ consistent with java's, but not an exact match. Is that what you want?

songy23:
I'm curious if we can remove the prefix `gccl/`. If not, the current value should be fine.

bogdandrutu:
@c24t it would be good to drop gccl if possible.

c24t:
We could do it, but it won't be easy. One (ugly, dangerous) option is to patch classes owned by the `google-api-core` and `google-cloud-monitoring`.

Core's [`ClientInfo.to_user_agent`](https://github.com/googleapis/google-cloud-python/blob/b69461c2e6c856783ca5b9b3eaafdc56d8fcc6db/api_core/google/api_core/gapic_v1/client_info.py#L70-L87) defines the UA string. The recommended way for client library developers (that's us!) to change the string is to set `ClientInfo.client_library_version`, which is what I did in #370.

The [`MetricServiceClient`](https://github.com/googleapis/google-cloud-python/blob/b69461c2e6c856783ca5b9b3eaafdc56d8fcc6db/monitoring/google/cloud/monitoring_v3/gapic/metric_service_client.py#L54) has a `client_info` attr that gets expanded into a user agent string and metadata header (`'x-goog-api-client'`) with each RPC call. See the wrapping logic [here](`google-cloud-monitoring`).

We could pass our own `ClientInfo` with an overridden `to_user_agent` to `MetricServiceClient`, but at the risk of mangling stackdriver's metrics and breaking the library when we upgrade cloud core or monitoring. If it's important that we control the user agent, or that it includes a fragment that isn't prefixed with `gccl/`, we might want to take it up with the gcloud library maintainers.

c24t:
@bogdandrutu do you still want to merge this change? It LGTM if so.

bogdandrutu:
So this PR changes from gccl/opencensus/version to gccl/opencensus-python/version. I think this is good enough for the moment.

Issue #423, opened 2018-12-03
Don't require project ID in StackdriverStatsExporter


c24t:
The `StackdriverStatsExporter` shouldn't [require a project ID](https://github.com/census-instrumentation/opencensus-python/blob/4ca0055cfbaed534edc0ff0e951070105e634608/opencensus/stats/exporters/stackdriver_exporter.py#L399-L400) if we can get the project from the environment with e.g. [`google.auth.default`](https://github.com/googleapis/google-cloud-python/blob/2842fa877e29a41b5039a7dc25adf28ebe90ebae/core/google/cloud/_helpers.py#L173-L187). 

`google.cloud.trace.client.Client` gives us this for free, which is why we don't require a project ID in the (trace) `StackdriverExporter`. `monitoring_v3.MetricServiceClient` doesn't, which is presumably why we require one in `StackdriverStatsExporter`.

vkryachko:
`monitoring_v3.MetricServiceClient` does not seem to require project_id or credentials and delegates to `google.auth.default`.

https://github.com/googleapis/google-cloud-python/blob/bea0d8c1d2b66e15d84ca3ab992f28138cf1b751/monitoring/google/cloud/monitoring_v3/gapic/metric_service_client.py#L112

PR #424, opened 2018-12-03
Fix stats example project ID

c24t:
Fixes a wart in the stackdriver stats example by fetching the gcloud project ID from the environment if it exists instead of trying to use the "opencensus-node" project.


bogdandrutu:
Maybe we want to not have any project_id in the example and use the default one as suggested here https://github.com/census-instrumentation/opencensus-python/issues/423

c24t:
That works for me, changed in c24e3d4.

PR #430, opened 2018-12-10
Release 0.1.10

c24t:
This diff applies the fix from #429 to the `v0.1.x` branch and creates a new `v0.1.10` release. It fixes a memory leak in the stackdriver exporter. 

See #412 for details on the last release and creation of the `v0.1.x` release branch.


c24t:
The changes to the exporter and tests come from #429. Not strictly necessary, but I thought it was worth replacing the hardcoded version on this branch too to avoid the risk of not changing both version numbers in a later release.

PR #434, opened 2018-12-26
Enforce str for some django attributes

geobeau:
Some attributes were of type unicode

### DIFF
--- a/opencensus/trace/ext/django/middleware.py
+++ b/opencensus/trace/ext/django/middleware.py
@@ -101,7 +101,7 @@ def _set_django_attributes(span, request):
         span.add_attribute('django.user.id', str(user_id))
 
     if user_name is not None:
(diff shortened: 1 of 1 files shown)


geobeau:
CC: @c24t

c24t:
LGTM, but I'm curious what error you see with unicode attributes. Maybe we should enforce `str`s in `add_attribute`.

geobeau:
A log saying the attribute couldn't be encoded. Maybe we can accept unicode type and cast it as str.

geobeau:
Ah it was coming from Jaeger exporter: https://github.com/census-instrumentation/opencensus-python/blob/b3c298f696108feb7f134b0c5ebdaf1fff8939df/opencensus/trace/exporters/jaeger_exporter.py#L298

PR #374, opened 2018-10-30
Fix: StackDriver exporter fails with "One or more Timeseries could not be written"

mayurkale22:
Fixes #350

PR #438, opened 2019-01-09
Fix stackdriver exporter

colincadams:
The main changes here are:
* Send correct point value type, based on aggregation output type
* Set the correct labels for each time series, only adding a singe point per time series

I updated the unit tests to verify this behavior. The only untested piece (I think) is the changes that I made to the batching logic. I didn't run the integration tests so maybe those will catch it, or I can add some more tests. Wanted to get this sent out for a first pass.

[EDIT]
#374 fixed the above main issues. This now fixes the batching logic and makes some other minor changes, as well as adding/expanding unit tests for the exporter

### DIFF
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -190,7 +190,8 @@ class StackdriverStatsExporter(base.StatsExporter):
                 time_series = next(full_time_series_list, None)
                 if time_series:
                     batch.append(time_series)
-            time_series.append(batch)
+            if batch:
+                time_series_batches.append(batch)
             if len(batch) < batch_size:
                 break
         return time_series_batches
@@ -480,16 +481,6 @@ def remove_non_alphanumeric(text):
     """
     return str(re.sub('[^0-9a-zA-Z ]+', '', text)).replace(" ", "")
 
-
-def as_float(value):
-    """ Converts a value to a float if possible
-          On success, it returns (converted_value, True)
-          On failure, it returns (None, False)
-    """
-    try:
-        return float(value), True
-    except Exception:  # Catch all exception including ValueError
-        return None, False
 
 def metric_labels_from_tags(columns, tag_values):
     return {key: value for key, value in zip(columns, tag_values) if value is not None}
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -248,7 +248,7 @@ class StackdriverStatsExporter(base.StatsExporter):
                 if isinstance(v_data.view.measure, measure.MeasureFloat):
                     point.value.double_value = float(agg.value)
             else:
-                raise Exception("unsupported aggregation type: %s" %
+                raise TypeError("Unsupported aggregation type: %s" %
                                 type(v_data.view.aggregation))
 
             start = datetime.strptime(v_data.start_time, EPOCH_PATTERN)
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -168,8 +168,9 @@ class StackdriverStatsExporter(base.StatsExporter):
         """ It receives an array of view_data object
             and create time series for each value
         """
+        view_data_set = utils.uniq(view_data)
         time_series_batches = self.create_batched_time_series(
-            view_data, MAX_TIME_SERIES_PER_UPLOAD)
+            view_data_set, MAX_TIME_SERIES_PER_UPLOAD)
         for time_series_batch in time_series_batches:
             self.client.create_time_series(
                 self.client.project_path(self.options.project_id),
@@ -241,6 +242,7 @@ class StackdriverStatsExporter(base.StatsExporter):
                     point.value.int64_value = int(agg.sum_data)
                 if isinstance(v_data.view.measure, measure.MeasureFloat):
                     point.value.double_value = float(agg.sum_data)
+            # TODO: Why is not?????
             elif aggregation_type is not aggregation.Type.LASTVALUE:
                 if isinstance(v_data.view.measure, measure.MeasureInt):
                     point.value.int64_value = int(agg.value)
@@ -260,8 +262,7 @@ class StackdriverStatsExporter(base.StatsExporter):
             secs = point.interval.end_time.seconds
             point.interval.end_time.nanos = int((timestamp_end - secs) * 10**9)
 
-            if aggregation_type is aggregation.Type.LASTVALUE:
-                    LastValueAggregationData:  # pragma: NO COVER
+            if aggregation_type is not aggregation.Type.LASTVALUE:
                 if timestamp_start == timestamp_end:
                     # avoiding start_time and end_time to be equal
                     timestamp_start = timestamp_start - 1
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -202,6 +202,7 @@ class StackdriverStatsExporter(base.StatsExporter):
         """ Create the TimeSeries object based on the view data
         """
         time_series_list = []
+        aggregation_type = v_data.view.aggregation.aggregation_type
         tag_agg = v_data.tag_value_aggregation_data_map
         for tag_value, agg in tag_agg.items():
             series = monitoring_v3.types.TimeSeries()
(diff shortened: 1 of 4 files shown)


opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -164,24 +165,32 @@ def upload_stats(self, view_data):
...
-        for i in range(0, len(full_time_series_list), limit):
-            time_series_lists.append(
-                [tsl for tsl in full_time_series_list[i:i + limit]])
-
-        return full_time_series_list
+        full_time_series_list = itertools.chain.from_iterable(
+            self.create_time_series_list(
+                v_data, self.options.resource, self.options.metric_prefix)
+                for v_data in view_data)
+        time_series_batches = []
```
c24t: It looks like you never append `batch`?

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -164,24 +165,32 @@ def upload_stats(self, view_data):
...
+        time_series_batches = self.create_batched_time_series(
+            view_data, MAX_TIME_SERIES_PER_UPLOAD)
+        for time_series_batch in time_series_batches:
             self.client.create_time_series(
                 self.client.project_path(self.options.project_id),
-                time_series_list)
+                time_series_batch)
 
-    def make_request(self, view_data, limit):
+    def create_batched_time_series(self, view_data, batch_size):
```
c24t: I think changing this to use an iterator is a good idea in general, but I'd save the change for a different PR and only include the changes you need to fix the exporter here. It would make the review quicker and reduce the risk of regressions.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -233,23 +231,25 @@ def create_time_series_list(self, v_data, option_resource_type,
...
+                    point.value.double_value = float(agg.sum_data)
+            elif aggregation_type is aggregation.Type.LASTVALUE:
                 if isinstance(v_data.view.measure, measure.MeasureInt):
                     point.value.int64_value = int(agg.value)
-                elif isinstance(v_data.view.measure, measure.MeasureFloat):
+                if isinstance(v_data.view.measure, measure.MeasureFloat):
                     point.value.double_value = float(agg.value)
             else:
-                point.value.string_value = str(tag_value[0])
+                raise Exception("unsupported aggregation type: %s" %
```
colincadams: Is it okay to throw an `Exception` at this point in the export? Seems it would be an artifact of their view being improperly configured, and should've already been thrown in `create_metric_descriptor`. Even if we shouldn't thrown an exception, it didn't make sense to me to set the value to that of the first tag.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -262,8 +262,7 @@ def create_time_series_list(self, v_data, option_resource_type,
             secs = point.interval.end_time.seconds
             point.interval.end_time.nanos = int((timestamp_end - secs) * 10**9)
 
-            if type(agg) is not aggregation.aggregation_data.\
-                    LastValueAggregationData:  # pragma: NO COVER
+            if aggregation_type is not aggregation.Type.LASTVALUE:
```
colincadams: Do we skip `LastValue` aggregations here because they don't cover a range, so it's okay if `start` and `end` are the same? I'd like to add a comment if we know why this is done.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -432,10 +431,10 @@ def get_task_value():
     """ getTaskValue returns a task label value in the format of
      "py-<pid>@<hostname>".
     """
-    task_value = "py@" + str(os.getpid())
     hostname = platform.uname()[1]
-    task_value += hostname if hostname is not None else "localhost"
-    return task_value
+    if not hostname:
```
colincadams: I know that this also bloats the PR, but the existing implementation didn't conform to the comment. Happy to revert or send as a separate change.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -471,25 +470,13 @@ def set_metric_labels(series, view, tag_values):
             "TagKeys and TagValues don't have same size."
         )  # pragma: NO COVER
 
-    for ii, tag_value in enumerate(tag_values):
-        if tag_value is not None:
-            metric_label = remove_non_alphanumeric(view.columns[ii])
-            series.metric.labels[metric_label] = tag_value
+    for key, value in zip(view.columns, tag_values):
```
colincadams: I thought using `zip` was more readable here, but happy to revert if desired.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -471,25 +470,13 @@ def set_metric_labels(series, view, tag_values):
...
+        if value is not None:
+            series.metric.labels[remove_non_alphanumeric(key)] = value
     series.metric.labels[OPENCENSUS_TASK] = get_task_value()
 
 
 def remove_non_alphanumeric(text):
     """ Remove characters not accepted in labels key
     """
     return str(re.sub('[^0-9a-zA-Z ]+', '', text)).replace(" ", "")
```
colincadams: One other thing I noticed @c24t, this removes `_` from labels but they are valid characters and I think are important visually. I'll open a separate PR for this.

c24t: And the same exporter in go replaces non-alphanumerics with underscores: https://github.com/c24t/opencensus-go-exporter-stackdriver/blob/281244432d1ddf7f0566cfd0964d79c2b6cea7c2/stats.go#L414-L423. Definitely a problem for another PR.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -233,23 +231,22 @@ def create_time_series_list(self, v_data, option_resource_type,
...
-                    agg, aggregation.aggregation_data
-                    .LastValueAggregationData):  # pragma: NO COVER
+            elif aggregation_type is aggregation.Type.SUM:
+                if isinstance(v_data.view.measure, measure.MeasureInt):
+                    # TODO: Add implementation of sum aggregation that does not
+                    # store it's data as a float.
+                    point.value.int64_value = int(agg.sum_data)
+                if isinstance(v_data.view.measure, measure.MeasureFloat):
+                    point.value.double_value = float(agg.sum_data)
+            elif aggregation_type is aggregation.Type.LASTVALUE:
```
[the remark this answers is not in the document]
c24t: Nope, that looks like a bug to me too.

opencensus/stats/exporters/stackdriver_exporter.py
c24t: No idea from previous PRs, you may know better what stackdriver expects here.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -168,40 +169,40 @@ def upload_stats(self, view_data):
             and create time series for each value
         """
         view_data_set = utils.uniq(view_data)
```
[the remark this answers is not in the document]
c24t: I changed this from @mayurkale22's original diff in https://github.com/census-instrumentation/opencensus-python/pull/374/commits/e6f1d654b0b85ed8aea5d660731d1458833626a7#diff-3c686c24a85a4df5a88f2c2d4683bf40R170. My understanding is that `view_data` shouldn't have duplicate `ViewData`s, but we don't enforce it anywhere.

@mayurkale22 may have more info here.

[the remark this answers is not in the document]
colincadams: Gotcha, happy to leave it for now I wasostly just curious!

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -168,40 +169,40 @@ def upload_stats(self, view_data):
...
+        while True:
+            batch = []
+            for _ in range(batch_size):
+                time_series = next(full_time_series_list, None)
+                if time_series is not None:
+                    batch.append(time_series)
+            if batch:
+                time_series_batches.append(batch)
+            if len(batch) < batch_size:
+                break
```
c24t: This batching logic comes up often enough to justify adding a function for it. I moved it to `utils.window` in ad81279, let me know what you think.

opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -471,25 +470,13 @@ def set_metric_labels(series, view, tag_values):
...
     series.metric.labels[OPENCENSUS_TASK] = get_task_value()
 
 
 def remove_non_alphanumeric(text):
     """ Remove characters not accepted in labels key
     """
     return str(re.sub('[^0-9a-zA-Z ]+', '', text)).replace(" ", "")
-
-
-def as_float(value):
```
c24t: Good call removing this.

colincadams:
Yeah that looks great! Good call.

I'll send the subsequent changes later today

Issue #349, opened 2018-10-14
stats/prometheus: running the example fails to send data to the Prometheus UI


PikBot:
Hi, I am trying to collect metrics using the Prometheus as exporter. For now I am just playing around with the [example](https://github.com/census-instrumentation/opencensus-python/blob/master/examples/stats/exporter/prometheus.py), but I am unable to collect any data. I get all zeroes. 
![screen shot 2018-10-15 at 02 14 12](https://user-images.githubusercontent.com/35295607/46921943-9d836e00-d01f-11e8-85b9-ed45bcae41ef.png)
I am running the example as it is, except for uncommenting these [lines](https://github.com/census-instrumentation/opencensus-python/blob/master/examples/stats/exporter/prometheus.py#L67). 

Version details:
opencensus==0.1.8
prometheus-client==0.3.1

I am running prometheus using docker. 

Thanks for any help!

songy23:
Can you move these lines before the `while` loop to see if the `measurements` got recorded:
https://github.com/census-instrumentation/opencensus-python/blob/master/examples/stats/exporter/prometheus.py#L71-L76

PikBot:
Hi @songy23, yes the measurements are getting recorded, I get this:
![screen shot 2018-10-16 at 00 24 18](https://user-images.githubusercontent.com/35295607/46971441-91f47d80-d0d9-11e8-90c7-2160bff0d1ef.png)

odeke-em:
Kindly pinging y'all as this issue affects anyone using the Prometheus exporter.

c24t:
Thanks @odeke-em, I'll take a look. This example is actually just supposed to... hang forever?

odeke-em:
The idea is that one has to type into it and hit enter or you can do
```shell
cat file | python example.py
```

odeke-em:
@c24t please ignore my previous message, I was referring to the example for the website.

The Python Prometheus example in this repo shouldn't hang forever AFAIK.

odeke-em:
If I add a sleep before main exits to allow Prometheus the ability to scrape the endpoint, it prints out some crash like this https://gist.github.com/odeke-em/ef0760b35263a360b482e651413d7bec

songy23:
I see - Prometheus only allows alphanumeric, colon and underscore in the metric name (https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels), so the view name in the sample code is not great ("my.org/views/video_size").

[...]

PR #440, opened 2019-01-10
Fix Prometheus Stats Exporter.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/349.

### DIFF
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -16,7 +16,7 @@ from prometheus_client import start_http_server
 from prometheus_client.core import CollectorRegistry
 from prometheus_client.core import GaugeMetricFamily
 from prometheus_client.core import CounterMetricFamily
-from prometheus_client.core import UnknownMetricFamily
+from prometheus_client.core import UntypedMetricFamily
 from prometheus_client.core import HistogramMetricFamily
 from prometheus_client.core import REGISTRY
 from opencensus.stats.exporters import base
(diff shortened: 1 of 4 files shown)


opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -356,12 +343,3 @@ def view_name(namespace, view):
     if namespace != "":
         name = namespace + "_"
     return name + view.name
-
-
-def view_signature(namespace, view):
```
songy23: IMO this is not needed. `View` and `ViewData` can be directly indexed by its name since view names are unique.

songy23:
Finally got the exporter working:
oc-python-prometheus

@c24t This PR is ready for review now. PTAL when you got time.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -201,30 +214,17 @@ def to_metric(self, desc, view):
...
-        registered = {}
-        for sign in self.registered_views:
-            registered[sign] = self.registered_views[sign]
-        for v_data in list(self.view_data):  # pragma: NO COVER
-            if not isinstance(v_data, str):
-                signature = view_signature(self.options.namespace, v_data.view)
-                desc = self.registered_views[signature]
-                metric = self.to_metric(desc,
-                                        self.view_data[v_data].view)
+        for v_name in list(self.view_name_to_data_map):
```
c24t: I know this is a mechanical change, but since you're touching it: we can lose the `list` here.

songy23: SGTM, and thanks for fixing this.

PR #443, opened 2019-01-14
Bump django to 1.11.18

c24t:


c24t:
This was to address [CVE-2019-3498](https://www.djangoproject.com/weblog/2019/jan/04/security-releases/). Note that This library wasn't affected by the bug, and in any case only depends on django for django integration tests and examples.

PR #447, opened 2019-01-15
Modify get_task_value to match comment.

colincadams:
This also has it check for an empty string in `uname` instead of `None`,
as the documentation states it returns an empty string if it can't be found.

### DIFF
--- a/opencensus/stats/exporters/stackdriver_exporter.py
+++ b/opencensus/stats/exporters/stackdriver_exporter.py
@@ -17,6 +17,7 @@ import logging
 import os
 import platform
 import re
+import string
 
 from datetime import datetime
 from google.api_core.gapic_v1 import client_info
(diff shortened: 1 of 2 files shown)


opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -421,10 +421,10 @@ def get_task_value():
     """ getTaskValue returns a task label value in the format of
      "py-<pid>@<hostname>".
     """
-    task_value = "py@" + str(os.getpid())
     hostname = platform.uname()[1]
-    task_value += hostname if hostname is not None else "localhost"
-    return task_value
+    if not hostname:
+        hostname = "localhost"
```
c24t: I'm surprised we default to localhost, but I know you didn't start the fire here.

PR #450, opened 2019-01-16
Release 0.1.11

### LINKED ISSUE #441: Release 0.1.11 with stackdriver trace exporter fixes
This should include the fix for #425. See #439.

c24t:
This diff applies the fix from #425 to the `v0.1.x` branch and creates a new `v0.1.11` release. It fixes a bug in the stackdriver exporter that caused each span to be exported separately.

This is also the first release to include a changelog. I think it's worth the effort to maintain a changelog, especially as we start supporting more than one (minor) version of the library at a once.

Fixes #441. See #412 for info on the release branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+
+## 0.1.11
+Released 2018-01-16
+
+- Fix a bug in the stackdriver exporter that caused spans to be exported
+  individually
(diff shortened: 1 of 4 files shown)


c24t:
CLA check passed on the original PR.

changelog.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
```
songy23: nit: better to use all capital case here, similar to CONTRIBUTING.md etc.

PR #452, opened 2019-01-17
Remove deprecated transports

### LINKED ISSUE #386: DRY up duplicate transport code
#256 copied the `transports` package from `trace/` into `common/`. We now have two nearly-identical copies of the code. See [`common/transports`](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/common/transports/__init__.py) and [`trace/exporters/transports`](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/exporters/transports/__init__.py).

Before merging these we need to check that the [new default options for stats transports](https://github.com/census-instrumentation/opencensus-python/pull/385) will work for tracing.

kornholi:
Closes #386.


c24t:
Fixes #386.

c24t:
Besides deleting the duplicate code this changes the batching for spans from 10/1sec to 200/60s and changes the way the queue is flushed from #384.

The new flushing behavior hasn't caused any problems for stats, and should be safe for traces. I'm interested to know how you the batch size change though. The new batch size is consistent with other clients, but it means that users sending between 200 and 600 spans/traces per minute now have to change the default.

kornholi:
We've been running with 200/60s for a couple months without issues, as the previous default ran through quota immediately. Perhaps we should bump up the batch size to 600?

kornholi:
Either way, I think we need to update the code to submit as many batches as possible per wait period. Otherwise services under high load will buffer stats/traces until they run out of memory, as only 600 entries can be submitted per minute.

c24t:
Bumping the batch size up sounds good to me.

PR #453, opened 2019-01-17
Stats: Fix recording with None TagMap.

songy23:
The `record` method takes an optional `TagMap`. If `TagMap` is present, recorded measurements would be broken down by the key-value pairs in it. If not, recorded measurements are dimensionless. See also https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Record.md#recording-stats.

Currently if we record measurements with none `TagMap`, an error will be raised:
```
tag_values = self.get_tag_values(tags=context.map,
AttributeError: 'NoneType' object has no attribute 'map'
```
Fix this bug by checking if `TagMap` is `None` before using it.

### DIFF
--- a/opencensus/stats/view_data.py
+++ b/opencensus/stats/view_data.py
@@ -83,7 +83,11 @@ class ViewData(object):
 
     def record(self, context, value, timestamp, attachments=None):
         """records the view data against context"""
-        tag_values = self.get_tag_values(tags=context.map,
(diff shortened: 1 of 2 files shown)


songy23:
@c24t Updated PR description, PTAL.

c24t:
Thanks @songy23!

songy23:
Thanks!

PR #462, opened 2019-01-22
Add Jaeger host/port settings

tpyo:
The Django middleware currently [doesn't use any Jaeger host settings](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/flask/flask_middleware.py#L149-L151).  

This PR applies the same settings used in the Flask middleware, with the addition of agent host/port.

I'd like to add the agent settings to the Flask middleware too, but will do so in another PR.


opencensus/trace/ext/django/middleware.py
```diff
@@ -155,10 +160,23 @@ def __init__(self, get_response=None):
...
+            _jaeger_agent_host_name = settings.params.get(
+                JAEGER_EXPORTER_AGENT_HOST_NAME, 'localhost')
+            _jaeger_agent_port = settings.params.get(
+                JAEGER_EXPORTER_AGENT_PORT, 6831)
             self.exporter = self._exporter(
                 service_name=_service_name,
-                transport=transport)
+                host_name=_jaeger_host_name,
+                port=_jaeger_port,
+                agent_host_name=_jaeger_agent_host_name,
```
reyang: Do we want agent_host_name and agent_port in Flask Middleware?

tpyo: Yep, can add to this PR if you like?

PR #467, opened 2019-01-24
Metrics point conversion

c24t:
This PR adds `to_point` functions to each stats aggregation data class and fixes an unrelated bug in `DistributionAggregationData` exemplars. This lays the groundwork for `MetricProducer`s, which will let us convert stats data into metrics data models for export.

Addresses #335.


opencensus/stats/aggregation_data.py
```diff
@@ -123,34 +162,37 @@ def __init__(self,
...
         self._counts_per_bucket = counts_per_bucket
-        self._bounds = bucket_boundaries.BucketBoundaries(
-            boundaries=bounds).boundaries
-        bucket = 0
-        for _ in self.bounds:
-            bucket = bucket + 1
-
-        # If there is no histogram, do not record an exemplar
-        self._exemplars = \
-            {bucket: exemplars} if len(self._bounds) > 0 else None
```
c24t: I'm assuming this is a typo and this line should have been indented. Score one for semantic whitespace detractors.

tests/unit/stats/test_aggregation_data.py
```diff
@@ -205,18 +272,17 @@ def test_constructor_with_exemplar(self):
...
             bounds=bounds)
 
-        self.assertEqual(1, dist_agg_data.mean_data)
-        self.assertEqual(0, dist_agg_data.count_data)
-        self.assertEqual(0, dist_agg_data.min)
-        self.assertEqual(1, dist_agg_data.max)
-        self.assertEqual(sum_of_sqd_deviations,
-                         dist_agg_data.sum_of_sqd_deviations)
-        self.assertEqual([1, 1, 1], dist_agg_data.counts_per_bucket)
-        self.assertEqual([exemplar_1, exemplar_2], dist_agg_data.exemplars[2])
```
c24t: This was asserting the wrong behavior: attaching all exemplars to the last bucket.

opencensus/metrics/export/metric_descriptor.py
```diff
@@ -84,13 +83,13 @@ class MetricDescriptorType(object):
...
     def to_type_class(cls, metric_descriptor_type):
         type_map = {
-            cls.GAUGE_INT64: int,
-            cls.GAUGE_DOUBLE: float,
-            cls.GAUGE_DISTRIBUTION: ValueDistribution,
-            cls.CUMULATIVE_INT64: int,
-            cls.CUMULATIVE_DOUBLE: float,
-            cls.CUMULATIVE_DISTRIBUTION: ValueDistribution,
-            cls.SUMMARY: ValueSummary
+            cls.GAUGE_INT64: value.ValueLong,
```
reyang: I have a general question, would this hurt the performance as we need to build the dictionary each time the function gets called (and optimization doesn't work since `cls` is parameterized)?

c24t: That's a good point. I assumed it was cheap enough to do that I didn't bother trying to optimize it, but this will get called on every stat conversion and there's no reason not to move this up into the class. Changed in 7a43a50.

c24t: It'd be interesting to profile the library on an example app to see what other easy improvements we could make.

opencensus/stats/aggregation_data.py
```diff
@@ -60,6 +75,18 @@ def sum_data(self):
...
         return self._sum_data
 
+    def to_point(self, timestamp):
+        """Get a Point conversion of this aggregation.
+
+        :type timestamp: :class: `datetime.datetime`
+        :param timestamp: The time to report the point as having been recorded.
+
+        :rtype: :class: `opencensus.metrics.export.point.Point`
+        :return: a :class: `opencensus.metrics.export.value.ValueLong`-valued
```
mayurkale22: Minor nit -> ```s/ValueLong/ValueDouble```

tests/unit/stats/test_aggregation_data.py
```diff
@@ -448,3 +517,39 @@ def test_increment_bucket_count(self):
...
+            exemplars=[None, ex_9, ex_99, None],
+        )
+        converted_point = dist_agg_data.to_point(timestamp)
+        self.assertTrue(isinstance(converted_point.value,
+                                   value.ValueDistribution))
+        self.assertEqual(converted_point.value.count, 99)
+        self.assertEqual(converted_point.value.sum, 4950)
+        self.assertEqual(converted_point.value.sum_of_squared_deviation,
+                         80850.0)
+        self.assertEqual([bb.count for bb in converted_point.value.buckets],
```
mayurkale22: Can you please add check for ```bucket_options```?

c24t: Good call, f9d64fb.

opencensus/metrics/export/metric_descriptor.py
reyang: Good point. We should have some level of performance test in the CI.

opencensus/stats/aggregation_data.py
c24t: Thanks, fixed in 9bbc9f4.

PR #469, opened 2019-01-25
Metrics view data conversion

c24t:
This PR adds a utility function to convert `ViewData`s into `Metric`s, following up on #467.

@mayurkale22 I'm sending this your way for early review. We could merge this change safely by itself, or extend it to include:
- metric conversions in the view manager and measure-to-view map
- metric producers

Edit: done:
- cache metric descriptor on view data

Addresses #335.

### DIFF
--- a/opencensus/stats/measure_to_view_map.py
+++ b/opencensus/stats/measure_to_view_map.py
@@ -58,7 +58,7 @@ class MeasureToViewMap(object):
         if view_data_list is not None:
             for view_data in view_data_list:
                 if view_data.view.name == view_name:
-                    view_data_copy = copy.deepcopy(view_data)
+                    view_data_copy = copy.copy(view_data)
                     view_data_copy.end()
                     return view_data_copy
 
--- a/opencensus/stats/measure_to_view_map.py
+++ b/opencensus/stats/measure_to_view_map.py
@@ -55,12 +55,21 @@ class MeasureToViewMap(object):
 
         view_data_list = self._measure_to_view_data_list_map.get(
             view.measure.name)
-        if view_data_list is not None:
-            for view_data in view_data_list:
-                if view_data.view.name == view_name:
-                    view_data_copy = copy.copy(view_data)
-                    view_data_copy.end()
-                    return view_data_copy
+
+        if not view_data_list:
+            return None
+
+        for view_data in view_data_list:
+            if view_data.view.name == view_name:
+                break
+        else:
+            return None
+
+        view_data_copy = copy.copy(view_data)
+        tvdam_copy = copy.deepcopy(view_data.tag_value_aggregation_data_map)
+        view_data_copy._tag_value_aggregation_data_map = tvdam_copy
+        view_data_copy.end()
+        return view_data_copy
 
     def filter_exported_views(self, all_views):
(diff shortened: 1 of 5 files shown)


opencensus/stats/measure_to_view_map.py
```diff
@@ -58,7 +58,7 @@ def get_view(self, view_name, timestamp):
         if view_data_list is not None:
             for view_data in view_data_list:
                 if view_data.view.name == view_name:
-                    view_data_copy = copy.deepcopy(view_data)
+                    view_data_copy = copy.copy(view_data)
```
c24t: @songy23 I'm assuming the `deepcopy` here was defensive, so the returned `ViewData` had a separate `_tag_value_aggregation_data_map`. But this is also creating new `View`s (and associated measures and aggregations) on each call, which I don't think you want.

Changing this to `copy` means that the "copies" returned here share the aggregation data map. A better solution might be to implement `ViewData.__copy__` or come up with a new immutable view data obj for export.

opencensus/stats/view.py
```diff
@@ -33,13 +40,19 @@ class View(object):
...
     def __init__(self, name, description, columns, measure, aggregation):
         self._name = name
         self._description = description
         self._columns = columns
         self._measure = measure
         self._aggregation = aggregation
 
+        # Cache the converted MetricDescriptor here to avoid creating it each
+        # time we convert a ViewData that realizes this View into a Metric.
+        self._md_cache_lock = threading.Lock()
```
c24t: Other stats classes are generally not threadsafe. I don't know that it's worth opening this can of worms in this PR.

opencensus/stats/measure_to_view_map.py
songy23: > I'm assuming the deepcopy here was defensive, so the returned ViewData had a separate _tag_value_aggregation_data_map.

That's correct.

> Changing this to copy means that the "copies" returned here share the aggregation data map. 

Ideally the `ViewData` returned should be immutable, i.e:
```python
vd1 = stats.get_view("some view")  // vd1 has data 1.0
vd1.record(1.0) // vd1 now has data 2.0
vd2 = stats.get_view("some view") // vd2 should still have data 1.0
```
So at least the `_tag_value_aggregation_data_map` should be deeply copied.

> A better solution might be to implement ViewData.__copy__ or come up with a new immutable view data obj for export.

+1 for the latter (using immutable view data obj for export). This is exactly what we do in Java.

c24t: Thanks @songy23. I changed this to deep-copy the aggregation map only in 5ad48a3. `MeasureToViewMap` would benefit from a total rewrite.

opencensus/stats/measure_to_view_map.py
```diff
@@ -55,12 +55,21 @@ def get_view(self, view_name, timestamp):
...
-                    return view_data_copy
+
+        if not view_data_list:
+            return None
+
+        for view_data in view_data_list:
+            if view_data.view.name == view_name:
+                break
+        else:
+            return None
```
c24t: I rewrote this for clarity, the only behavior change is the copy below. Note that we're not using `timestamp` here, there are likely other bugs hiding in this class.

PR #241, opened 2018-07-30
Add mysqldb integration

pitabwire:
integration to enable instrumentation of mysql queries when using mysqldb library to access the database

PR #472, opened 2019-01-30
Add MySQLdb integration

tpyo:
This PR adds a test to the MySQLdb integration additions in #241 by @pitabwire.

### DIFF
--- a/opencensus/trace/config_integration.py
+++ b/opencensus/trace/config_integration.py
@@ -17,7 +17,7 @@ import logging
 
 log = logging.getLogger(__name__)
 
-SUPPORTED_INTEGRATIONS = ['httplib', 'mysql', 'postgresql', 'pymysql',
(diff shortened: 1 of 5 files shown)


tpyo:
@reyang Happy to pick this up again but wasn't sure what to do about the `mysql_config` requirement for the mysqlclient lib.  Do you have any suggestions?  Maybe it should just be mocked?

reyang:
@tpyo Yep! Actually we are splitting the contrib packages, I believe this would solve the dependency problem (by having `opencensus-ext-mysqldb`, which has a package dependency on `mysqlclient`). Please refer to https://github.com/census-instrumentation/opencensus-python/tree/master/contrib. Please give it a try, I would love to hear your feedbacks!

tpyo:
@reyang  Ah, that's perfect!  Thanks, will give it a try soon.

PR #474, opened 2019-01-30
span context: set the default recorded flag to 0, and fix default trace options bug

### LINKED ISSUE #159: Missing support for "forced tracing"

### LINKED ISSUE #298: All samplers are ignored as the default span context is force enabled
Since #159 was merged, sampler decisions are ignored (even AlwaysOffSampler) as the default span context is force enabled. It intended to solve a valid use-case but now every span context is force enabled. Unless you manually send a span context with `o=0`.

A quick and dirty solution could be changing `trace_options.DEFAULT` and `span_context.DEFAULT_OPTIONS` to `'0'`. Better yet, the default context should set `o` to `int(sampler.should_sample(...))`. However, neither solves that `o=0` does not disable tracing.

wsabransky:
Closes #298 .

Currently, the default trace option for a new trace context is to set `enabled` to true. This causes all samplers applied to be ignored [here](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/tracer.py#L76). We should keep the parent sampling decision if it is true, but ignore it if false. A parent shouldn't be able to force a downstream service to not sample.

Also fixes bug described [here](https://github.com/census-instrumentation/opencensus-python/pull/474#issuecomment-469915775).


songy23:
Please rebase and fix lint.

songy23:
> The lint error appears to be wholly unrelated to this PR, in a file not touched. Do you want it fixed in this change?

I see, For some reason lint didn't fail on my local machine. For now please ignore the error and just rebase against master.

bogdandrutu:
@wsabransky any chance we can get some tests :)

wsabransky:
@bogdandrutu @c24t Sorry about the delay, I'll add some tests today.

wsabransky:
@c24t @bogdandrutu Okay, apparently there were already tests that cover this case, but they weren't included by the test runner because of how they were named.

Additionally, a scoping problem was causing the default trace options to be overwritten at runtime in some cases.

wsabransky:
@c24t any chance we could get this merged?

c24t:
@wsabransky it looks like this change breaks a jaeger integration test. I may have time to check it out later, but it can't merge as-is.

wsabransky:
@c24t looks like it was just a fixture that needed to be updated to reflect the new behavior in this PR. I've made the change.

c24t:
Thanks @wsabransky!

Issue #468, opened 2019-01-25
exporter/stats/prometheus: apply sanitization for metric names replacing "/" with "_"


PikBot:
When metrics have "/" or any other unpleasant character, this Prometheus exporter crashes e.g.
```python
latency_view = view_module.View("demo/latency", "The distribution of the latencies",
                                [key_method],
                                m_latency_ms,
                                # Latency in buckets:
                                # [>=0ms, >=25ms, >=50ms, >=75ms, >=100ms, >=200ms, >=400ms, >=600ms, >=800ms, >=1s, >=2s, >=4s, >=6s]
                                aggregation_module.DistributionAggregation([0, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000]))
```

Other languages handle this gracefully and provide sanitization to translate such characters into "_" such as with OpenCensus Go https://github.com/census-instrumentation/opencensus-go/blob/4baeb639308e8502f1bd4d713236b7978ab8f125/exporter/prometheus/prometheus.go#L253

whose code is in https://github.com/census-instrumentation/opencensus-go/blob/4baeb639308e8502f1bd4d713236b7978ab8f125/internal/sanitize.go#L22-L50
```go
const labelKeySizeLimit = 100

// Sanitize returns a string that is trunacated to 100 characters if it's too
// long, and replaces non-alphanumeric characters to underscores.
func Sanitize(s string) string {
	if len(s) == 0 {
		return s
	}
	if len(s) > labelKeySizeLimit {
		s = s[:labelKeySizeLimit]
	}
	s = strings.Map(sanitizeRune, s)
	if unicode.IsDigit(rune(s[0])) {
		s = "key_" + s
	}
	if s[0] == '_' {
		s = "key" + s
	}
	return s
}

// converts anything that is not a letter or digit to an underscore
func sanitizeRune(r rune) rune {
	if unicode.IsLetter(r) || unicode.IsDigit(r) {
		return r
	}
	// Everything else turns into an underscore
	return '_'
}
```

For a full error stack, given the source code in this gist https://gist.github.com/PikBot/902bd205d53081d373334d0e1430e772
please see:
```
Exception happened during processing of request from ('127.0.0.1', 64332)
Traceback (most recent call last):
  File "/Users/prakritibansal/anaconda2/lib/python2.7/SocketServer.py", line 596, in process_request_thread
    self.finish_request(request, client_address)
[...]

PR #475, opened 2019-01-31
Exporter/Prometheus: Sanitize metric name and key.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/468.

### DIFF
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -24,6 +24,8 @@ from opencensus.common.transports import sync
 from opencensus.stats import aggregation_data as aggregation_data_module
 from opencensus.stats.exporters import base
 
+import re
+
 
 class Options(object):
     """ Options contains options for configuring the exporter.
@@ -125,7 +127,7 @@ class Collector(object):
         if v_name not in self.registered_views:
             desc = {'name': v_name,
                     'documentation': view.description,
-                    'labels': list(view.columns)}
+                    'labels': list(map(sanitize, view.columns))}
             self.registered_views[v_name] = desc
             self.registry.register(self)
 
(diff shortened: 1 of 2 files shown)


opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -345,4 +347,11 @@ def get_view_name(namespace, view):
...
         name = namespace + "_"
-    return name + view.name
+    return sanitize(name + view.name)
+
+
+def sanitize(key):
+    """ sanitize the given metric name or label according to Prometheus rule.
+    Replace all characters other than [A-Za-z0-9_] with '_'.
+    """
+    return re.sub("[^a-zA-Z0-9_]", "_", key)
```
reyang: Suggest to pre-compile the regular expression for performance given this will be applied for all columns.

For example: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/tracestate.py#L25

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -345,4 +347,14 @@ def get_view_name(namespace, view):
     name = ""
     if namespace != "":
         name = namespace + "_"
-    return name + view.name
+    return sanitize(name + view.name)
+
+
+_PROMETHEUS_FORMAT_RE = re.compile('[^a-zA-Z0-9_]')
```
odeke-em: The more appropriate/correct and concise regex would perhaps be:
  `re.compile('[^\w]', re.UNICODE|re.IGNORECASE)`

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -345,4 +347,14 @@ def get_view_name(namespace, view):
...
+
+
+def sanitize(key):
+    """ sanitize the given metric name or label according to Prometheus rule.
+    Replace all characters other than [A-Za-z0-9_] with '_'.
+    """
+    return re.sub(_PROMETHEUS_FORMAT_RE, "_", key)
```
odeke-em: I think we want to use the precompiled regexp so basically please fix this to
```python
return _PROMETHEUS_FORMAT_RE.sub('_', key)
```

opencensus/stats/exporters/prometheus_exporter.py
odeke-em: And let's call it, `_NON_LETTERS_NOR_DIGITS_RE`

songy23: Neat - updated. Thanks for suggesting on Python code style!

PR #476, opened 2019-01-31
Metrics: add MetricsProducer and convert stats

### LINKED ISSUE #335: Add Metrics APIs as per "metrics.proto"

c24t:
This PR adds a `MetricProducer` class and adds the plumbing to allow `Stats` to produce metrics, building on #469. It also fixes a bug converting `DistributionAggregationData`s without bucket bounds into metrics `Points`.

Closes #335.

### DIFF
--- a/opencensus/metrics/export/metric_producer.py
+++ b/opencensus/metrics/export/metric_producer.py
@@ -1,0 +1,67 @@
+# Copyright 2019, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import threading
+
+
+class MetricProducer(object):
+    """Produces a set of metrics for export."""
+
+    def get_metrics(self):
+        """Get a set of metrics to be exported.
+
+        :rtype: set(:class: `opencensus.metrics.export.metric.Metric`)
+        :return: A set of metrics to be exported.
+        """
+        raise NotImplementedError  # pragma: NO COVER
+
+
+class MetricProducerManager(object):
+    """Container class for MetricProducers to be used by exporters.
+
+    :type metric_producers: iterable(class: 'MetricProducer')
+    :param metric_producers: Optional initial metric producers.
+    """
+
+    def __init__(self, metric_producers=None):
+        if metric_producers is None:
+            self.metric_producers = set()
+        else:
+            self.metric_producers = set(metric_producers)
+        self.mp_lock = threading.Lock()
+
+    def add(self, metric_producer):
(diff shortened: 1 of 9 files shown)


opencensus/stats/stats.py
```diff
@@ -12,24 +12,29 @@
...
     """Stats defines a View Manager and a Stats Recorder in order for the
     collection of Stats
     """
+
     def __init__(self):
-        self._stats_recorder = StatsRecorder()
-        self._view_manager = ViewManager()
-
-    @property
-    def stats_recorder(self):
```
c24t: I'd understand making an attribute protected and then exposing it as a property if we were returning a copy, but we don't seem to be doing that here.

c24t: https://github.com/census-instrumentation/opencensus-python/issues/477

opencensus/stats/stats.py
```diff
@@ -12,24 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime
+
+from opencensus.metrics.export.metric_producer import MetricProducer
```
c24t: I kept this consistent with other imports, but we should decide on class- or module-level imports across the library.

c24t: #478

opencensus/metrics/export/metric_producer.py
```diff
@@ -0,0 +1,67 @@
...
+        :type metric_producer: :class: 'MetricProducer'
+        :param metric_producer: The metric producer to remove.
+        """
+        if metric_producer is None:
+            raise ValueError
+        try:
+            with self.mp_lock:
+                self.metric_producers.remove(metric_producer)
+        except KeyError:
+            pass
```
mayurkale22: ```getAllMetricProducer``` is missing?

opencensus/metrics/export/metric_producer.py
```diff
@@ -0,0 +1,67 @@
...
+        with self.mp_lock:
+            self.metric_producers.add(metric_producer)
+
+    def remove(self, metric_producer):
+        """Remove a metric producer.
+
+        :type metric_producer: :class: 'MetricProducer'
+        :param metric_producer: The metric producer to remove.
+        """
+        if metric_producer is None:
```
mayurkale22: I am not sure If we want to do None check here...

opencensus/metrics/export/metric_producer.py
mayurkale22: I see, ok.

c24t: Added in fdd002f, if we go to the trouble of adding a lock to the class we might as well use it here.

Issue #481, opened 2019-02-05
Override trace sampling decision


c24t:
It should be possible to override the sampling decision for a given span by setting the span's sampler. Currently we always sample a span if the sampling bit is set in the span's `TraceOptions`. This bit indicates that the parent span was sampled.

For example, setting the sampler to `AlwaysOffSampler` should cause the span never to be sampled, even if the parent span was sampled. This isn't the case now because we [short-circuit the sampler](https://github.com/census-instrumentation/opencensus-python/blob/520a64b1bb3e296af917fe7a51e4779fe0a0396e/opencensus/trace/tracer.py#L76-L77) if the sampling bit is set.

Compare to the java client which sets the sampler in the [SpanBuilder](https://github.com/census-instrumentation/opencensus-java/blob/e8cba95228e9f104b325a99b3499d0da4be84e8d/api/src/main/java/io/opencensus/trace/SpanBuilder.java#L332-L335). In java we pass the span context to the sampler, which ultimately makes the sampling decision.

PR #482, opened 2019-02-05
Don't enable sampling by default

c24t:
This is a WIP PR to disable sampling using default trace options. This started as a drive-by fix to replace a string-valued bit field with an int and expanded as I compared `TraceOptions` and `SpanContext` to their counterparts in java.

@liyanhui1228 I'd appreciate some comments here since it's not obvious that sampling shouldn't be enabled by default, and there are a few design choices here I don't understand.

Addresses #481.

### DIFF
--- a/opencensus/trace/span_context.py
+++ b/opencensus/trace/span_context.py
@@ -18,19 +18,13 @@ import logging
 import re
 import uuid
 
-from opencensus.trace import trace_options
+from opencensus.trace import trace_options as trace_options_module
 
 _INVALID_TRACE_ID = '0' * 32
 INVALID_SPAN_ID = '0' * 16
 
 TRACE_ID_PATTERN = re.compile('[0-9a-f]{32}?')
 SPAN_ID_PATTERN = re.compile('[0-9a-f]{16}?')
-
-# Default options, enable tracing
-DEFAULT_OPTIONS = 1
-
-# Default trace options
-DEFAULT = trace_options.TraceOptions(DEFAULT_OPTIONS)
 
 
 class SpanContext(object):
@@ -64,7 +58,7 @@ class SpanContext(object):
             trace_id = generate_trace_id()
 
         if trace_options is None:
-            trace_options = DEFAULT
+            trace_options = trace_options_module.TraceOptions()
 
         self.from_header = from_header
         self.trace_id = self._check_trace_id(trace_id)
--- a/opencensus/trace/trace_options.py
+++ b/opencensus/trace/trace_options.py
@@ -18,40 +18,43 @@ import logging
 _ENABLED_BITMASK = 1 << 0
 
 # Default trace options
-DEFAULT = '0'
+DEFAULT = 0x0
 
 
 class TraceOptions(object):
     """A class that represents global trace options.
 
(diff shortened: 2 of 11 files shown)


opencensus/trace/span_context.py
```diff
@@ -64,7 +58,7 @@ def __init__(
             trace_id = generate_trace_id()
 
         if trace_options is None:
-            trace_options = DEFAULT
+            trace_options = trace_options_module.TraceOptions()
```
c24t: Re-using the default trace options here meant that tracers would set each others' contexts' sampling bit.

https://github.com/census-instrumentation/opencensus-python/blob/035a6dcbf45447468a18a2755a21fa0561384a2a/opencensus/trace/tracer.py#L84

I suspect there are other bugs like this lurking in the module.

c24t: TODO: add a test for tracer isolation.

opencensus/trace/trace_options.py
```diff
@@ -18,40 +18,43 @@
...
 
     def __repr__(self):
         fmt = '{}(enabled={})'
         return fmt.format(
             type(self).__name__,
-            self.get_enabled,
+            self.get_enabled(),
         )
 
+    # TODO: deprecate, replace with is_sampled()
```
c24t: TODO in #481.

opencensus/trace/trace_options.py
```diff
@@ -18,40 +18,43 @@
...
+    :type trace_options_byte: int
     :param trace_options_byte: 1 byte bitmap for trace options.
     """
 
-    def __init__(self, trace_options_byte=None):
-        if trace_options_byte is None:
+    def __init__(self, trace_options_byte=DEFAULT):
+        if not isinstance(trace_options_byte, int):
+            logging.warning("trace_options_byte must be an int, this will be "
+                            "an error in future versions")
```
c24t: This is technically an API change, so I figure the right approach is to log a warning here and make this an error in the next minor version release. We would benefit from some convention for deprecation, compare to java's `@deprecated` annotation.

tests/unit/trace/exporters/test_jaeger_exporter.py
```diff
@@ -308,7 +308,7 @@ def test_translate_to_jaeger(self):
                 operationName='test1',
                 startTime=1502820146071158,
                 duration=10000000,
-                flags=1,
+                flags=0,
```
c24t: TODO: check that this is the behavior we want.

opencensus/trace/tracer.py
```diff
@@ -73,7 +73,7 @@ def should_sample(self):
         :rtype: bool
         :returns: Whether to trace the request or not.
         """
-        return self.span_context.trace_options.enabled \
+        return self.span_context.trace_options.get_enabled() \
             or self.sampler.should_sample(self.span_context.trace_id)
```
tsloughter: Related to this PR, `should_sample` needs to take the trace options as well. For example, the way the probability sampler should work is that if trace options is true then return true, else check against the configured probability. This way if a remote parent is sampled the children will be too, but still allowing for a sampler to override that.

Issue #480, opened 2019-02-04
Exporters/Prometheus: Unable to record metrics 


PikBot:
On running the following [code](https://gist.github.com/PikBot/3c46d5d74a110de55595ba6f4110584a), I get the following error: 
```
Exception happened during processing of request from ('127.0.0.1', 50691)
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/socketserver.py", line 625, in process_request_thread
    self.finish_request(request, client_address)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/socketserver.py", line 354, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/server.py", line 422, in handle
    self.handle_one_request()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/server.py", line 410, in handle_one_request
    method()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/prometheus_client/exposition.py", line 146, in do_GET
    output = encoder(registry)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/prometheus_client/openmetrics/exposition.py", line 26, in generate_latest
    for k, v in sorted(s.labels.items())]))
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/prometheus_client/openmetrics/exposition.py", line 26, in <listcomp>
    for k, v in sorted(s.labels.items())]))
AttributeError: 'NoneType' object has no attribute 'replace'
```
I have tried running the example for Prometheus exporter as well (that doesn't seem to work for me either).
Not sure whether it is related to [this](https://github.com/prometheus/client_python/issues/18)? 
Any suggestions on as to what could be wrong, would be helpful! 
@odeke-em

PikBot:
Thank you @songy23 !

PR #486, opened 2019-02-05
Exporter/Prometheus: Convert None label values to empty string.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/480.

/cc @PikBot

### DIFF
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -169,7 +169,7 @@ class Collector(object):
         # Prometheus requires that all tag values be strings hence
         # the need to cast none to the empty string before exporting. See
         # https://github.com/census-instrumentation/opencensus-python/issues/480
-        tag_values = list(map(cast_none_to_empty_str, tag_values))
+        tag_values = [tv if tv else "" for tv in tag_values]
 
         if isinstance(agg_data, aggregation_data_module.CountAggregationData):
             metric = CounterMetricFamily(name=metric_name,
@@ -363,12 +363,3 @@ def sanitize(key):
     Replace all characters other than [A-Za-z0-9_] with '_'.
     """
     return _NON_LETTERS_NOR_DIGITS_RE.sub('_', key)
-
-
-def cast_none_to_empty_str(label_value):
-    """ convert None label to '' since Prometheus doesn't allow label values
-    to be None.
-    """
-    if label_value is None:
-        return ""
-    return label_value
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -166,11 +166,16 @@ class Collector(object):
         metric_description = desc['documentation']
         label_keys = desc['labels']
 
+        # Prometheus requires that all tag values be strings hence
+        # the need to cast none to the empty string before exporting. See
+        # https://github.com/census-instrumentation/opencensus-python/issues/480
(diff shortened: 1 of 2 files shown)


opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -165,12 +165,13 @@ def to_metric(self, desc, tag_values, agg_data):
         metric_name = desc['name']
         metric_description = desc['documentation']
         label_keys = desc['labels']
+        tag_values = list(map(cast_none_to_empty_str, tag_values))
```
odeke-em: Let's add a comment above this code to say
```python
# Prometheus requires that all tag values be strings hence
# the need to cast none to the empty string before exporting.
# See https://github.com/census-instrumentation/opencensus-python/issues/480
tag_values = list(map(cast_none_to_empty_str, tag_values)
```

songy23: SG, done. Also fixed a bug in histogram buckets.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -184,6 +184,10 @@ def to_metric(self, desc, tag_values, agg_data):
             for ii, bound in enumerate(agg_data.bounds):
                 cum_count += agg_data.counts_per_bucket[ii]
                 points[str(bound)] = cum_count
+            # Prometheus requires buckets to be sorted, and +Inf present.
+            # In OpenCensus we don't have +Inf in the bucket bonds so need to
```
songy23: See also https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java#L192-L193.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -183,10 +188,14 @@ def to_metric(self, desc, tag_values, agg_data):
...
+            # append it here.
+            points["+Inf"] = agg_data.count_data
```
odeke-em: What's this bug fix in here? A test demonstrating the problem would be useful.  I thought this code runs through the Prometheus Python client first before being exported? I think it would be good for us to file and diagnose that bug separately than sneak in a bug fix here without figuring out its extent.

songy23: See https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics_core.py#L204.

I've removed this commit from this PR for now.

opencensus/stats/exporters/prometheus_exporter.py
c24t: Nit, but I think it's a bit more readable and idiomatic to do:

```python
tag_values = [tv for tv in tag_values if tv else ""]
```

c24t: And why wrap the comment at <79 chars?

songy23: Updated, thanks for the suggestion.

> And why wrap the comment at <79 chars?

Test session "lint" checks line length and will fail on lines with >80 chars, including comments.

songy23:
Thanks for reviewing!

opencensus/stats/exporters/prometheus_exporter.py
c24t: I mean why not wrap at the last word boundary before 79, the line is wrapped early. Not a big deal in any case.

PR #492, opened 2019-02-06
Exporter/Prometheus: Upgrade version and fix buckets and labels.

songy23:
- Upgrade to latest Prometheus client.
- Use `UnknownMetricFamily` instead of `UntypedMetricFamily`, as the latter is now deprecated.
- Use ordered list for histogram buckets. Previously we use unordered map and that led to unpredictable behavior.
- Label keys and values must match when uploading metrics.
- Update unit tests to test against more precise results.

### DIFF
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -184,7 +184,7 @@ class Collector(object):
                         aggregation_data_module.DistributionAggregationData):
 
             assert(agg_data.bounds == sorted(agg_data.bounds))
-            # buckets are a list of bucket. Each bucket is another list with
+            # buckets are a list of buckets. Each bucket is another list with
             # a pair of bucket name and value, or a triple of bucket name,
             # value, and exemplar. buckets need to be in order.
             buckets = []
--- a/opencensus/stats/exporters/prometheus_exporter.py
+++ b/opencensus/stats/exporters/prometheus_exporter.py
@@ -18,7 +18,7 @@ from prometheus_client.core import CounterMetricFamily
 from prometheus_client.core import GaugeMetricFamily
 from prometheus_client.core import HistogramMetricFamily
 from prometheus_client.core import REGISTRY
-from prometheus_client.core import UntypedMetricFamily
+from prometheus_client.core import UnknownMetricFamily
 
 from opencensus.common.transports import sync
 from opencensus.stats import aggregation_data as aggregation_data_module
@@ -158,7 +158,7 @@ class Collector(object):
 
         :rtype: :class:`~prometheus_client.core.CounterMetricFamily` or
                 :class:`~prometheus_client.core.HistogramMetricFamily` or
-                :class:`~prometheus_client.core.UntypedMetricFamily` or
+                :class:`~prometheus_client.core.UnknownMetricFamily` or
                 :class:`~prometheus_client.core.GaugeMetricFamily`
         :returns: A Prometheus metric object
         """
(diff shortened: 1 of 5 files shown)


tests/system/stats/prometheus/prometheus_stats_test.py
```diff
@@ -74,11 +74,11 @@ def test_prometheus_stats(self):
             import urllib2
             contents = urllib2.urlopen("http://localhost:9303/metrics").read()
 
-        self.assertIn(b'# TYPE opencensus_request_count_view counter',
+        self.assertIn(b'# TYPE opencensus_request_count_view_total counter',
```
songy23: The latest Prometheus client automatically append "_total" suffix to count metrics.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -158,14 +158,15 @@ def to_metric(self, desc, tag_values, agg_data):
...
-                :class:`~prometheus_client.core.UntypedMetricFamily` or
+                :class:`~prometheus_client.core.UnknownMetricFamily` or
                 :class:`~prometheus_client.core.GaugeMetricFamily`
         :returns: A Prometheus metric object
         """
         metric_name = desc['name']
         metric_description = desc['documentation']
         label_keys = desc['labels']
 
+        assert(len(tag_values) == len(label_keys))
```
odeke-em: This is unfortunately going to panic in user code, am not sure if that's something we want?
Sure it demos a bug in the library but redeploying versions of an entire stack will be painful.
Perhaps if we suspect such problems, we can pad or truncate tag values depending on label_keys
Perhaps let's make a TODO for an action item later on and think about it.

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -183,26 +184,30 @@ def to_metric(self, desc, tag_values, agg_data):
                         aggregation_data_module.DistributionAggregationData):
 
             assert(agg_data.bounds == sorted(agg_data.bounds))
-            points = {}
-            cum_count = 0
+            # buckets are a list of bucket. Each bucket is another list with
```
odeke-em: s/list of bucket./list of buckets./g

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -183,26 +184,30 @@ def to_metric(self, desc, tag_values, agg_data):
...
+            # a pair of bucket name and value, or a triple of bucket name,
+            # value, and exemplar. buckets need to be in order.
+            buckets = []
+            cum_count = 0  # Prometheus buckets expect cumulative count.
             for ii, bound in enumerate(agg_data.bounds):
                 cum_count += agg_data.counts_per_bucket[ii]
-                points[str(bound)] = cum_count
+                bucket = [str(bound), cum_count]
```
odeke-em: Using lists is kind of overkill and potentially expensive.
Why not tuples that properly match the grouping we are using of pairs, or triples?
```go
cur = (str(bound), cum_count,)
buckets = append(cur)
```

opencensus/stats/exporters/prometheus_exporter.py
```diff
@@ -183,26 +184,30 @@ def to_metric(self, desc, tag_values, agg_data):
...
             metric.add_metric(labels=tag_values,
-                              buckets=list(points.items()),
+                              buckets=buckets,
                               sum_value=agg_data.sum,)
             return metric
 
         elif isinstance(agg_data,
                         aggregation_data_module.SumAggregationDataFloat):
-            metric = UntypedMetricFamily(name=metric_name,
+            metric = UnknownMetricFamily(name=metric_name,
```
odeke-em: What is the `SummaryMetricFamily` then?

opencensus/stats/exporters/prometheus_exporter.py
songy23: SG, done.

opencensus/stats/exporters/prometheus_exporter.py
songy23: Used list because Prometheus client comments said explicitly they expected a list of lists as buckets (https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics_core.py#L201):
> buckets: A list of lists. Each inner list can be a pair of bucket name and value, or a triple of bucket name, value, and exemplar. The buckets must be sorted, and +Inf present.

Guess tuples will work in the same way. Updated.

opencensus/stats/exporters/prometheus_exporter.py
songy23: > This is unfortunately going to panic in user code, am not sure if that's something we want?

No this won't. `to_metric()` should only be called within our library and users are not expected to call it directly. If for some reason the label keys and values don't match, that indicates an internal error in our library. 

In Java we have the exact same check: https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java#L143-L144.

opencensus/stats/exporters/prometheus_exporter.py
songy23: `SummaryMetricFamily` is for `Summary` metrics. `Sum` metrics should be converted to `UnknownMetricFamily`.

opencensus/stats/exporters/prometheus_exporter.py
odeke-em: @songy23 oh I see, I didn't know that, but perhaps please cite that in your commit. My apologies, please revert to the original that you had as I was assuming that you had just used list of list out of convenience.

songy23: No worries - changed it back to lists.

c24t:
> Use ordered list for histogram buckets. Previously we use unordered map and that led to unpredictable behavior.

Good catch! I'm surprised to see the bounds need to be sorted but it [looks to be the case](https://github.com/prometheus/client_python/blob/624bb61e6f15e0c3739fec853edaea2b91d1674f/prometheus_client/metrics_core.py#L204).

Changing the prometheus client version seems significant and user-facing enough to be worth adding a note to the changelog. What do you think?

songy23:
> Changing the prometheus client version seems significant and user-facing enough to be worth adding a note to the changelog. What do you think?

Make sense to me, added to changelog.

Issue #445, opened 2019-01-14
Refactor package distribution


reyang:
Currently OpenCensus Python is released as a monolithic package ```pip install opencensus``` which has the following disadvantages:
1. One will have to install all the dependencies even if most of them are not being used. This also adds concerns on deployment time/size and security.
2. Some of the exporters / integration with 3rd party libraries have to bind with the core OpenCensus Python release schedule. This is problematic for both the core OpenCensus Python and 3rd party integrations when it comes to the release / hotfix schedule.

The proposal is to refactor the current monolithic package using the namespace packaging approach https://packaging.python.org/guides/packaging-namespace-packages/.
Specifically, https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages will be used given we need to support both Python 2.x and 3.x.

reyang:
@c24t

c24t:
Following up from a conversation with @reyang:

We have two related problems here: how to split up the existing single distribution package (following the packaging docs, I'll call this a _distribution_), and how/whether to split the code in this repo into multiple repos. For now I'm assuming that we want to keep the existing package/namespace structure, and it looks like namespace packages let us do exactly that while moving the code between distributions and repos.

It's clear that we need a separate distribution for third party trace integrations (`opencensus.trace.ext`), but not clear whether we should move these into separate repos. For comparison, java client integrations are kept in the client repo, but in a [separate contrib package](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib). Keeping multiple distributions in the same repo is standard practice (see e.g. the [GCP python client library repo](https://github.com/googleapis/google-cloud-python), which includes a few dozen distributions), but it does make development more complicated.

We may also want to move the exporters (`opencensus.trace.exporters`, `opencensus.stats.exporters`)  into a separate distribution, or move each exporter into its own distribution. Unless we decide that we want a single distribution per repo, I think there's less reason to move these than third party integrations.

Then there are other components like the [google cloud format propagator](https://github.com/census-instrumentation/opencensus-python/blob/aa5935574754a337b5fdb1adbd45fd813c408db0/opencensus/trace/propagation/google_cloud_format.py) that are backend-specific, but can't be easily extracted from the library. I don't think we should touch this now, but worth keeping in mind if we want to prioritize having a small core library later.

c24t:
> One will have to install all the dependencies even if most of them are not being used

[...]

PR #493, opened 2019-02-06
refactor namespace package - cleanup opencensus root folder

reyang:

In order to allow other packages (e.g. extensions) sharing the same opencensus namespace, the opencensus root folder will have to follow https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages.

With this change, we've updated the `__init__.py` to meet with the namespace package requirement, and cleaned up the opencensus folder according to https://github.com/pypa/sample-namespace-packages/tree/master/pkgutil.


opencensus/stats/exporters/stackdriver_exporter.py
```diff
@@ -23,7 +23,7 @@
 from google.api_core.gapic_v1 import client_info
 from google.cloud import monitoring_v3
 
-from opencensus.__version__ import __version__
+from opencensus.common.__version__ import __version__
```
reyang: @c24t I wish to get your opinion as I was struggling with the ordering, should we put this below `from opencensus.common import utils` for alphabetic order?

reyang: Also the file name, do we want just version.py? (personally I would prefer version.py than __version__.py, and version.py has been used in the official sample https://packaging.python.org/guides/single-sourcing-package-version/)
I did some research and seems there is no consistent pattern https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc_version.py.

opencensus/__init__.py
```diff
@@ -1,20 +1 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-try:
-    import pkg_resources
-    pkg_resources.declare_namespace(__name__)
-except ImportError:  # pragma: NO COVER
-    import pkgutil
-    __path__ = pkgutil.extend_path(__path__, __name__)
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```
c24t: Looking at [`extend_path`](https://github.com/python/cpython/blob/2.7/Lib/pkgutil.py#L480-L552) I think you're supposed to include only this line to avoid import-time effects that change the path, but that comments are fine. But there may be some black magic I'm missing here.

reyang:
> Did you have to move the version because of some packaging requirement? In any case it LGTM.

Yep, for the approach we're taking (which is the only option that supports both Python 2 and 3), the only file allowed under the namespace folder is the `__init__.py`. If we have extra files under the namespace folder, they will be accidentally removed when any sub package (e.g. opencensus.ext.foobar) got uninstalled.

opencensus/stats/exporters/stackdriver_exporter.py
reyang: Got it, changing to version.py now.

opencensus/__init__.py
reyang: I thought about this and decided to remove the copyright comment because:
1. If 3rd party extension has the `__init__.py` without the copyright info (which we should expect), the file will get overwritten.
2. The official document https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages stated the following. I try to follow the doc rather instead of relying on the actual implementation. 

`The __init__.py file for the namespace package needs to contain only the following:`

```
__import__('pkg_resources').declare_namespace(__name__)
```

PR #494, opened 2019-02-07
Add gauges

c24t:
This PR adds `LongGauge` and `DoubleGauge` metrics classes following the [java](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/metrics/LongGaugeImpl.java) [implementation](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java), and incidentally adds/changes some related classes' repr methods to help inspect gauges.


opencensus/metrics/export/gauge.py
```diff
@@ -0,0 +1,202 @@
...
+    The constructor arguments are used to create a
+    :class:`opencensus.metrics.export.metric_descriptor.MetricDescriptor` for
+    converted metrics. See that class for details.
+    """
+
+    def __init__(self, name, description, unit, label_keys):
+        self._len_label_keys = len(label_keys)
+        self.descriptor = metric_descriptor.MetricDescriptor(
+            name, description, unit, self.descriptor_type, label_keys)
+        self.points = OrderedDict()
```
c24t: This is so the points appear in the converted metric's timeseries list in the order that they're added, may not be necessary.

opencensus/metrics/export/gauge.py
```diff
@@ -0,0 +1,222 @@
...
+        self._value_lock = threading.Lock()
+
+    def __repr__(self):
+        return ("{}({})"
+                .format(
+                    type(self).__name__,
+                    self.value
+                ))
+
+    def add(self, val):
```
colincadams: total fly by comment as I watch the metric progress: it strikes me as odd for a gauge to have an `add` method instead of just `set` and seems like it could lead to misuse by users, do you know the rationale behind this?

Is it because the value is threadsafe with the lock private, so a user couldn't do an atomic add without the functionality baked in?

c24t: I think the motivation here is to support data sources that only expose a delta.

I'm following the [java implementation](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/metrics/LongGauge.java#L140-L146) for now, but we need to add a description of the gauges API (including `add`) to [the spec](https://github.com/census-instrumentation/opencensus-specs).

> Is it because the value is threadsafe with the lock private, so a user couldn't do an atomic add without the functionality baked in?

It's actually the other way around: we wouldn't have to lock in `set` if we didn't want to provide an atomic `add`. That said, there's nothing (besides common sense) preventing the user from acquiring the lock themselves and changing the value.

opencensus/metrics/export/gauge.py
```diff
@@ -0,0 +1,222 @@
...
+
+    End users should use :class:`LongGauge` or :class:`DoubleGauge` instead of
+    using this class directly.
+
+    The constructor arguments are used to create a
+    :class:`opencensus.metrics.export.metric_descriptor.MetricDescriptor` for
+    converted metrics. See that class for details.
+    """
+
+    def __init__(self, name, description, unit, label_keys):
```
mayurkale22: I think we should validate all input params like ```name```, ```label_keys``` etc.

opencensus/metrics/export/gauge.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __repr__(self):
+        return ('{}(descriptor.name="{}", points={})'
+                .format(
+                    type(self).__name__,
+                    self.descriptor.name,
+                    self.points
+                ))
+
+    def get_time_series(self, label_values):
```
mayurkale22: I think is ```getDefaultTimeSeries ``` missing. https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/metrics/DoubleGauge.java#L104

opencensus/metrics/export/gauge.py
c24t: Just that they're not null or empty? We check `label_keys` in the `MetricDescriptor` constructor: https://github.com/census-instrumentation/opencensus-python/blob/88ad971ccf497a43f5c08e77f2817b14b931d003/opencensus/metrics/export/metric_descriptor.py#L133-L137

opencensus/metrics/export/gauge.py
c24t: That's true, I didn't understand the rationale for the default time series in the java client. The goal is that there's a single default point (read: time series) for each gauge that isn't associated with any label values?

c24t: Added in a2e56cf, let me know what you think.

PR #498, opened 2019-02-12
Add derived gauges

c24t:
This PR follows #494 and adds derived gauge classes `DerivedLongGauge`, `DerivedDoubleGauge`, and `DerivedGaugePoint` similar to the [corresponding classes in the java client](https://github.com/census-instrumentation/opencensus-java/tree/master/impl_core/src/main/java/io/opencensus/implcore/metrics). It also changes `TimeSeries` to allow null-valued points to deal with derived gauges tracking functions that have been garbage collected.

### DIFF
--- a/opencensus/common/utils.py
+++ b/opencensus/common/utils.py
@@ -115,7 +115,7 @@ def get_weakref(func):
     """Get a weak reference to bound or unbound `func`.
 
     If `func` is unbound (i.e. has no __self__ attr) get a weakref.ref,
-    otherwise get a wrapper that simulates simulates weakref.ref.
+    otherwise get a wrapper that simulates weakref.ref.
     """
     if func is None:
         raise ValueError
--- a/opencensus/common/utils.py
+++ b/opencensus/common/utils.py
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from functools import wraps
+try:
+    from weakref import WeakMethod
+except ImportError:
+    from opencensus.common.backports import WeakMethod
+
 import calendar
 import datetime
 import weakref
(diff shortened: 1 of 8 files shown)


c24t:
@reyang I'd be interested to hear your feedback on this PR too, in particular your thoughts on the `Gauge` type hierarchy -- whether we should use abstract classes (`GaugePoint`) as ersatz interfaces, whether it's worth using parent types (`BaseGauge`) to avoid duplicating code in the children, whether you think we'll be doomed to object-oriented hell for using multiple inheritance (`*GaugeMixin`), etc.

opencensus/common/utils.py
```diff
@@ -103,3 +109,16 @@ def window(ible, length):
             yield elts
         else:
             break
+
+
+def get_weakref(func):
+    """Get a weak reference to bound or unbound `func`.
+
+    If `func` is unbound (i.e. has no __self__ attr) get a weakref.ref,
+    otherwise get a wrapper that simulates simulates weakref.ref.
```
reyang: minor: remove duplicated `simulates`

c24t: Thanks, fixed.

opencensus/metrics/export/gauge.py
```diff
@@ -16,14 +16,48 @@
...
+    :return: A list of one `TimeSeries` for each point in `points`.
+    """
+    ts_list = []
+    for lv, gp in points.items():
+        point = point_module.Point(gp.to_point_value(), timestamp)
+        ts_list.append(time_series.TimeSeries(lv, [point], timestamp))
+    return ts_list
+
+
+class GaugePoint(object):
```
reyang: We could probably put the __repr__ implementation in the base class.

reyang: Probably extract the lock initialization in to `GaugePoint.__init__` as well?

```
class GaugePoint(object):
    def __init__(self):
        self._value_lock = threading.Lock()
```

c24t: I didn't do this because I wanted `DerivedGaugePoint` to be a `GaugePoint` too, and this would mean that we couldn't call its parent's constructor (or create an unused lock if we did).

I think there's good reason to have abstract classes like these where all the methods are unimplemented, but you can argue that classes should either be abstract or concrete -- and not like `Gauge` and `BaseGauge` that implement some methods and not others. What do you think?

reyang: I see, makes sense.

I don't have strong opinion here, both approaches have their valid reasons and I don't see a clear winner. Go with your choice :)

PR #500, opened 2019-02-12
Bump django to 1.11.20

c24t:
Another week, another [django CVE](https://www.djangoproject.com/weblog/2019/feb/11/security-releases/). See #443 for the last one.

We only have a test dependency on django now, but even still we don't want to list an insecure version number in a requirements file. Once #445 is done we can remove all references to django from this repo.

Update: changed to `1.11.20` to work around a botched django release.


c24t:
Tests are failing because there was a [problem with the django 1.11.19 release](https://code.djangoproject.com/ticket/30175), and it looks like we're still picking up the bad distribution in CI.

songy23:
Please rebase.

Issue #411, opened 2018-11-27
Remove min/max from Distribution


mayurkale22:
**Remove support for min/max in the stats Distribution to make it compatible with Metrics.**

Counterpart available here: census-instrumentation/opencensus-java#1542

PR #501, opened 2019-02-12
Stats: Remove min and max from Distribution.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/411.

`min` and `max` have been removed from `Distribution` in all other languages.


tests/unit/stats/test_aggregation_data.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2018, OpenCensus Authors
+# Copyright 2018, OpenCensus Authors
```
c24t: ?

songy23: Removed, seems to be a bad copy-and-paste.

PR #503, opened 2019-02-13
Add gauge registry

c24t:
This PR adds a registry class for gauges. The class exists so we can wrap up a set of `Gauge`s inside a metric producer. Calling `registry.get_metrics()` returns the set of metrics we get by calling `gauge.get_metrics(ts)` on each registered gauge for the current timestamp `ts`.

`Registry` roughly follows the [`MetricRegistry`](https://github.com/census-instrumentation/opencensus-java/blob/4125599ad34e074eb604428e09c8468020f84054/api/src/main/java/io/opencensus/metrics/MetricRegistry.java) implementation in the java client, but exposes a single `add_gauge` method instead of separate methods for each gauge type. 

The advantages of this change are: 
- it gives us a simpler API, less code
- avoids duplicating the arguments to `Gauge` constructors in `Registry` methods that create gauges
- makes gauge construction explicit

And the disadvantages are:
- it makes the library more verbose, users now need to construct gauges themselves
- users may forget to register gauges

This PR follows #498.


opencensus/metrics/registry.py
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime
+import threading
+
+from opencensus.metrics.export import metric_producer
+
+
+class Registry(metric_producer.MetricProducer):
```
reyang: Do we want to name this class as `opencensus.metrics.GaugeRegistry` or `opencensus.metrics.gauge.Registry`? `opencensus.metrics.Registry` is a very generic name.

opencensus/metrics/registry.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_metrics(self):
+        """Get a metric for each gauge in the registry at the current time.
+
+        :rtype: set(:class:`opencensus.metrics.export.metric.Metric`)
+        :return: A set of `Metric`s, one for each registered gauge.
+        """
+        now = datetime.now()
+        metrics = set()
+        with self._gauges_lock:
```
reyang: Depending on whether it is CPython or something else, we might not need the threading.Lock while getting `values()`.

opencensus/metrics/registry.py
```diff
@@ -0,0 +1,73 @@
...
+            self.gauges[name] = gauge
+
+    def get_metrics(self):
+        """Get a metric for each gauge in the registry at the current time.
+
+        :rtype: set(:class:`opencensus.metrics.export.metric.Metric`)
+        :return: A set of `Metric`s, one for each registered gauge.
+        """
+        now = datetime.now()
+        metrics = set()
```
reyang: Is there a preference over `set` than `tuple`?

opencensus/metrics/registry.py
c24t: That makes sense, `gauge.Registry` works for me.

opencensus/metrics/registry.py
c24t: No preference, just to emphasize that it's unordered. The java client [returns an `ArrayList`](https://github.com/c24t/opencensus-java/blob/c0f3a4cf4c437b9a4ac983f1bd44781f60d0182f/impl_core/src/main/java/io/opencensus/implcore/metrics/MetricRegistryImpl.java#L142-L149) but the order isn't fixed.

opencensus/metrics/registry.py
c24t: What about if another thread changes a value while we're iterating here? E.g.

```python
import time
from concurrent.futures import ThreadPoolExecutor

tpe1 = ThreadPoolExecutor(max_workers=1)
tpe2 = ThreadPoolExecutor(max_workers=1)

dd = dict(zip(range(10), range(10)))

def read_values():
    global dd
    for vv in dd.values():
        print(vv, end=", ")
        time.sleep(.01)
    print()
        
def mess_with_values():
    global dd
    for kk in dd:
        dd[kk] = -dd[kk]
        time.sleep(.01)


tpe1.submit(read_values)
tpe2.submit(mess_with_values)
```
```
0, 1, 2, 3, 4, -5, -6, -7, -8, -9, 
```

c24t: I think there are better ways to do this, all the locking in gauges so far has been pretty naive. One obvious improvement is to use a read/write lock to let more than one thread read at once. I didn't spend time optimizing this yet because I don't think we're likely to have a bottleneck here, but I did want to make sure the class was threadsafe.

reyang: I thought we are only adding new gauges (add_gauge) instead of altering the existing gauges?
And after we take a snapshot of the registered gauges, we get the metrics based on a pre-calculated value `now`.

c24t: I thought we'd want to remove gauges too, but after talking to @bogdandrutu it turns out that this is by design.

Looking more closely at `dictview` I think you're right that we don't need a lock here.

Issue #366, opened 2018-10-26
Improved Tracing for stand alone client applications


sduskis:
We're working on getting traces for stand-alone client applications in https://github.com/googleapis/google-cloud-python. To start with, we're trying to get traces for some of our system tests for clients based on grpc architecture: https://github.com/googleapis/google-cloud-python/pull/6298.  Our expectation was that we'd see multiple traces.  Instead, we saw multiple spans in a single trace.  @liyanhui1228 suggested that this was expected behavior.  That's different from what I've seen in Java.

We would like a mechanism to configure the `Tracer` to do the following:

- If there is a parent span configure in some "context", attach grpc spans to that parent span,.
- If there isn't a parent span, then the grpc span should result in an independent trace.

@c24t was in that meeting, and suggested that I add this issue.  Adding @dinooliva as an FYI.

sduskis:
FYI @tseaver, @crwilcox, @theacodes.

c24t:
Fixed in #539.

PR #505, opened 2019-02-15
Make root span trace IDs unique

### LINKED ISSUE #182: trace: clashing RPCs share the same traceID
I built code for a demo for "OpenCensus for Python gRPC developers" at https://github.com/orijtech/opencensus-for-grpc-python-developers and blogpost https://medium.com/@orijtech/opencensus-for-python-grpc-developers-9e460e054395

Given this code below
### Server
```python
import grpc
import os
import time
from concurrent import futures

import defs_pb2_grpc as proto
import defs_pb2 as pb

# Tracing related imports
from opencensus.trace.exporters import stackdriver_exporter
from opencensus.trace.samplers import always_on
from opencensus.trace.ext.grpc import server_interceptor

# Create the exporters:
# 1. Stackdriver
stackdriverExporter = stackdriver_exporter.StackdriverExporter(
            project_id=os.environ.get('OCGRPC_PROJECTID', 'census-demos'),
            transport=BackgroundThreadTransport)

class CapitalizeServer(proto.FetchServicer):
    def __init__(self, *args, **kwargs):
        super(CapitalizeServer, self).__init__()

    def Capitalize(self, request, context):
        tracer = Tracer(sampler=always_on.AlwaysOnSampler(), exporter=stackdriverExporter)
        with tracer.span(name='Capitalize') as span:
            data = request.data
            span.add_annotation('Data in', len=len(data))
            return pb.Payload(data=data.upper())

def main():
    # Setup and start the gRPC server with the OpenCensus integration/interceptor
    tracer_interceptor = server_interceptor.OpenCensusServerInterceptor(
            always_on.AlwaysOnSampler(), stackdriverExporter)

    server = grpc.server(
            futures.ThreadPoolExecutor(max_workers=10),
            interceptors=(tracer_interceptor,))
    proto.add_FetchServicer_to_server(CapitalizeServer(), server)
    server.add_insecure_port('[::]:9778')
    server.start()

    try:
        while True:
            time.sleep(60 * 60)
    except KeyboardInterrupt:
        server.stop(0)

if __name__ == '__main__':
    main()
```

## Client
```shell
import grpc

import defs_pb2_grpc as proto
import defs_pb2 as pb

def main():
    channel = grpc.insecure_channel('localhost:9778')
    stub = proto.FetchStub(channel)

    while True:
        lineIn = input('> ')
        capitalized = stub.Capitalize(pb.Payload(data=bytes(lineIn, encoding='utf-8')))
        print('< %s\n'%(capitalized.data.decode('utf-8')))

if __name__ == '__main__':
    main()
```

However, while running multiple clients to simulate traffic and popularity of a service, I noticed that a bunch of traces were bunched up together as viewed in the Stackdriver UI
screen shot 2018-05-24 at 3 23 56 pm

Bouncing around ideas with @sebright and @songy23, @sebright gave a postulation of perhaps the same traceID being shared which alas seems to the be problem, as per our log examinations

screen shot 2018-05-24 at 3 28 16 pm
screen shot 2018-05-24 at 3 27 55 pm

This has a problem of bunching up totally different traces from different clients that are concurrently occuring and seems perhaps like a bug?

c24t:
This PR fixes a longstanding bug: that all spans generated by a given `ContextTracer` share the same trace ID regardless of the context. This meant sequential root spans were reported as siblings in the same trace instead of each belonging to its own trace.

Fixes #182 and #366. See https://github.com/census-instrumentation/opencensus-python/issues/182#issuecomment-461647639 for more detail.

cc @tab1293, @sduskis.

### DIFF
--- a/opencensus/trace/tracers/context_tracer.py
+++ b/opencensus/trace/tracers/context_tracer.py
@@ -16,10 +16,11 @@ import logging
 import threading
 
 from opencensus.trace import execution_context
-from opencensus.trace.span_context import SpanContext
 from opencensus.trace import span as trace_span
 from opencensus.trace import span_data as span_data_module
 from opencensus.trace.exporters import print_exporter
+from opencensus.trace.span_context import SpanContext
+from opencensus.trace.span_context import generate_trace_id
 from opencensus.trace.tracers import base
 
 
@@ -116,6 +117,8 @@ class ContextTracer(base.Tracer):
             execution_context.set_current_span(cur_span.parent_span)
         else:
             execution_context.set_current_span(None)
+            self.span_context.trace_id = generate_trace_id()
+            self.trace_id = self.span_context.trace_id
 
         with self._spans_list_condition:
             if cur_span in self._spans_list:
--- a/tests/unit/trace/tracers/test_context_tracer.py
+++ b/tests/unit/trace/tracers/test_context_tracer.py
@@ -136,13 +136,20 @@ class TestContextTracer(unittest.TestCase):
         mock_span.attributes = {}
         mock_span.__iter__ = mock.Mock(return_value=iter([mock_span]))
         parent_span_id = '6e0c63257de34c92'
+        mock_span.parent_span = mock.Mock(spec=span.Span)
         mock_span.parent_span.span_id = parent_span_id
         mock_current_span.return_value = mock_span
+        old_trace_id = tracer.trace_id
+        self.assertEqual(tracer.trace_id, tracer.span_context.trace_id)
         tracer.end_span()
 
         self.assertTrue(mock_span.finish.called)
         self.assertEqual(tracer.span_context.span_id, parent_span_id)
         self.assertFalse(tracer.exporter.export.called)
+
+        # Check that we don't change the trace ID if the parent span exists
+        self.assertEqual(tracer.trace_id, old_trace_id)
+        self.assertEqual(tracer.trace_id, tracer.span_context.trace_id)
 
     @mock.patch.object(context_tracer.ContextTracer, 'current_span')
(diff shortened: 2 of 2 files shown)


opencensus/trace/tracers/context_tracer.py
```diff
@@ -116,6 +117,8 @@ def end_span(self, *args, **kwargs):
             execution_context.set_current_span(cur_span.parent_span)
         else:
             execution_context.set_current_span(None)
+            self.span_context.trace_id = generate_trace_id()
+            self.trace_id = self.span_context.trace_id
```
c24t: `tracer.trace_id` is unused, and I don't know why it's included on the tracer. Removing it is a problem for another PR.

sduskis:
@c24t, what's the story with this PR?

nikhaldi:
I'll mention here too that this PR does not fix #182 for me while my proposed change #539 does.

c24t:
> what's the story with this PR?

I hadn't merged it because it was breaking a system test that was asserting the wrong behavior. Still TODO.

> this PR does not fix #182 for me while my proposed change #539 does

This PR deals with siblings spans having the same trace ID when the parent trace ID is null, #539 changes the grpc client interceptor so that it uses the current tracer to set trace IDs instead of the tracer that was current at the time the interceptor was created. Both problems affect the `google-cloud-python` integration, and both cause clashing traceIDs, but each solves a different root problem.

@nikhaldi does that match your understanding?

nikhaldi:
I see what you mean, thanks. I don't have enough background about how people use the API to judge if the case of a null parent trace id is valid. My reading of the code tracking down my issues was that the API requires a new tracer to be created for every individual trace. But if that's not case this PR seems reasonable.

Issue #509, opened 2019-02-20
Stats: Timestamps not updated when new measurement is recorded


colincadams:
A call to `MeasurementMap.record` creates a timestamp and then passes that through `MeasureToViewMap.record` -> `ViewData.record` -> `AggregationData.add_sample` but is never persisted anywhere.

This means that the timestamp of when the view is registered is used whenever measurements are exported. For stackdriver this means that the first time the time series are sent from a particular instance it succeeds, but any subsequent call fails because the timestamp is too close to the prior timestamp or if the instance has been up for more than a day it fails as the timestamp is too old.

I would expect the `_end_time` of `ViewData` to be updated when a new measurement is recorded, which would fix this problem for cumulative metrics. Depending on how `_start_time` and `_end_time` are used when exporting, this may be the correct solution for gauges as well.

songy23:
`end_time` is set when `ViewData` gets exported: https://github.com/census-instrumentation/opencensus-python/blob/f669aa9553b120e2b454fed6581ad13db8818ccc/opencensus/stats/measure_to_view_map.py#L72. So `end_time` will always be newer than `start_time`.

We don't use the timestamp associated with measurements except in `Exemplar`s, since `ViewData` represents aggregated data. During aggregation each individual measurement and timestamp are dropped.

colincadams:
`get_view` doesn't seem to be called on export. Maybe it should be? Instead in `record` the items in the underlying map are just iterated over, a list is created, and that is passed to `export`.

https://github.com/census-instrumentation/opencensus-python/blob/f669aa9553b120e2b454fed6581ad13db8818ccc/opencensus/stats/measure_to_view_map.py#L129

I don't completely understand why it being aggregated data means the timestamp shouldn't be used. To me the aggregated data represents the data for the interval up to and including the `timestamp` passed in, so the start and end should reflect that. Having `end_time` set there instead of mutating it in the getter seems less error prone.

Otherwise maybe `end_time` should always be `None` for the ones in the map, and only set on export? Then maybe the errors would be more obvious? That begs the question as to why `ViewData` even has start and end times however. If they are simply going to be set on export, and a `ViewData` itself doesn't have a fixed interval then I would argue it shouldn't have timestamps and the timestamp should just be an additional argument to `export` on the `MeasureToViewMap`.

What do you think?

songy23:
[...]

PR #512, opened 2019-02-21
Stats: Make defensive copy and set end time for ViewData when export.

songy23:
Updates https://github.com/census-instrumentation/opencensus-python/issues/509.

### DIFF
--- a/opencensus/stats/measure_to_view_map.py
+++ b/opencensus/stats/measure_to_view_map.py
@@ -66,7 +66,7 @@ class MeasureToViewMap(object):
         else:
             return None
 
-        return self.copy_view_data(view_data)
+        return self.copy_and_finalize_view_data(view_data)
 
     def filter_exported_views(self, all_views):
         """returns the subset of the given view that should be exported"""
@@ -120,7 +120,8 @@ class MeasureToViewMap(object):
 
     def export(self, view_datas):
         """export view datas to registered exporters"""
-        view_datas_copy = [self.copy_view_data(vd) for vd in view_datas]
+        view_datas_copy = \
+            [self.copy_and_finalize_view_data(vd) for vd in view_datas]
         if len(self.exporters) > 0:
             for e in self.exporters:
                 e.export(view_datas_copy)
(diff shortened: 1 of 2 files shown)


opencensus/stats/measure_to_view_map.py
```diff
@@ -145,3 +142,11 @@ def get_metrics(self, timestamp):
                 metric = metric_utils.view_data_to_metric(vd, timestamp)
                 if metric is not None:
                     yield metric
+
+    # TODO(issue #470): remove this method once we export immutable stats.
+    def copy_view_data(self, view_data):
```
colincadams: nit: Not sure how others feel about naming, but maybe something like `finalize_view_data` to make it clear that this also modifies it? Not super important given that this should go away soon.

tests/unit/stats/test_measure_to_view_map.py
```diff
@@ -378,3 +378,20 @@ def test_export(self):
         measure_to_view_map._registered_measures = {}
         measure_to_view_map.export(view_data)
         self.assertTrue(True)
+
+    def test_export_duplicates_viewdata(self):
```
c24t: This test might be too close to just asserting the implementation to be useful, you may want to change it to e.g. check that the end timestamp is getting set correctly. Since this PR is a bugfix I think it ought to include a test to prevent a regression, up to you whether you think this does that.

opencensus/stats/measure_to_view_map.py
songy23: Actually this method does both - it first makes a copy of the original `view_data` and then sets the `end_time` of that copy. I renamed it to `copy_and_finalize_view_data` to make this clear.

tests/unit/stats/test_measure_to_view_map.py
songy23: Thanks for adding the test, I've fixed a failure with mock and made it test if `end_time` is properly updated.

songy23:
Thanks for the review!

Issue #457, opened 2019-01-17
Implement Resources


c24t:
Implement the [Resources spec](https://github.com/census-instrumentation/opencensus-specs/blob/ed99dcbaf689afbf3195908bb107fc835d6bb0e3/resource/Resource.md), including:
- setting resources for custom environments
- encoding and decoding resources
- merging resources

See also the existing [`MonitoredResource` classes](https://github.com/census-instrumentation/opencensus-python/blob/989470edce8fa7a82813e3846f774fd4927b5b07/opencensus/common/monitored_resource_util/monitored_resource.py).

c24t:
See the PR that added resources to the java client: https://github.com/census-instrumentation/opencensus-java/pull/1555/.

PR #513, opened 2019-02-21
Add support for custom resources

c24t:
This PR adds support for `Resource`s based on environment variables. Users can set `OC_RESOURCE_TYPE` and `OC_RESOURCE_LABELS` and get a custom resource.

Note that we're not automatically creating the resource or propagating it yet.

Addresses #457.

### DIFF
--- a/opencensus/common/resource.py
+++ b/opencensus/common/resource.py
@@ -14,12 +14,35 @@
 
 
 from copy import copy
+import os
 import re
 
+
+OC_RESOURCE_TYPE = 'OC_RESOURCE_TYPE'
+OC_RESOURCE_LABELS = 'OC_RESOURCE_LABELS'
 
 # Matches anything outside ASCII 32-126 inclusive
 NON_PRINTABLE_ASCII = re.compile(
     r'[^ !"#$%&\'()*+,\-./:;<=>?@\[\\\]^_`{|}~0-9a-zA-Z]')
+
+# Label key/value tokens, may be quoted
+WORD_RES = r'(\'[^\']*\'|"[^"]*"|[^\s,=]+)'
+
+KV_RE = re.compile(r"""
+    \s*                 # ignore leading spaces
+    (?P<key>{word_re})  # capture the key word
+    \s*=\s*
+    (?P<val>{word_re})  # capture the value word
+    \s*                 # ignore trailing spaces
+    """.format(word_re=WORD_RES), re.VERBOSE)
+
+LABELS_RE = re.compile(r"""
+    ^\s*{word_re}\s*=\s*{word_re}\s*     # KV_RE without the named groups
+    (,\s*{word_re}\s*=\s*{word_re}\s*)*  # more KV pairs, comma delimited
+    $
+    """.format(word_re=WORD_RES), re.VERBOSE)
+
+UNQUOTE_RE = re.compile(r'^([\'"]?)([^\1]*)(\1)$')
 
 
 def merge_resources(r1, r2):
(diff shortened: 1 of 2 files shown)


opencensus/common/resource.py
```diff
@@ -14,13 +14,36 @@
...
+OC_RESOURCE_LABELS = 'OC_RESOURCE_LABELS'
+
 # Matches anything outside ASCII 32-126 inclusive
 NON_PRINTABLE_ASCII = re.compile(
     r'[^ !"#$%&\'()*+,\-./:;<=>?@\[\\\]^_`{|}~0-9a-zA-Z]')
 
+# Label key/value tokens, may be quoted
+WORD_RES = r'(\'[^\']*\'|"[^"]*"|[^\s,=]+)'
+
+KV_RE = re.compile(r"""
```
reyang: Do we intend this to be _public_ (consumed directly by the users)? If not, we can add `_` prefix to the name.

opencensus/common/resource.py
```diff
@@ -124,3 +147,57 @@ def merge(self, other):
...
+    {'k1': 'v1', 'k2': 'v2'}
+    >>> parse_labels("k1='v1,=z1'")
+    {'k1': 'v1,=z1'}
+    """
+    if not LABELS_RE.match(labels_str):
+        return None
+    labels = {}
+    for kv in KV_RE.finditer(labels_str):
+        gd = kv.groupdict()
+        labels[unquote(gd['key'])] = unquote(gd['val'])
```
reyang: Do we want to check for duplicated keys and throw error here?

opencensus/common/resource.py
c24t: I've erred on the side of leaving things public since I think it's a losing battle to add a leading underscore to everything that's not in the API, but I think you're right that it's worth hiding module level constants like this since it's less obvious which ones are supposed to be part of the API.

opencensus/common/resource.py
c24t: Sure, added in 694ed5a.

opencensus/common/resource.py
reyang: I actually don't have strong opinions since I heard about the Python philosophy `We are all consenting adults here.`.

Looking at the current code, we do follow the underscore convention for constants though (at least in the majority of files).

https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/trace_options.py#L18
https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/tracestate.py#L18

opencensus/common/resource.py
songy23: In other languages if there're duplicated key-value pairs, we simply overwrite with the last value. May be better to be consistent here.

c24t: Changed to log a warning instead in a265e0e, which I think is a good compromise.

tests/unit/common/test_resource.py
```diff
@@ -142,3 +142,79 @@ def test_check_ascii_256(self):
...
+
+        self.assertEqual(
+            resource_module.parse_labels('k1="one/two/three"'),
+            {'k1': 'one/two/three'})
+
+        self.assertEqual(
+            resource_module.parse_labels('k1="one\\two\\three"'),
+            {'k1': 'one\\two\\three'})
+
+        with self.assertRaises(ValueError):
```
songy23: Please fix this.

c24t: Sorry @songy23! That's what I get for not rerunning the tests...

Issue #514, opened 2019-02-21
Set k8s resource labels for GCE, AWS


c24t:
Resource labels for containers on kubernetes (previously "GKE") aren't mutually exclusive with labels for GCE and AWS environments. Resources created for these containers should also include GCE/AWS labels if we detect the environment.

See [`ResourceUtils` in the java client](https://github.com/census-instrumentation/opencensus-java/blob/4125599ad34e074eb604428e09c8468020f84054/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/ResourceUtils.java#L45-L54).

c24t:
Fixed in #515.

Issue #510, opened 2019-02-20
Replace gke_container resource with k8s_container


c24t:
From https://github.com/census-instrumentation/opencensus-python/pull/508#pullrequestreview-205548636:

> Note that `gke_container` resource is now replaced with `k8s_containter` resource, and it needs to be updated in the following PRs.

PR #515, opened 2019-02-22
Merge custom resources on detection

### LINKED ISSUE #457: Implement Resources

c24t:
This PR changes resource detection (`monitored_resource.get_instance`) to include custom type and resource labels if available, and to include labels from multiple resources if we detect more than one known environment (e.g. kubernetes on AWS).

Fixes #457 and #514.

cc @mayurkale22


opencensus/common/resource.py
```diff
@@ -49,22 +49,32 @@
...
-    labels.update(r1.labels)
-    return Resource(type_, labels)
+    if not resource_list:
+        raise ValueError
+    rtype = None
+    for rr in resource_list:
+        if rr.type:
+            rtype = rr.type
+            break
+    last, rest = resource_list[::-1][0], resource_list[::-1][1:]
```
reyang: Consider this:
```
labels = {}
for rr in reversed(resource_list):
    labels.update(rr.labels)
```

opencensus/common/resource.py
```diff
@@ -49,22 +49,32 @@
 _UNQUOTE_RE = re.compile(r'^([\'"]?)([^\1]*)(\1)$')
 
 
-def merge_resources(r1, r2):
-    """Merge two resources to get a new resource.
+def merge_resources(resource_list):
```
reyang: @c24t I'd like to hear about your opinion.

Shall we consider the signature of `os.path.join`? https://docs.python.org/2/library/os.path.html#os.path.join

opencensus/common/resource.py
c24t: Yeah, that's much nicer.

opencensus/common/resource.py
c24t: Part of the motivation for making the arg a list was to match the [signature in java](https://github.com/c24t/opencensus-java/blob/c0f3a4cf4c437b9a4ac983f1bd44781f60d0182f/api/src/main/java/io/opencensus/resource/Resource.java#L124), but since `*args` in python is more idiomatic than varargs in java, maybe this is the better approach here, and I like this pattern as a user.

You think callers are more likely to have to make a list to call `merge_resources` than have to splat a list if we mimic `os.path.join`?

c24t: I put up https://github.com/c24t/opencensus-python/pull/1/files#diff-21856523a57af46b9f02f815944713e3R89 to see what this would look like.

c24t:
@songy23 good call, let's wait on this until after #510.

opencensus/common/resource.py
reyang: I don't have strong opinion here, just putting another option and see which one you would prefer.
Please do what you think is better. Either way looks good to me :)

PR #518, opened 2019-02-23
Tracing: Add Double/Float64 Attribute.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/517.

PR #526, opened 2019-02-26
Exporter/OcAgent: Support double attribute.

songy23:
As a follow-up of https://github.com/census-instrumentation/opencensus-python/pull/518#pullrequestreview-207608423.


c24t:
Thanks @songy23!

songy23:
Thanks :)

PR #528, opened 2019-02-27
Replace gke_container with k8s_container for stackdriver

### LINKED ISSUE #510: Replace gke_container resource with k8s_container

c24t:
This PR replaces the `gke_container` monitored resource type with the new `k8s_container` for stackdriver following [the migration guide](https://cloud.google.com/monitoring/kubernetes-engine/migration#incompatible).

It also moves the detection utils for kubernetes out of `gcp_metadata_config` and into a new module. It still depends on GCP metadata, so it won't work on non-GKE kubernetes environments, but this does match the behavior of the other clients.

See https://github.com/census-instrumentation/opencensus-java/pull/1467 for the same change in the java client.

This PR fixes #510 and unblocks #515.

@songy23 @mayurkale22 there's a good chance I'm still missing something here, please take a look.


tests/unit/common/monitored_resource_util/test_k8s_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    from unittest import mock
+
+import os
+import unittest
+
+from opencensus.common.monitored_resource import monitored_resource
+from opencensus.common.monitored_resource import k8s_utils
+
+
+class TestK8SUtils(unittest.TestCase):
```
c24t: These were moved from `test_gcp_metadata_config.py`. Note that `instance_id` label doesn't appear anymore.

opencensus/common/monitored_resource/k8s_utils.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+
+import os
+
+from opencensus.common.monitored_resource import gcp_metadata_config
+
+_K8S_ATTRIBUTES = {
+    # ProjectID is the identifier of the GCP project associated with this
+    # resource, such as "my-project".
+    'project_id': 'project/project-id',
```
songy23: `project-id` and `location` aren't part of the K8S resource. For K8S container resource we have:
- k8s.io/cluster/name
- k8s.io/namespace/name
- k8s.io/pod/name
- k8s.io/container/name

See the latest Java implementation: https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/K8sContainerResource.java.

I'll update the specs in a minute.

opencensus/common/monitored_resource/monitored_resource.py
```diff
@@ -54,8 +42,8 @@ def get_instance():
 
     Supported environments include:
 
-    1. 'gke_container'
-    - https://cloud.google.com/monitoring/api/resources#tag_gke_container
+    1. 'k8s_container'
+    - https://cloud.google.com/monitoring/api/resources#tag_k8s_container
```
songy23: Those URLs (including the ones below) shouldn't point to Stackdriver documentation, since this is our core library (not exporters). Please update this to point to the [specs](https://github.com/census-instrumentation/opencensus-specs/tree/master/resource).

opencensus/common/monitored_resource/k8s_utils.py
c24t: Thanks, updated in baa06f1.

opencensus/common/monitored_resource/monitored_resource.py
c24t: I dropped the whole comment since we link to the resource spec from the `Resource` docstring.

opencensus/common/monitored_resource/k8s_utils.py
```diff
@@ -0,0 +1,64 @@
...
+CLUSTER_NAME_KEY = 'k8s.io/cluster/name'
+
+# ID of the instance the container is running on
+NAMESPACE_NAME_KEY = 'k8s.io/namespace/name'
+
+# Container pod ID
+POD_NAME_KEY = 'k8s.io/pod/name'
+
+# Container name
+CONTAINER_NAME_KEY = 'k8s.io/container'
```
songy23: This should be 'k8s.io/container/name': https://github.com/census-instrumentation/opencensus-specs/pull/235/files#diff-b965ed607d4cefb5c41b99223e8e0ac0R16.

PR #529, opened 2019-02-27
refactor namespace package - cleanup opencensus.common namespace

reyang:

In order for namespace foo.bar.baz to work, the folder foo, bar and baz can only contain `__init__.py` file and sub folders, no other files are allowed.

In this PR I moved existing files under opencensus.common to sub folders, without changing the import path.

With this change, we can install/remove/reinstall/update opencenus and [opencensus-common-correlationcontext](https://github.com/census-instrumentation/opencensus-python/tree/master/contrib/opencensus-common-correlationcontext) in arbitrary order.


opencensus/common/backports/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from opencensus.common.backports.backports import WeakMethod
+
+
+__all__ = [
+    'WeakMethod',
+]
```
c24t: I'm worried about these alias-style init modules getting out of sync, and it'll be tedious updating these with each change to the code. 

What about moving the code from e.g. the old `backports.py` into `backports/__init__.py` instead? This is still pretty ugly -- lots of code in `__init__.py` files -- but it fixes the stutter and solves the problem of keeping code in sync.

reyang: Either way it is ugly as you said.
I'm fine with moving code to `__init__.py` directly, will do it now.

PR #531, opened 2019-02-28
refactor namespace package - add test integration and release for contrib package

reyang:
1. Added opencensus-common-correlationcontext as a dependency of opencensus package.
2. Include opencensus-common-correlationcontext in the unit test / CI / release process.
3. Minor cleanup/formatting (make a consistent use of single-quote in the setup.py).


scripts/twine_upload.sh
```diff
@@ -29,4 +29,6 @@ python3 -m pip install --upgrade twine wheel setuptools
 
 # Build the distribution and upload.
 python3 setup.py bdist_wheel
+cd contrib/opencensus-common-correlationcontext
+python3 setup.py bdist_wheel --dist-dir ./../../dist/
```
reyang: Note: this will **not work** since I need to add pushd/popd. I will need a PyPI account and test this on Linux.

I plan to implement a loop, discovering all the sub-folders under contrib and publish them one by one.

.coveragerc
```diff
@@ -3,6 +3,8 @@ branch = True
 omit = 
        *gen/jaeger*
        *gen/opencensus*
+       opencensus/__init__.py
```
reyang: This is a workaround: the `__init__.py` file got replaced by another namespace package, which confused the coverage test.

setup.py
```diff
@@ -16,17 +16,19 @@
...
-    "stackdriver": ['google-cloud-trace>=0.20.1, <0.30'],
-    "prometheus_client": ['prometheus_client==0.3.1'],
-    "requests": ['wrapt==1.10.11']
+    'prometheus_client': ['prometheus_client==0.3.1'],
+    'requests': ['wrapt==1.10.11'],
+    'stackdriver': ['google-cloud-trace>=0.20.1, <0.30'],
 }
 
 install_requires = [
+    'opencensus-common-correlationcontext == {}'.format(__version__),
```
c24t: I'm surprised you want to keep the versions in sync.

So installing `opencensus` will install `opencensus-common-correlationcontext`, but not vice-versa. Other packages like exporters will install `opencensus` on install, but not vice-versa. Is that right?

I think you could shorten the distribution name to e.g. `opencensus-correlation`, or is there a reason you need to match the package name? In any case that's a matter of style, up to you.

reyang: I actually plan to extract the version to a common place and share it across all packages in this repo (with the cost of generating the version.py file on the fly during build/packaging/publishing). Packages in different repo (e.g. opencensus-ext-microsoft-azure-exporter) can have their own version. Do you think this is a good or bad idea?

Yes, `opencensus` depends on `opencensus-common-correlationcontext`. Exporters will depend on `opencensus`. Currently I'm doing dirty job in nox.py in order to get local installation/test working. I haven't yet found a solution to tell pip to look for a specified local path before going to the PyPI indexer.

The pros: all packages will have the same version, easy to maintain and troubleshoot.
The cons: even if we don't update the code for certain package, the version will still increase on each release.

reyang: As you have observed, I'm trying to match the package name with the namespace - e.g. namespace foo.bar.baz will have package name foo-bar-baz. I thought this would help other plugin/extension developers to keep names consistent.

Do you have a better idea? I've always had problem naming things.

c24t: I think one of the big benefits of having a different distribution for each package is that they can have different version numbers, especially because of the semantics for breaking changes. If e.g. we release a major- or minor-version change to `foo` and a patch-version update to `bar` users should still be able to safely update `bar` without changing application code. They'd have to go look at the changelog to tell this if we keep the version numbers in sync.

I think it's fine to have different names for the distribution and package, and personally prefer short names.

reyang: Per our discussion:
1. Changed to short package name `opencensus-correlation`.
2. Each package will have its independent version and changelog. (This would benefit us in the long run if we have git submodule).

setup.py
```diff
@@ -16,17 +16,19 @@
...
-    "stackdriver": ['google-cloud-trace>=0.20.1, <0.30'],
-    "prometheus_client": ['prometheus_client==0.3.1'],
-    "requests": ['wrapt==1.10.11']
+    'prometheus_client': ['prometheus_client==0.3.1'],
+    'requests': ['wrapt==1.10.11'],
+    'stackdriver': ['google-cloud-trace>=0.20.1, <0.30'],
 }
 
 install_requires = [
+    'opencensus-correlation == 0.2.dev0',
```
colincadams: when we try to install opencensus at this commit or newer, we get an error:
```
Obtaining opencensus from git+https://github.com/census-instrumentation/opencensus-python.git@d37b6d267307a136631881e593c2bc8921a786b6#egg=opencensus
  Skipping because already up-to-date.
Collecting opencensus-correlation==0.2.dev0 (from opencensus)

  Could not find a version that satisfies the requirement opencensus-correlation==0.2.dev0 (from opencensus) (from versions: )
No matching distribution found for opencensus-correlation==0.2.dev0 (from opencensus)
```

Has any version of it been released yet? Should it be required or an extra? I'm not totally clear on the new packaging story, and know that our use case of installing for git isn't necessary supported, but want to know if we can make it work until there's a new release with #534 included :grinning:

reyang: @colincadams Sorry for the inconvenience! It hasn't been released yet. OpenCensus will have dependency on `opencensus-correlation` and `opencensus-tracecontext` (not there yet), the other contrib packages (e.g. extensions, exporters) will have dependency on OpenCensus.

The workaround is to install the package from a cloned git repo, please refer to
https://github.com/census-instrumentation/opencensus-python/blob/master/nox.py#L21

Issue #533, opened 2019-03-04
exporter/stackdriver: reports zone as fully qualified resource


colincadams:
When running an app engine flex app, open census defaults reporting to the `gce_instance` monitored resource (seems fine) with `zone` set to the fully qualified zone returned from [the metadata api](https://cloud.google.com/compute/docs/storing-retrieving-metadata#default):

```
OpenCensus monitored resource: "{'type': 'gce_instance', 'labels': {'instance_id': '[instance-id]', 'project_id': '[project-id]', 'zone': 'projects/[numeric-project-id]/zones/us-east1-c'}}"
```

This causes all points to fail like this:

```
async_/_thread_main : StackdriverStatsExporter failed to emit data.Dropping 1 objects from queue.
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/google/api_core/grpc_helpers.py", line 57, in error_remapped_callable
    return callable_(*args, **kwargs)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/grpc/_channel.py", line 550, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/grpc/_channel.py", line 467, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
	status = StatusCode.INVALID_ARGUMENT
	details = "One or more TimeSeries could not be written: Unrecognized region or location.: timeSeries[0]"
	debug_error_string = "{"created":"@1551716760.822772067","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1036,"grpc_message":"One or more TimeSeries could not be written: Unrecognized region or location.: timeSeries[0]","grpc_status":3}"
>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/src/opencensus/opencensus/common/transports/async_.py", line 105, in _thread_main
    self.exporter.emit(data)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/src/opencensus/opencensus/stats/exporters/stackdriver_exporter.py", line 153, in emit
    self.handle_upload(view_data)
[...]

PR #534, opened 2019-03-04
fix zone value in monitored resource

colincadams:
EDIT: Fixes https://github.com/census-instrumentation/opencensus-python/issues/533.


opencensus/common/monitored_resource/gcp_metadata_config.py
```diff
@@ -155,4 +159,8 @@ def _get_attribute(attribute_key):
             # to a normal string:
             attribute_value = attribute_value.decode('utf-8')
 
+        transformation = _ATTRIBUTE_TRANSFORMATIONS.get(attribute_key)
```
colincadams: instead of this transform, this could be structured like the java version, creating a method for each of the values and then add any transformation the the method for the given value.
https://github.com/census-instrumentation/opencensus-java/blob/25475e71ecf31faa3fbf57ca8077a67d9246318f/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/GcpMetadataConfig.java

PR #535, opened 2019-03-04
Componentize the remaining integration extensions

reyang:
This PR follows the approach in #532 for all the remaining extensions.


contrib/opencensus-ext-django/setup.py
```diff
@@ -0,0 +1,51 @@
...
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+    ],
+    description='OpenCensus Django Integration',
+    include_package_data=True,
+    long_description=open('README.rst').read(),
+    install_requires=[
+        'Django >= 1.11.0, <= 1.11.20',
```
c24t: Not `< 2.2.0`? We support versions later than `1.11.20` now.

Thinking about this again, setting a maximum version for ext dependencies might come back to bite us because either (1) we rely on dependencies following semantic versioning and set the max version as the next minor version to be released, in which case we fail if a patch update breaks our integration or (2) we set the max version as the current version (including the patch level) and are forced to update this file with each dependency release.

I don't have a proposed solution here, just noting that this might cause problems for dependencies that don't follow semvar or have different versioning semantics than we expect.

contrib/opencensus-ext-dbapi/opencensus/ext/dbapi/__init__.py
```diff
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from opencensus.trace.ext.dbapi import trace
+from opencensus.ext.dbapi import trace
```
c24t: As in https://github.com/census-instrumentation/opencensus-python/pull/529#pullrequestreview-208769605, why not move the contents of `trace.py` here?

contrib/opencensus-ext-django/setup.py
reyang: I actually tried Django 2.x and it failed the Py3.6 system test. https://github.com/census-instrumentation/opencensus-python/commit/eae1dee960080e48f77f6b6a3def2926fc4f9b2f

I will try to put the maximum Django version that we support here.

contrib/opencensus-ext-dbapi/opencensus/ext/dbapi/__init__.py
reyang: I was trying to avoid too many changes in this PR. Both `__init__.py` and `trace.py` were copied from the original `dbapi` folder, with the minimum change to replace namespace. We can do the clean up in a separate PR after the componentization is done.

contrib/opencensus-ext-django/setup.py
c24t: Ok, no need to fix it this PR. Can you open a ticket for the failing system test?

reyang: #538 created.

PR #536, opened 2019-03-05
Add TraceLogger

c24t:
This PR adds a logger that implements the [opencensus log correlation spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/LogCorrelation.md). The logger adds custom `traceId`, `spanId`, and `spanSampled` attributes to generated `LogRecords`.

See @sebright's [log4j2 integration](https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/OpenCensusTraceContextDataInjector.java) ([PR](https://github.com/census-instrumentation/opencensus-java/pull/1371)) for a similar change in the java client.

cc @otoolep, @sebright


opencensus/log.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2019, OpenCensus Authors
```
reyang: We need to put this inside a child directory.
`./opencensus/` is a namspace path, log.py will get removed if user uninstalls other packages like opencensus-ext-mysql.

c24t: Right, I started this branch before your changes landed and hadn't updated the paths yet. Fixed in 24fea93.

opencensus/common/monitored_resource/gcp_metadata_config.py
```diff
@@ -37,6 +37,10 @@
     ZONE_KEY: 'instance/zone'
 }
 
+_ATTRIBUTE_URI_TRANSFORMATIONS = {
+    _GCE_ATTRIBUTES[ZONE_KEY]:
+        lambda v: v[v.rfind('/') + 1:] if '/' in v else v
```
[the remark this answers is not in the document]
c24t: This is actually from #534. I picked it up from master, don't know why it appeared in this diff.

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+
+    :rtype: :class:`LogAttrs`
+    :return: The current span's trace ID, span ID, and sampling decision.
+    """
+    try:
+        tracer = execution_context.get_opencensus_tracer()
+        if tracer is None:
+            raise RuntimeError
+    except Exception:  # noqa
+        _meta_logger.error("Failed to get opencensus tracer")
```
reyang: minor: do we prefer to use single quote for literal strings (and use double quote only if it could help to avoid backslashes)?

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+# limitations under the License.
+
+from collections import namedtuple
+from copy import copy
+import logging
+
+from opencensus.trace import execution_context
+
+
+_meta_logger = logging.getLogger(__name__)
```
reyang: @c24t do you think we would encourage the library and extensions to leverage_meta_logger for internal logs (from the lib/exts)?

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+    try:
+        span_id = tracer.span_context.span_id
+        if span_id is None:
+            span_id = ATTR_DEFAULTS.span_id
+    except Exception:  # noqa
+        _meta_logger.error("Failed to get opencensus span ID")
+        span_id = ATTR_DEFAULTS.span_id
+
+    try:
+        sampling_decision = tracer.span_context.trace_options.get_enabled
```
reyang: Curious, why do we want to be specific about trace_options.get_enabled instead of the whole trace_options (e.g. if there are other flags in the future)?

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+from opencensus.trace import execution_context
+
+
+_meta_logger = logging.getLogger(__name__)
+
+TRACE_ID_KEY = 'traceId'
+SPAN_ID_KEY = 'spanId'
+SAMPLING_DECISION_KEY = 'traceSampled'
+
+LogAttrs = namedtuple('LogAttrs', ['trace_id', 'span_id', 'sampling_decision'])
```
reyang: Do we want to add `SpanContext.tracestate`? Tracestate is designed for interop with other tracing systems, do we expect extensions (or exporters) to extend the log_attrs?

For example, tracing system FOOBAR has a legacy correlation_id in `SpanContext.tracestate['correlation_id@foobar']`, the FOOBAR log exporter would expect to have that correlation_id associated with log entries.

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+    except Exception:  # noqa
+        _meta_logger.error("Failed to get opencensus tracer")
+        return ATTR_DEFAULTS
+
+    try:
+        trace_id = tracer.span_context.trace_id
+        if trace_id is None:
+            trace_id = ATTR_DEFAULTS.trace_id
+    except Exception:  # noqa
+        _meta_logger.error("Failed to get opencensus trace ID")
```
reyang: I wish to understand more about the design here.

Do we expect it to be a very rare case when there is no trace_id/span_id?
For example, if a batch application is fetching items from a queue, and process the queue items. I would expect the item processing logic to have trace_id/span_id/etc., while the app initialization and queue fetching logic could be context free. Would this generate too many error log entries?

opencensus/log/__init__.py
reyang: A similar question regarding tags, do we expect to put tags to the log entries in the future?

opencensus/log/__init__.py
c24t: This is the same convention we use for (internal) logging elsewhere, I just didn't name this one `logger` since I thought it might confuse users. I think `logger` is usually fine.

opencensus/log/__init__.py
c24t: This is just to follow the spec in this PR. I think there's good reason to include (a configurable subset of) the trace options, tracestate, tags, etc. in the future.

opencensus/log/__init__.py
c24t: I expect we will eventually, but we probably don't want to dump a lot of context into logs by default. In any case, not in this PR.

opencensus/log/__init__.py
c24t: I'd expect it to be rare that we don't have a span ID, and that we should always have a trace ID if OC is configured correctly. The broad exception handling here is because I really don't want an opencensus error to break application logging.

What do you think about adding config to attach a `NullHandler` to the `opencensus.log` logger so users don't see these errors unless they opt in?

opencensus/log/__init__.py
c24t: I use `'` for symbol-like strings and `"` for text-like strings as a matter of habit, but I'm happy to pick some other convention for this library.

opencensus/log/__init__.py
c24t: We also need to check `get_enabled` here because of the roundabout way we're handling sampling decisions. We'll need to change this after fixing #481.

reyang: Got it. Thanks for the background.

opencensus/log/__init__.py
reyang: That sounds okay to me.
Do we consider putting these logs at `debug level`?

c24t: I think `warn` is justified here if users have to specifically enable this logger, since I wouldn't expect to have to add a handler and set a lower log level to see these lines as a user.

Looking at https://docs.python.org/3.7/howto/logging.html#library-config, we've got three options:
1. don't set a default handler: `warn` and higher logs to stderr unless users configure a handler of their own
2. attach a `NullHandler` to the `opencensus.log` logger: users won't see logs at any level unless they configure their own handler
3. attach a `NullHandler` to the `opencensus` logger: same as (2) but applies to the whole package, users may enable logging for specific subpackages

We're doing (1) now, which seems like a sane default, but (3) might be the best long-term option.

reyang: Makes sense.

otoolep:
LGTM.

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+_meta_logger = logging.getLogger(__name__)
+
+TRACE_ID_KEY = 'traceId'
+SPAN_ID_KEY = 'spanId'
+SAMPLING_DECISION_KEY = 'traceSampled'
+
+LogAttrs = namedtuple('LogAttrs', ['trace_id', 'span_id', 'sampling_decision'])
+ATTR_DEFAULTS = LogAttrs('', '', False)
```
sebright: Would it be better to use the invalid trace and span IDs as defaults?  The Log4j version does that, though I don't know if it makes sense here.

https://github.com/census-instrumentation/opencensus-java/blob/b552db4432d5cc1a4767f49ed8da114631d06b99/contrib/log_correlation/log4j2/src/test/java/io/opencensus/contrib/logcorrelation/log4j2/OpenCensusTraceContextDataInjectorTest.java#L92-L108

opencensus/log/__init__.py
c24t: Particular integrations may not be enabled, but we should always get at least a trace ID here since we create a default tracer and span context on the first call to `get_opencensus_tracer`.

opencensus/log/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+        _set_extra_attrs(extra)
+        kwargs['extra'] = extra
+
+        return (msg, kwargs)
+
+
+# This is the idiomatic way to stack logger customizations, see
+# https://docs.python.org/3.7/library/logging.html#logging.getLoggerClass
+class TraceLogger(logging.getLoggerClass()):
+    """Logger class that adds openensus context attrs to records."""
```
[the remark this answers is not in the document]
c24t: Thanks!

opencensus/log/__init__.py
c24t: That seems like good default behavior, and we may as well keep this consistent between clients. Changed in 7fa82f2.

PR #540, opened 2019-03-05
Componentize the remaining integration extensions (2nd part)

reyang:
This PR follows the approach in #535 for the remaining extensions.

1. Reverted opencensus/trace/config_integration.py (which has been changed in #535) back since all the extensions are now under the same namespace.
2. Put test_utils.py under tests/unit/common.
3. Abandoned the opencensus/trace/ext folder.
4. Updated CHANGELOG.md.


opencensus/trace/config_integration.py
```diff
@@ -17,30 +17,23 @@
 
 log = logging.getLogger(__name__)
 
-SUPPORTED_INTEGRATIONS = {
-    'httplib': 'opencensus.trace.ext.httplib.trace',
-    'requests': 'opencensus.trace.ext.requests.trace',
-    'google_cloud_clientlibs': 'opencensus.trace.ext.google_cloud_clientlibs.trace',  # noqa
-    'threading': 'opencensus.trace.ext.threading.trace',
+SUPPORTED_INTEGRATIONS = ['httplib', 'mysql', 'postgresql', 'pymysql',
```
reyang: @c24t I wish to know your opinion on this.

This "config integration" seems to be introduced for:
1. Frameworks such like Django which prefers configuration.
2. Convenience, e.g. having settings.py file (or even environment variable) which can be updated easily.

From dependency perspective, this is like "reflection", the core package is "depending" on contrib packages.

reyang: I would prefer to keep this feature, but refactor in the following way:
1. remove hard-coded integration names.
2. scan `opencensus.ext` namespace, try to import the extensions.
3. log warning if module not found.

nox.py
```diff
@@ -24,11 +24,16 @@ def _install_dev_packages(session):
...
+    session.install('-e', 'contrib/opencensus-ext-grpc')
+    session.install('-e', 'contrib/opencensus-ext-httplib')
     session.install('-e', 'contrib/opencensus-ext-mysql')
     session.install('-e', 'contrib/opencensus-ext-postgresql')
     session.install('-e', 'contrib/opencensus-ext-pymysql')
     session.install('-e', 'contrib/opencensus-ext-pyramid')
+    session.install('-e', 'contrib/opencensus-ext-requests')
     session.install('-e', 'contrib/opencensus-ext-sqlalchemy')
+    session.install('-e', 'contrib/opencensus-ext-threading')
+    session.install('-e', 'contrib/opencensus-ext-google-cloud-clientlibs')
```
reyang: Here we don't follow the alphabetic order since `ext-google-cloud-clientlibs` has dependency on `ext-grpc` and `ext-requests`.
Ideally we don't want to maintain the order here, just having someting like "install all the packages under contrib, let pip figure out the dependency, resolve package dependency from contrib folder before consulting PyPI". Currently I haven't figured out how to do this.

opencensus/trace/config_integration.py
c24t: That sounds like a good plan to me, although scanning installed packages may introduce some problems. Another complicating factor is that not every integration implements `trace.trace_integration`.

reyang:
> Can you remove `wrapt` from `setup.py` now that requests is out in its own package?

Yep! I'll do the clean up in my next PR (clean up dependencies, move the samples).

PR #542, opened 2019-03-06
grpc client interceptor should set current span back to previous one in _callback

tjzo:
Client interceptor called execution_context.set_current_span(current_span) in _start_client_span, and called execution_context.set_current_span(current_span) in _callback again, so the previous span(maybe root span) was missing, and jaeger displayed as trace-without-root-span. I thought it should record the previous span and set it back in _callback function.

### DIFF
--- a/opencensus/.Python
+++ b/opencensus/.Python
@@ -1,0 +1,1 @@
+/usr/local/Cellar/python/3.7.2_1/Frameworks/Python.framework/Versions/3.7/Python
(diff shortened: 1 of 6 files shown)


tjzo:
I signed it!

tjzo:
I signed it!

reyang:
@tjzo please take a look at #539.

tjzo:
Thanks @reyang! this commit can fix the same issue.

PR #545, opened 2019-03-06
Fix subspan export bug

### LINKED ISSUE #537: Subspans not exported for stackdriver exporter and print exporter
When running the code below I am not able to see **subspans** in google stackdriver trace. I am only able to see root span `span1` but no `span2` 

opencensus version is `0.2.0` (installed with pip)

```python
from opencensus.trace import tracer as tracer_module
from opencensus.trace.exporters import stackdriver_exporter
from opencensus.trace import execution_context
from opencensus.trace.samplers.probability import ProbabilitySampler

exporter = stackdriver_exporter.StackdriverExporter(project_id="myprojectid")
sampler = ProbabilitySampler(rate=1.0)
tracer = tracer_module.Tracer(sampler=sampler, exporter=exporter)


with tracer.span(name='span1') as span1:
    x = 1
    with execution_context.get_current_span().span(name="span2") as span2:
        y = 1
```

same code will produce correct results (`span1` -> `span2` visible) when using `opencensus==0.1.5`

it also occurs when using default exporter:

```python 
from opencensus.trace import tracer as tracer_module
tracer = tracer_module.Tracer()

with tracer.span(name='root') as root:
    (lambda x: x)(1)
    with root.span(name="child") as child:
        (lambda x: x)(1)
```

outputs:

```python
[SpanData(name='root',...), SpanData(name='root', ...)]
```

c24t:
Fixes a bug from #419 that caused spans created with `Span.span` not to be exported from `ContextTracer`.

Fixes #537.

cc @int8


tests/unit/trace/tracers/test_context_tracer.py
```diff
@@ -67,6 +67,44 @@ def test_finish_with_spans(self):
...
+    def test_finish_with_tracer_subspans(self):
+        tracer = context_tracer.ContextTracer()
+        parent = tracer.start_span('parent')
+        child = tracer.start_span('child')
+        self.assertEqual(child.parent_span, parent)
+        tracer.exporter = mock.Mock()
+        tracer.finish()
+
+        self.assertEqual(tracer.exporter.export.call_count, 2)
+        [[[[c_sd]], _], [[[p_sd]], _]] = tracer.exporter.export.call_args_list
```
c24t: Subspans created with `Tracer.start_span` get exported separately. This test should pass on all versions.

tests/unit/trace/tracers/test_context_tracer.py
```diff
@@ -67,6 +67,44 @@ def test_finish_with_spans(self):
...
+    def test_finish_with_span_subspans(self):
+        tracer = context_tracer.ContextTracer()
+        parent = tracer.start_span('parent')
+        child = parent.span('child')
+        self.assertEqual(child.parent_span, parent)
+        tracer.exporter = mock.Mock()
+        tracer.finish()
+
+        self.assertEqual(tracer.exporter.export.call_count, 1)
+        [[[[c_sd, p_sd]], _]] = tracer.exporter.export.call_args_list
```
c24t: Subspans created with `Span.span` get exported together. This is the regression test: it fails on any rev between 25362e7 and this one.

This is testing some admittedly weird behavior. Creating a subspan this way means it's not set as the current span in the execution context.

tests/system/trace/basic_trace/basic_trace_system_test.py
```diff
@@ -65,8 +63,15 @@ def test_tracer(self):
...
+
         for span in spans:
-            if span.get('displayName').get('value') == 'root_span':
-                self.assertEqual(str(span.get('parentSpanId')), span_id)
+            if span['displayName']['value'] == 'root_span':
+                self.assertEqual(span['parentSpanId'], span_id)
+                self.assertEqual(span['childSpanCount'], 1)
             else:
-                self.assertEqual(span.get('parentSpanId'), parent_span_id)
+                self.assertEqual(span['displayName']['value'], 'child_span')
```
c24t: This test could have caught the bug: both exported spans were the same (root) span here.

PR #546, opened 2019-03-07
refactor namespace package - componentize ocagent trace exporter

reyang:


contrib/opencensus-exporter-ocagent-trace/README.rst
```diff
@@ -0,0 +1,16 @@
+OpenCensus OC-Agent Trace Exporter
+============================================================================
+
+Installation
+------------
+
+::
+
+    pip install opencensus-exporter-ocagent-trace
```
reyang: @c24t I wish to get your feedback on the package name and namespace.

contrib/opencensus-exporter-ocagent-trace/opencensus/ext/ocagent/trace_exporter/__init__.py
```diff
@@ -21,14 +21,14 @@
 
 from opencensus.common.transports import sync
 from opencensus.common.version import __version__
-from opencensus.trace.exporters import base
-from opencensus.trace.exporters.gen.opencensus.agent.common.v1 \
+from opencensus.ext.ocagent.trace_exporter import utils
```
reyang: Ideally we shouldn't have to modify these generated files.
I think we might want to put the generated files as a separate package/repo (e.g. `opencensus-protocol`), and reference it from `opencensus-exporter-ocagent-trace`.

contrib/opencensus-ext-django/tests/test_django_middleware.py
```diff
@@ -25,7 +25,7 @@
 from opencensus.trace.exporters import print_exporter
 from opencensus.trace.exporters import zipkin_exporter
 from opencensus.trace.exporters import jaeger_exporter
-from opencensus.trace.exporters.ocagent import trace_exporter
+from opencensus.ext.ocagent import trace_exporter
```
reyang: Looks like we need to decouple exporters from framework integrations.

https://github.com/census-instrumentation/opencensus-python/blob/80455edc4a461440caaa7cfd0bff873ea33a8244/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py#L43

contrib/opencensus-exporter-ocagent-trace/README.rst
reyang: I chose `opencensus-exporter-xyz` instead of `opencensus-ext` to make it easier for people to tell exporters from integrations.
I chose namespace like `opencensus.ext.ocagent.trace_exporter`, so it is possible for the trace exporter and stats/log exporter to share the same namespace `opencensus.ext.ocagent`.
This also makes it possible to have a single package `opencensus-exporter-ocagent` which provides multiple exporters in the future.

reyang: @tjzo I'd like to hear your feedback as well.

tjzo: I think the namespace opencensus.ext.ocagent. would be fine.
And I would like to name the package opencensus-ext-exporter-ocagent. Keep all the package name starts with the same prefix.
The go and ruby package name is opencensus-go-exporter-ocagent / opencensus-ruby-exporter-ocagent, but they are all in their own repo. If we just place them in this repo, I think opencensus-ext prefix is fine.

reyang: @tjzo Thanks for the quick reply! I'm fine with either `opencensus-ext-exporter-ocagent` or `opencensus-exporter-ocagent`.

@c24t Looks like OpenCensus Go does support "one exporter" for both trace and metrics/stats:
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/060e6898ab87d943e2ff1090efb927fbc5181cea/ocagent.go#L103

c24t: The distribution name, package name, and directory structure all look good to me, but it's a little odd that exporters don't have `ext` in the distribution name but do have it in the namespace/package name. Do you think it's more important to distinguish between exporter and integration distributions than packages?

What about using the `opencensus.exp` namespace for exporter code and using the prefix `opencensus-exp` for exporter distributions? E.g. `opencensus-exp-ocagent` This makes the distribution and package names more consistent (nice to have, but not necessarily a goal) and the distribution names shorter.

This is largely a matter of taste, and I'm fine with the names you proposed. Just a preference for shorter names.

contrib/opencensus-exporter-ocagent-trace/opencensus/ext/ocagent/trace_exporter/__init__.py
c24t: These generated files are specific to the agent though right? I think it's fine to leave them in the agent package.

For shared generated files, as far as integrations and exporters are concerned, what's the difference between leaving them in the core library and packaging them up separately?

nox.py
```diff
@@ -21,6 +21,9 @@
 def _install_dev_packages(session):
     session.install('-e', 'contrib/opencensus-correlation')
     session.install('-e', '.')
+
+    session.install('-e', 'contrib/opencensus-exporter-ocagent-trace')
```
c24t: FWIW I think you're right that it'd be fine to have a single distribution for both tracing and metrics exporters for a given backend. We do risk pulling in extra dependencies, but since (I suspect) many vendors use the same client for both, this shouldn't matter much in practice.

contrib/opencensus-exporter-ocagent-trace/opencensus/ext/ocagent/trace_exporter/__init__.py
reyang: Packaging them separately gives us a chance to publish the generated package automatically whenever we have a new protocol version, for example:
1. You make a change in `opencensus-proto` repo.
2. The `opencensus-protocol` package got build automatically and published to PyPI.
3. Existing ocagent exporter will not be affected since it has a strict dependency on the previous version of `opencensus-protocol` package.
4. If you're ready to move forward, update the ocagent exporter code and bump `opencensus-protocol` dependency version.

In this way we don't have to manually generate/copy files. It also makes the repo a bit smaller.

contrib/opencensus-exporter-ocagent-trace/README.rst
reyang: @tjzo @c24t probably we can just try not to distinguish between exporters and integrations.
How about just calling it `opencensus-ext-ocagent`?

I tend to avoid `exp` since we will have to use something like `prop` or `pro` for propagators if we decided to extract them.

reyang: I'll merge the PR for now and continue working on extracting other exporters.
Please let me know your feedbacks and we can adjust the package name anytime before we release.

PR #543, opened 2019-03-06
Integrate with pymongo

tjzo:
Just integrate with pymongo

PR #548, opened 2019-03-07
Refactor config_integration

reyang:


reyang:
@tjzo follow up on #543

opencensus/trace/config_integration.py
```diff
@@ -32,17 +25,19 @@ def trace_integrations(integrations, tracer=None):
     integrated = []
 
     for item in integrations:
+        module_name = 'opencensus.ext.{}.trace'.format(item)
+        try:
+            module = importlib.import_module(module_name)
+        except Exception as e:
+            log.warning('Failed to import module: {}'.format(item))
```
c24t: This might be misleading since the module that failed import was e.g. `opencensus.ext.httplib`, not just `httplib` -- and installing `httplib` wouldn't fix anything.

Why log this twice?

reyang: Good point!

Regarding logging twice, this was the original behavior which I decided to carry forward. The 1st warning message is OpenCensus level warning (warn the user that certain integration didn't happen), the 2nd message is the actual runtime exception with the call stack (e.g. which is if some underlying package dependency is broken).

reyang:
> LGTM assuming building `SUPPORTED_INTEGRATIONS` by inspecting installed packages is out of scope for this PR, or that we want to remove it completely.

I intend to remove it completely as the concept is misleading.

If we scan the ext folder and try to discover all the packages which provides config integration (by looking at the `trace_integration`, this would only give "currently installed integrations".

An external package (e.g. https://pypi.org/project/ocredis/) could implement the config integration, the current hard-coded approach won't scale.

reyang:
@tjzo what's your opinion?

tjzo:
I think we should have an abstract class, and all the subclass should implement the trace_integration method. By scanning the ext package, we should load all the subclasses. Also the scan method should accept other packages(both ext and user specified packages are scanned), so external modules can be loaded as well.

reyang:
Why would we want to scan all the ext packages? There is perf cost, do we know what are the outcomings?

tjzo:
So just scan the ext package, that is accepted. And the parameter should be list of subclasses rather than external package, avoiding scanning external packages.

tjzo:
> accepted

acceptable, misspelling

tjzo:
> > LGTM assuming building `SUPPORTED_INTEGRATIONS` by inspecting installed packages is out of scope for this PR, or that we want to remove it completely.
> 
> I intend to remove it completely as the concept is misleading.
> 
> If we scan the ext folder and try to discover all the packages which provides config integration (by looking at the `trace_integration`, this would only give "currently installed integrations".
> 
> An external package (e.g. https://pypi.org/project/ocredis/) could implement the config integration, the current hard-coded approach won't scale.

[ocredis](https://pypi.org/project/ocredis/) doesn't code so well, I think we need an integration like [redis-opentracing](https://pypi.org/project/redis-opentracing/)
My company is using redis, but I don't wan't to use [ocredis](https://pypi.org/project/ocredis/). 😂 Trying to make a better solution.

c24t:
> By scanning the ext package, we should load all the subclasses. Also the scan method should accept other packages(both ext and user specified packages are scanned), so external modules can be loaded as well.

If the user is specifying packages/classes to scan already then why would they use `SUPPORTED_INTEGRATIONS` instead of just passing the package names to `trace_integrations`? I don't see a compelling use case for `SUPPORTED_INTEGRATIONS` and think it does make sense to remove it.

opencensus/trace/config_integration.py
reyang: Fixed the misleading warning message.

reyang:
> [ocredis](https://pypi.org/project/ocredis/) doesn't code so well, I think we need an integration like [redis-opentracing](https://pypi.org/project/redis-opentracing/)
> My company is using redis, but I don't wan't to use [ocredis](https://pypi.org/project/ocredis/). 😂 Trying to make a better solution.

@tjzo Would you help to create an issue in https://github.com/opencensus-integrations/ocredispy/issues and describe the problems? Thanks!

PR #552, opened 2019-03-11
Replace isoformat with strftime

### LINKED ISSUE #502: Standardize timestamps
Right now we use three types of timestamp internally: tz-naive `datetime`s (`datetime.now()`), UTC `datetimes` (`datetime.utcnow()`), and UTC date strings (`datetime.utcnow().isoformat() + 'Z'`).

We should pick a single internal timestamp type. I suggest that we use `int` nanoseconds or milliseconds since the UTC epoch.

Epoch times are still affected by leap seconds -- there are timestamps that describe two separate actual times during positive leap seconds, and (theoretically) timestamps that don't correspond to any actual time because of negative leap seconds. This poses a problem for tracing because we may report spans that end before they start, or are otherwise nonsensical or out of order. This is an edge case for sure, but one that we need to consider.

The java client solves this (and other problems) by passing around a [monotonic `Clock`](https://github.com/census-instrumentation/opencensus-java/blob/4125599ad34e074eb604428e09c8468020f84054/impl_core/src/main/java/io/opencensus/implcore/trace/SpanBuilderImpl.java#L100-L108). We should consider doing something similar in python.

Another option is the ["leap smear"](https://developers.google.com/time/smear). Google's NTP and [AWS](https://aws.amazon.com/blogs/aws/look-before-you-leap-the-coming-leap-second-and-aws/) take this approach, but this is likely to cause problems for clients.

### LINKED ISSUE #547: datetime parsing in tracer fails sometimes
Stacktrace:

```
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/tracers/context_tracer.py", line 123, in end_span
    self.exporter.export(span_datas)
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/exporters/jaeger_exporter.py", line 158, in export
    self.transport.export(span_datas)
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/exporters/transports/sync.py", line 23, in export 
    self.exporter.emit(span_datas)
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/exporters/jaeger_exporter.py", line 138, in emit
    jaeger_spans = self.translate_to_jaeger(span_datas)
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/exporters/jaeger_exporter.py", line 177, in translate_to_jaeger
    start_timestamp_ms = timestamp_to_microseconds(span.start_time)
  File "/env/local/lib/python2.7/site-packages/opencensus/trace/utils.py", line 71, in timestamp_to_microseconds
    timestamp_str = datetime.datetime.strptime(timestamp, ISO_DATETIME_REGEX)
  File "/usr/lib/python2.7/_strptime.py", line 332, in _strptime
    (data_string, format))
ValueError: time data '2019-03-07T09:24:40Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'
```

The problem is python's standard library module `datetime.datetime`. The `isoformat()` method generates a formatted datetime string _without_ milliseconds if milliseconds is `0`.

c24t:
Replace calls to `datetime.isoformat` with a util function to fix #547. Ultimately we should replace internal `datetime`s with ints to fix #502.


opencensus/common/utils/__init__.py
```diff
@@ -69,6 +69,13 @@ def check_str_length(str_to_check, limit=MAX_LENGTH):
     return (result, truncated_byte_count)
 
 
+def isoz(ts=None):
```
reyang: to_iso_string? The name isoz could be confusing.

c24t: Sure, changed in 77a1227.

PR #555, opened 2019-03-11
v0.3.x master version updates

c24t:
These are the version number changes from #554 applied to `master` (as opposed to `v0.3.x`). Note that we're using dev version numbers on the development branch only.

### DIFF
--- a/contrib/opencensus-correlation/version.py
+++ b/contrib/opencensus-correlation/version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.2.dev0'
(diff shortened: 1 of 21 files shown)


contrib/opencensus-correlation/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.2.dev0'
+__version__ = '0.4.dev0'
```
c24t: Dev releases are ordered before corresponding regular releases (see https://www.python.org/dev/peps/pep-0440/#developmental-releases), so this should generally be one minor version ahead of the latest released version.

Issue #454, opened 2019-01-17
Refactor stats exporters to use metrics data model


c24t:
Stats exporters should use the new [metrics data model](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto) in place of stats (i.e. `View`, `ViewData`, and aggregation classes).

See the same change in the java client for the [stackdriver](https://github.com/census-instrumentation/opencensus-java/pull/1501) and [prometheus](https://github.com/census-instrumentation/opencensus-java/pull/1509) exporters.

This change depends on #335.

PR #427, opened 2018-12-05
Avoid resolving monitored resource every exported timeseries.

bogdandrutu:
Few more fixes:
1. Use gke_container as resource type instead of k8s_container when provided as gke_container.
2. Add support for k8s_container when provided in the options
3. Use provided default_monitored_labels if type is one of gce_instance, gke_container, aws_ec2_instance.

PR #560, opened 2019-03-18
Use metrics instead of stats in SD exporter

c24t:
This is the first of several breaking changes to replace the stats data model (`ViewData`,  `AggregationData`, etc.) with metrics in stats/metrics exporters. 

This PR changes `StackdriverStatsExporter` to:
 - use `Metric` instead of `ViewData` internally
 - remove `on_register_view` hook, lazily register metric descriptors with stackdriver at export time instead of on view creation
 - prepare to move from push to poll model by removing transports, `emit`, etc.

The meat of this PR is in `create_time_series_list`, which now takes a list of `Metric`s.

cc @colincadams 

Addresses #454.

### DIFF
--- a/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
+++ b/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
@@ -12,14 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime
 import itertools
 import logging
 import os
 import platform
 import re
 import string
+import threading
 
-from datetime import datetime
 from google.api_core.gapic_v1 import client_info
 from google.cloud import monitoring_v3
 
@@ -27,9 +28,11 @@ from opencensus.common import utils
 from opencensus.common.monitored_resource import monitored_resource
 from opencensus.common.transports import async_
 from opencensus.common.version import __version__
+from opencensus.metrics.export import metric_descriptor
 from opencensus.stats import aggregation
 from opencensus.stats import base_exporter
 from opencensus.stats import measure
+
 
 MAX_TIME_SERIES_PER_UPLOAD = 200
 OPENCENSUS_TASK = "opencensus_task"
@@ -110,18 +113,14 @@ class Options(object):
         return self._default_monitoring_labels
 
 
-class StackdriverStatsExporter(base_exporter.StatsExporter):
+class StackdriverStatsExporter(object):
     """Stats exporter for the Stackdriver Monitoring backend."""
 
-    def __init__(self,
-                 options=Options(),
-                 client=None,
-                 default_labels={},
-                 transport=async_.AsyncTransport):
+    def __init__(self, options=Options(), client=None):
         self._options = options
         self._client = client
-        self._transport = transport(self)
-        self._default_labels = default_labels
+        self._md_cache = {}
+        self._md_lock = threading.Lock()
 
     @property
     def options(self):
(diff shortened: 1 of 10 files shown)


contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -110,18 +113,14 @@ def default_monitoring_labels(self):
         return self._default_monitoring_labels
 
 
-class StackdriverStatsExporter(base_exporter.StatsExporter):
+class StackdriverStatsExporter(object):
```
c24t: We'll update `StatsExporter` after making the same changes to the prometheus exporter. This is an API change.

opencensus/metrics/export/metric_descriptor.py
```diff
@@ -90,13 +90,30 @@ class MetricDescriptorType(object):
...
+    }
+
     @classmethod
     def to_type_class(cls, metric_descriptor_type):
         try:
             return cls._type_map[metric_descriptor_type]
         except KeyError:
             raise ValueError("Unknown MetricDescriptorType value")
 
+    @classmethod
```
c24t: Consider replacing this one-off enum class with `enum34`.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -131,209 +130,174 @@ def options(self):
...
-        elif view_aggregation.aggregation_type is agg_type.LASTVALUE:
-            metric_kind = metric_desc.MetricKind.GAUGE
-            if isinstance(view_measure, measure.MeasureInt):
-                value_type = metric_desc.ValueType.INT64
-            if isinstance(view_measure, measure.MeasureFloat):
-                value_type = metric_desc.ValueType.DOUBLE
+    def create_time_series_list(self, metric):
+        tsl = []
+        for ts in metric.time_series:
+            tsl.append(self._convert_series(metric, ts))
```
reyang: Curious, would this `[self._convert_series(metric, ts) for ts in metric.time_series]` give better readability and perf? And would tuple work here?

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -131,209 +130,174 @@ def options(self):
...
+        start_time = sd_point.interval.start_time
+        start_time.seconds = int(timestamp_start)
+        start_secs = start_time.seconds
+        start_time.nanos = int((timestamp_start - start_secs) * 1e9)
+
+    def get_descriptor_type(self, oc_md):
+        """Get a SD descriptor type for an OC metric descriptor."""
+        return namespaced_view_name(oc_md.name, self.options.metric_prefix)
+
+    def get_metric_descriptor(self, oc_md):
```
reyang: Minor suggestion, probably we can use a predefined dict to do the lookup?

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -131,209 +130,174 @@ def options(self):
...
+            metric.descriptor.name, self.options.metric_prefix)
+
+        series.metric.labels[OPENCENSUS_TASK] = get_task_value()
+
+        for key, val in zip(metric.descriptor.label_keys, ts.label_values):
+            if val.value is not None:
+                safe_key = sanitize_label(key.key)
+                series.metric.labels[safe_key] = val.value
+
+        set_monitored_resource(series, self.options.resource)
```
colincadams: Outside the scope of this pr, but I'm curious what the motivation is to do the monitored resource detection every time we export every series. Instead would it make sense to detect and convert the labels once when the exporter is created? Then we can just set it on each series each time.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
colincadams: agreed. In fact one already exists for gauge: `metric_utils.is_gauge`. I think it would make sense to create one for cumulative and for the value types as well.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
colincadams: nvm, #427 addresses this

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: Done in bce0b86, good call.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: Changed in bce0b86. No strong opinion on lists vs tuples, but it's fine for the caller to modify this one and we generally use lists internally like this.

reyang: Got it. Thanks!

c24t:
e89c84e is a mostly mechanical change to the existing tests to convert `ViewData`s to metrics before converting to timeseries. These should be replaced with `Metric`s or mocks before merging into master, and we need new tests to cover the exporter's new polling behavior.

Note that this is being merged into the `stackdriver-metrics` branch. I'm separating the changes to the stackdriver client here to make the diffs easier to digest, and will open a PR from this branch once all the metrics changes in the client are done.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -42,6 +46,25 @@
 EPOCH_PATTERN = "%Y-%m-%dT%H:%M:%S.%fZ"
 GLOBAL_RESOURCE_TYPE = 'global'
 
+# OC metric descriptor type to SD metric kind and value type
+OC_MD_TO_SD_TYPE = {
```
songy23: There should be one more mapping for `Summary` type. See Java: https://github.com/census-instrumentation/opencensus-java/pull/1591. You can file an issue to add support for it later.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -131,209 +150,157 @@ def options(self):
...
+    def _convert_series(self, metric, ts):
+        """Convert an OC timeseries to a SD series."""
+        series = monitoring_v3.types.TimeSeries()
+        series.metric.type = namespaced_view_name(
+            metric.descriptor.name, self.options.metric_prefix)
+
+        series.metric.labels[OPENCENSUS_TASK] = get_task_value()
+
+        for key, val in zip(metric.descriptor.label_keys, ts.label_values):
+            if val.value is not None:
```
songy23: If value is `None`, consider replacing with an empty string. Other SD backend will return an error because key-value doesn't match.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -131,209 +150,157 @@ def options(self):
...
+            sd_point.value.double_value = float(point.value.value)
+
+        elif (metric.descriptor.type ==
+              metric_descriptor.MetricDescriptorType.GAUGE_INT64):
+            sd_point.value.int64_value = int(point.value.value)
+
+        elif (metric.descriptor.type ==
+              metric_descriptor.MetricDescriptorType.GAUGE_DOUBLE):
+            sd_point.value.double_value = float(point.value.value)
+
```
songy23:
```python
elif (metric.descriptor.type == 
       metric_descriptor.MetricDescriptorType.SUMMARY):
# TODO...
```

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
songy23: Apologies please ignore this comment, that's a requirement for Prometheus.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: Opened https://github.com/census-instrumentation/opencensus-python/issues/567.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: Added in a98553e.

c24t:
Screen Shot 2019-03-20 at 4 12 17 PM

> `ResourceExhausted: 429 Your metric descriptor quota has been exhausted.`

Our stackdriver system tests create (and then don't delete) metric descriptors in GCP.

PR #568, opened 2019-03-19
fix lint

reyang:
I made a mistake while refactoring the packages - all the contrib packages are not covered by lint and code coverage test.
This PR should fix the lint problem.

### DIFF
--- a/nox.py
+++ b/nox.py
@@ -116,8 +116,8 @@ def lint(session):
 
     session.run(
         'flake8',
-        '--exclude=opencensus/trace/exporters/gen/',
-        'opencensus/', 'tests/', 'examples/')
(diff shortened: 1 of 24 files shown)


nox.py
```diff
@@ -116,8 +116,8 @@ def lint(session):
 
     session.run(
         'flake8',
-        '--exclude=opencensus/trace/exporters/gen/',
-        'opencensus/', 'tests/', 'examples/')
+        '--exclude=contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/trace_exporter/gen/',
```
reyang: Ideally we want to have a config file (e.g. lint.rc) under the ocagent ext folder. Flake doesn't seem to support this, probably we should revisit when we switch to pylint.

PR #571, opened 2019-03-21
Use new GCP project credentials

c24t:
This PR replaces the (encrypted) GCP credentials used by CI for stackdriver system tests. The new credentials point at a new GCP project. @liyanhui1228 is the owner of the previously-used project, and we were overdue to change ownership. The motivating reason for making this change now is that we just [blew out our stackdriver metrics descriptor quota](https://github.com/census-instrumentation/opencensus-python/pull/560#issuecomment-475064652).

I'll update the [CI env vars](https://circleci.com/gh/census-instrumentation/opencensus-python/edit#env-vars) before merging this PR.


reyang:
Thanks @c24t! I just hit the same CI error.

c24t:
Updated the key for an account with more liberal permissions, let's see if CI passes...

Issue #564, opened 2019-03-19
Refactor context


reyang:
This is to support https://github.com/census-instrumentation/opencensus-specs/issues/247.
1. Provide a generic context, which is maintained collaboratively by the Python runtime and the OpenCensus library.
2. Provide asyncio/async/await support.
3. Refactor existing context (span_context, trace/execution_context, tags/execution_context, correlation_context).
4. Provide callback API for span start/end.

@c24t @bogdandrutu @discostu105

bogdandrutu:
I think this should be under context/ not under contrib. The core library should not depend on a contrib package.

reyang:
@bogdandrutu
1. Do you mean we want to have this package under context/ instead contrib/ to avoid the confusion?
2. Or, do you mean this context should be part of the `opencensus` package instead of having its own package?

Having a small package of the generic context encourages other libraries to interop (similar like gRPC Context in Java) without having to depend on OpenCensus.

c24t:
I just talked to @bogdandrutu, he means putting the package under `opencensus/context/` or a new top level `context/`.

reyang:
Thanks for clarification @c24t ! Please let me know your preference on the naming and folder structure.

Considering we will need to move CorrelationContext and TraceContext as well, probably we just rename `contrib` to a general name `packages`?

reyang:
Personally I like [the way how OpenCensus Node.js is organized](https://github.com/census-instrumentation/opencensus-node/tree/master/packages), so we avoid inventing 1st class and 2nd class citizens. It also gives us flexibility on having `opencensus-api` and `opencensus-impl` in the future.

c24t:
Moving everything into a top-level `packages/` sounds fine to me, and would make it easy for people browsing the library to find a particular package quickly. But I'm more than happy having first-class `opencensus/` and `context/` dirs and a second-class `contrib/`. Up to you.

bogdandrutu:
I preferred the way @c24t proposed. 

@reyang why do we need to move the CorrelationContext there? Correlation context is part of opencensus/tags. I am not sure I understand.

reyang:
@bogdandrutu CorrelationContext is a separate package which implements the W3C spec.

The opencensus/tags will be implemented leveraging CorrelationContext, while CorrelationContext itself could be used for other purpose (e.g. application developer can put TransactionId and UserIdentity).

reyang:
[...]

PR #573, opened 2019-03-21
Refactor execution_context to use RuntimeContext

reyang:
This is part of the #564 effort.

Today we have `trace/execution_context`, `tags/execution_context` and `stats/executino_context` built on top of Thread-local Storage, which I plan to refactor.

This PR simply refactored execution_context from Thread-local Storage to RuntimeContext (introduced in #566).

### DIFF
--- a/opencensus/trace/execution_context.py
+++ b/opencensus/trace/execution_context.py
@@ -76,4 +76,5 @@ def clean():
 
 
 def clear():
+    """Clear the context, used in test."""
     clean()
--- a/opencensus/trace/execution_context.py
+++ b/opencensus/trace/execution_context.py
@@ -12,81 +12,68 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import threading
-
+from opencensus.common.runtime_context import RuntimeContext
 from opencensus.trace.tracers import noop_tracer
 
-_thread_local = threading.local()
+_attrs_slot = RuntimeContext.register_slot('attrs', lambda: {})
+_current_span_slot = RuntimeContext.register_slot('current_span', None)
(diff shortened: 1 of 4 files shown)


opencensus/trace/execution_context.py
```diff
@@ -12,81 +12,68 @@
...
-    if hasattr(_thread_local, 'current_span'):
-        delattr(_thread_local, 'current_span')
-    if hasattr(_thread_local, 'tracer'):
-        delattr(_thread_local, 'tracer')
+    _attrs_slot.clear()
+    _current_span_slot.clear()
+    _tracer_slot.clear()
 
 
 def clear():
```
reyang: @c24t I think having a test-only method exposed publicly is evil. I plan to remove it or make it "private". Please let me know your thoughts.

BTW, `_thread_local.__dict__.clear()` will remove Thread-local storage introduced by other components (e.g. stats), not sure if this is intended. @liyanhui1228 might have some background on this?

opencensus/trace/execution_context.py
```diff
@@ -12,81 +12,69 @@
...
-    _span = get_current_span()
-    _attrs = get_opencensus_attrs()
-    return _tracer, _span, _attrs
+    attrs = RuntimeContext.attrs
+    current_span = RuntimeContext.current_span
+    tracer = RuntimeContext.tracer
+    return tracer, current_span, attrs
 
 
 def set_opencensus_full_context(tracer, span, attrs):
```
reyang: This `set_opencensus_full_context` probably should be removed (in another PR):
1. It is not the "full context" as it doesn't cover tags.
2. `RuntimeContext.apply` provides the intended functionality, which can be used by propagating context to explicit threads.

PR #562, opened 2019-03-19
V0.3.1 ext release

c24t:
Version number updates for ext packages added for #445. Once this PR is merged into the `v0.3.x` branch we can create a `v0.3.1` release for the core library and push the new (`0.1.0`) ext packages to PyPI.

In the long term we need a separate release process for ext packages. See #544 for details.

Fixes #561.

PR #556, opened 2019-03-12
Refactor stats exporters into ext packages

reyang:
This is part of the effort #445.

1. Extracted Prometheus Stats exporter into `opencensus.ext.prometheus`.
2. Extracted Stackdriver Stats exporter and reuse `opencensus.ext.stackdriver`.
3. Renamed `opencensus.stats.exporters.base` to `opencensus.stats.base_exporter`.
4. Moved Prometheus/Stackdriver specific document/samples to ext packages.

@c24t Would you prefer to use the stackdriver exporter package for both trace and stats, or separate them into two packages?

PR #574, opened 2019-03-22
Prepare for prometheus ext release

c24t:
This PR includes only the prometheus changes from #556, to be merged onto the `v0.3.x` release branch. The prometheus client wasn't released as part of #562 because it wasn't on master at the time.

There have been breaking changes on master since the release (moving the `opencensus.stats.exporter` package), so we can't blindly copy from master without making a new core library version release, which meant doing some package surgery here.

We'll likely release the other changes to contrib packages from #556 (docs, examples, etc.) together with the next version of the core library. I'll upload `opencensus-ext-prometheus` to PyPI manually once this PR is merged.


c24t:
Uploaded the prometheus exporter package to https://pypi.org/project/opencensus-ext-prometheus/.

PR #576, opened 2019-03-22
Fix code coverage for ext packages

reyang:
Include extension packages in code coverage test.

### DIFF
--- a/nox.py
+++ b/nox.py
@@ -61,7 +61,7 @@ def unit(session, py):
     session.run(
         'py.test',
         '--quiet',
-        '--cov=contrib', '--cov=opencensus',
+        '--cov=opencensus', '--cov=contrib',
         '--cov-append',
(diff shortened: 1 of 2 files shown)


nox.py
```diff
@@ -61,7 +61,6 @@ def unit(session, py):
     session.run(
         'py.test',
         '--quiet',
-        '--cov=opencensus',
```
c24t: You don't need to specify a project? I've been running with `opencensus` for the core library and `opencensus.ext` for the contrib packages.

reyang: I was hoping that by removing this line, `[run] source` in `.coveragerc` would take effect, but I realized it was overwritten by `pytest-cov`:
https://github.com/pytest-dev/pytest-cov/blob/41eea2e3e28e6f6849104d6526bb633e638f8a0f/src/pytest_cov/plugin.py#L56

PR #577, opened 2019-03-25
Switch SD exporter to polling

c24t:
This PR adds support for polling metrics from the stackdriver exporter. It builds on #560, which removed the ability to _push_ metrics to the exporter via `MeasureToViewMap.record`.

I've added a `PeriodicTask` class (similar to the java client's [`IntervalMetricReader`](https://github.com/census-instrumentation/opencensus-java/blob/b47f0708421c9a6dc352e3e85e2c64d2f22043a8/exporters/metrics/util/src/main/java/io/opencensus/exporter/metrics/util/IntervalMetricReader.java)) that periodically calls `exporter.export(producer.get_metrics())`. To keep the API changes to a minimum I've changed `new_stats_exporter` to create and start the task, but we may want to change this to make the exporter ignorant of the transport instead.
  
@songy23 and @colincadams I'm interested to hear your thoughts on the API changes here.

### DIFF
--- a/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
+++ b/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
@@ -212,17 +212,6 @@ class TestStackdriverStatsExporter(unittest.TestCase):
 
         self.assertEqual(registered_exporters, 1)
 
-    def test_set_metric_labels(self):
-        series = monitoring_v3.types.TimeSeries()
-        tag_value = tag_value_module.TagValue("1200")
-        stackdriver.set_metric_labels(series, VIDEO_SIZE_VIEW, [tag_value])
-        self.assertEqual(len(series.metric.labels), 2)
-
-    def test_set_metric_labels_with_None(self):
-        series = monitoring_v3.types.TimeSeries()
-        stackdriver.set_metric_labels(series, VIDEO_SIZE_VIEW, [None])
-        self.assertEqual(len(series.metric.labels), 1)
-
     @mock.patch('os.getpid', return_value=12345)
     @mock.patch(
         'platform.uname',
--- a/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
+++ b/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
@@ -364,11 +364,11 @@ class TestAsyncStatsExport(unittest.TestCase):
     """Check that metrics are exported using the exporter thread."""
 
     @mock.patch('opencensus.ext.stackdriver.stats_exporter'
-                '.stats.Stats')
+                '.stats.stats')
     def test_export_empty(self, mock_stats, mock_client):
         """Check that we don't attempt to export empty metric sets."""
 
-        mock_stats.return_value.get_metrics.return_value = []
+        mock_stats.get_metrics.return_value = []
 
         with patch_sd_transport():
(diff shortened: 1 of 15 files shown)


contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
```diff
@@ -115,20 +124,23 @@ def test_constructor_param(self):
         self.assertEqual(exporter.options.project_id, project_id)
 
     def test_blank_project(self):
-        self.assertRaises(ValueError, stackdriver.new_stats_exporter,
-                          stackdriver.Options(project_id=""))
+        with patch_sd_transport():
```
c24t: Ideally the `PeriodicTask` would shut itself down gracefully at the end of the test and we wouldn't have to patch tests that don't rely on the transport.

contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
```diff
@@ -115,20 +124,23 @@ def test_constructor_param(self):
...
-            exporter_created = stackdriver.new_stats_exporter(
-                stackdriver.Options(project_id=1))
+        with patch_sd_transport():
+            with patch_client:
+                exporter_created, transport = stackdriver.new_stats_exporter(
+                    stackdriver.Options(project_id=1))
 
         self.assertIsInstance(exporter_created,
                               stackdriver.StackdriverStatsExporter)
+        transport.stop()
```
c24t: Ideally we also wouldn't have to call `stop`.

opencensus/metrics/transport.py
```diff
@@ -0,0 +1,143 @@
+# Copyright 2019, OpenCensus Authors
```
c24t: Note that this `transport` serves a different purpose than `opencensus.common.transports`.

On master:
```
exporter.export ->  exporter.transport.export -> exporter.emit -> ... -> exporter.create_time_series
```
vs. on this branch:
```
background_thread -> exporter.export -> exporter.create_time_series
```

I'm open to a different name or structure here.

contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
c24t: This was as easy as making the class a daemon thread, fixed in d693cf8.

opencensus/metrics/transport.py
songy23: This serves a similar purpose to https://github.com/census-instrumentation/opencensus-go/blob/master/metric/metricexport/reader.go#L62, what about `interval_metric_reader`?

colincadams: High level question about this flow since I'm a little behind how `stackdriver-metrics` works. What happens when `measure_to_view_map.export` is called now? I see it is deprecated is it a no-op somehow?

c24t:
> Would the goal be to eventually do this for all exporters

All stats/metrics exporters at least. The idea is that metrics are generated by producers, and the whole stats module is a single producer. The API for exporting traces will still let us push.

opencensus/metrics/transport.py
c24t: Nothing! We should be able to remove this and many other supporting classes (e.g. `ViewData`) once stackdriver and prometheus are migrated.

c24t: @songy23 it's long, but maybe worth it for the consistency between packages.

Issue #402, opened 2018-11-20
Migrate to CircleCI 2.0


c24t:
We're seeing a deprecation notice on https://circleci.com/gh/census-instrumentation/opencensus-python, time to upgrade.

![screen shot 2018-11-19 at 5 42 44 pm](https://user-images.githubusercontent.com/43799105/48746084-c56b8000-ec22-11e8-9733-d96d7f18a8bb.png)

c24t:
See https://circleci.com/docs/2.0/workflows/#git-tag-job-execution for notes on updating our build-on-tag workflow.

c24t:
Consider moving to travis, like the other OC client libraries.

reyang:
@lzchen

bogdandrutu:
OpenTelemetry is also going to use Travis.

lzchen:
Travis CI gives us a lot of functionality and freedom for configurations, especially related to running different versions of Python. To take advantage of these customizations, we could consider moving our build automation from nox to tox (https://tox.readthedocs.io/en/latest/). There's a plugin available that integrates tox with TravisCI (https://tox-travis.readthedocs.io/en/stable/)

Nox's ability to run specific versions of Python per function can be replicated using functionality available in TravisCI. Nox also lacks a lot of the configurations that would be useful (such as caching dependencies). There's also the case of nox not being widely used so there's is very little support or documentation on it.

lzchen:
There's also version specific logic that exists which will fail the coverage aggregation check (because nox aggregates results from different run versions). A possibility could be lowering the coverage threshold for aggregated test results.

reyang:
@lzchen, what's your thinking on running test against different versions of python on local dev box?

lzchen:
@reyang  Running tests on different versions locally would be done sequentially, which might take a lot of time, especially with the installation of dependencies. With Travis, new virtual envs would be boot up at the same time and run in parallel and we can configure them to share from the same dependency list.

bu3alwa:
Is this still needed? If so can I give it a go?

lzchen:
Yes! The `tox.ini` file is already setup and we are using Travis CI but have not completed migrations yet.
Take what's left in `noxfile.py` and move it into `tox.ini`.

bu3alwa:
There doesn't seem to be much to port over, looks like most of it is there. There was only one line to add. Am I missing something? Also, there is a todo deployment in  `tox.ini` is there any requirements for that?

[...]

PR #579, opened 2019-03-26
Update default python version to 3.7

c24t:
Add python 3.7 to supported versions, run unit and system tests against 3.7, and update our default CI target version to 3.7.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
       ignore:
         - gh-pages
     docker:
-      - image: googleapis/nox:0.17.0
+      - image: googleapis/python-multi:latest
       - image: mysql:5.7
         environment:
           MYSQL_ROOT_HOST: "%"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-  - '3.6'
+  - '3.7'
 
 install:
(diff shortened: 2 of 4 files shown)


reyang:
@zoidbergwill FYI

c24t:
The default travis base image doesn't support 3.7, so we need to explicitly require ubuntu 16.04. See https://github.com/travis-ci/travis-ci/issues/9069 and https://github.com/travis-ci/travis-ci/issues/9815.

c24t:
Our old circle base image ([googleapis/nox](https://hub.docker.com/r/googleapis/nox/)) doesn't support 3.7. I've changed this to [googleapis/python-multi](https://hub.docker.com/r/googleapis/python-multi), but we may want to update the nox image instead.

c24t:
The new circle image doesn't include openssl, looks like we need to update the old image or start using one of our own.

zoidbergwill:
Not sure where the source is for the nox image, but we should probably update it to a newer version of nox too. Seems like some things have changed when fiddling with it for my PR

reyang:
@lzchen is working on #402.

reyang:
Python 3.7 support has been added by #665.
Although we are still in progress of retiring Circle CI, we already got Python 3.7 test covered by Travis CI.

Issue #318, opened 2018-09-19
Move proto-generated files to OpenCensus-Proto.


songy23:
I'm new to the dependency management in Python so this may be a naive question:

For the OC-Agent Exporter, we put the proto-generated Python files as a vendor under [opencensus/trace/exporters/gen/opencensus](https://github.com/census-instrumentation/opencensus-python/tree/master/opencensus/trace/exporters/gen/opencensus). AFAIS this is not convenient since every time when we made a new release on [OpenCensus-Proto](https://github.com/census-instrumentation/opencensus-proto), those gen-files need to be updated separately. This has also caused some trouble with Pylint before (https://github.com/census-instrumentation/opencensus-python/pull/266#issuecomment-418474228).

If possible, I think a better way is to host the generated Python files on [OpenCensus-Proto](https://github.com/census-instrumentation/opencensus-proto), release them as separate Python library, and have the OC-Agent Exporter import it. Note that Go [already did it](https://github.com/census-instrumentation/opencensus-proto/tree/master/gen-go).

/cc @lmolkova @liyanhui1228

liyanhui1228:
That sounds good to me. Stackdriver Trace exporter also uses the same strategy, we host the auto-gen stackdriver trace Python files in [google-cloud-python/trace](https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/trace), and release it to PyPI as google-cloud-trace library. Then it is used in stackdriver trace exporter.

PR #589, opened 2019-03-29
Regenerate ocagent proto files.

songy23:
Regenerate gen-proto files with the latest OpenCensus-proto, similar to #524. Everything is auto-generated by proto compiler except a few fixes in the import paths.

/cc @lmolkova


songy23:
Ideally the proto generated files should be hosted at https://github.com/census-instrumentation/opencensus-proto/pull/196 and remove `opencensus-ext-ocagent/opencensus/ext/ocagent/trace_exporter/gen` (tracked in https://github.com/census-instrumentation/opencensus-python/issues/318). Once that's done we won't need to have PRs like this to update the gen files in this repo.

reyang:
> Ideally the proto generated files should be hosted at census-instrumentation/opencensus-proto#196 and we can remove opencensus-ext-ocagent/opencensus/ext/ocagent/trace_exporter/gen (tracked in #318). Once that's done we won't need to have PRs like this to update the gen files in this repo.

I agree with you! I actually suggested to @c24t that we should have the generated stuff as a standalone Python package, and release it automatically each time when we update the `opencensus-proto`. In this way if we want to take the new protocol, we can just update the package reference from the exporter.

@songy23 @c24t what's your thinking?

songy23:
> I agree with you! I actually suggested to @c24t that we should have the generated stuff as a standalone Python package, and release it automatically each time when we update the opencensus-proto. In this way if we want to take the new protocol, we can just update the package reference from the exporter.

Yes, I'm currently working on it: https://github.com/census-instrumentation/opencensus-proto/pull/196. Once I have the release set up in opencensus-proto, I'll come back and remove the gen-files in the Python repo.

PR #593, opened 2019-04-02
Stackdriver poll-based metrics

### LINKED ISSUE #470: Don't expose mutable view data
`MeasureToViewMap` exposes mutable `ViewData`s to exporters. Consider exporting a immutable view of `ViewData` instead. 

See https://github.com/census-instrumentation/opencensus-python/pull/469#discussion_r251619805 for some context.

### LINKED ISSUE #583: 400 Errors From Stackdriver Stats Exporter
Running the [opencensus-ext-stackdriver example](https://github.com/census-instrumentation/opencensus-python/blob/49e4cb30414295f04980f96a565e9f234c96cce6/contrib/opencensus-ext-stackdriver/examples/stackdriver.py) results in errors when writing timeseries to Stackdriver. I was able to reproduce this in a fresh venv even when recording only two metrics.

Environment:
```
Python 3.6
$ pip freeze
cachetools==3.1.0
certifi==2019.3.9
chardet==3.0.4
google-api-core==1.8.2
google-auth==1.6.3
google-cloud-core==0.29.1
google-cloud-monitoring==0.31.1
google-cloud-trace==0.20.2
googleapis-common-protos==1.5.9
grpcio==1.19.0
idna==2.8
opencensus==0.3.1
opencensus-ext-stackdriver==0.1.1
protobuf==3.7.1
pyasn1==0.4.5
pyasn1-modules==0.2.4
pytz==2018.9
requests==2.21.0
rsa==4.0
six==1.12.0
urllib3==1.24.1
```

```
  File ".venv/lib/python3.6/site-packages/opencensus/common/transports/async_.py", line 105, in _thread_main
    self.exporter.emit(data)
  File ".venv/lib/python3.6/site-packages/opencensus/ext/stackdriver/stats_exporter/__init__.py", line 153, in emit
    self.handle_upload(view_data)
  File ".venv/lib/python3.6/site-packages/opencensus/ext/stackdriver/stats_exporter/__init__.py", line 165, in handle_upload
    self.upload_stats(view_data)
  File ".venv/lib/python3.6/site-packages/opencensus/ext/stackdriver/stats_exporter/__init__.py", line 177, in upload_stats
    time_series_batch)
  File ".venv/lib/python3.6/site-packages/google/cloud/monitoring_v3/gapic/metric_service_client.py", line 897, in create_time_series
    request, retry=retry, timeout=timeout, metadata=metadata
  File ".venv/lib/python3.6/site-packages/google/api_core/gapic_v1/method.py", line 143, in __call__
    return wrapped_func(*args, **kwargs)
  File ".venv/lib/python3.6/site-packages/google/api_core/retry.py", line 270, in retry_wrapped_func
    on_error=on_error,
  File ".venv/lib/python3.6/site-packages/google/api_core/retry.py", line 179, in retry_target
    return target()
  File ".venv/lib/python3.6/site-packages/google/api_core/timeout.py", line 214, in func_with_timeout
    return func(*args, **kwargs)
  File ".venv/lib/python3.6/site-packages/google/api_core/grpc_helpers.py", line 59, in error_remapped_callable
    six.raise_from(exceptions.from_grpc_error(exc), exc)
  File "<string>", line 3, in raise_from
google.api_core.exceptions.InvalidArgument: 400 One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/task_latency_distribution, Timestamps: {Youngest Existing: '2019/03/27-10:27:48.522', New: '2019/03/27-10:27:49.523'}}: timeSeries[0]
```

Do you have any advice on how to proceed? Thanks!

c24t:
This PR changes `StackdriverStatsExporter` to use metrics data models internally, and switches from a push-based export model (export on each call to `record`) to a poll-based one (periodically call `stats.get_metrics` and export the results). It also allows the stackdriver exporter to export gauge metrics. It includes the changes from #560 and #577.

This PR includes multiple breaking behavior and API changes, including starting (and now returning) a background thread from `stats_exporter.new_stats_exporter`, and exposing `Stats` as a singleton via `stats.stats`.

Fixes #583 by exporting less frequently than the stackdriver minimum sampling period.

Fixes #470 by replacing `ViewData` with metrics objects that are created on the fly.

Addresses #454, which we can close after making the [same changes to the prometheus exporter](https://github.com/census-instrumentation/opencensus-python/issues/594).


contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -110,18 +130,14 @@ def default_monitoring_labels(self):
...
-class StackdriverStatsExporter(base_exporter.StatsExporter):
+class StackdriverStatsExporter(object):
     """Stats exporter for the Stackdriver Monitoring backend."""
 
-    def __init__(self,
-                 options=Options(),
-                 client=None,
-                 default_labels={},
-                 transport=async_.AsyncTransport):
+    def __init__(self, options=Options(), client=None):
```
reyang: If people try something like `xyz = StackdriverStatsExporter()` and `xyz.options.project_id = 'foobar'`, this would overwrite the default argument. Do we want to prevent this case?

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -402,21 +363,34 @@ def get_user_agent_slug():
     return "opencensus-python/{}".format(__version__)
```
reyang: Curious (although it is not added by this PR) - which version do we want here (opencensus or ext package or both)?

opencensus/metrics/transport.py
```diff
@@ -0,0 +1,100 @@
...
+
+    :type interval: int or float
+    :param interval: Seconds between calls to the function.
+    """
+
+    daemon = True
+
+    def __init__(self, func, interval=None, **kwargs):
+        super(PeriodicTask, self).__init__(**kwargs)
+        if interval is None:
```
reyang: Curious, what's the difference with `def __init__(self, func, interval=DEFAULT_INTERVAL, **kwargs)`?

opencensus/metrics/transport.py
```diff
@@ -0,0 +1,100 @@
...
+
+DEFAULT_INTERVAL = 60
+GRACE_PERIOD = 5
+
+
+class TransportError(Exception):
+    pass
+
+
+class PeriodicTask(threading.Thread):
```
reyang: This could be useful for other exporters (e.g. trace exporters), please consider putting it in `opencensus.common`.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: That seems fine to me, any reason to prevent it?

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: The other clients report the core library version, but I think there's a good argument for replacing this with just the ext version.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
reyang: For example:
x = StackdriverStatsExporter()
x.options.project_id = 'foobar'
y = StackdriverStatsExporter()
# here y.options.project_id was polluted with 'foobar'

opencensus/metrics/transport.py
c24t: Just that it makes it easy to make the same argument optional in the calling func.

`interval` comes all the way from `stackdriver.stats_exporter.new_stats_exporter`, if we used `DEFAULT_INTERVAL` here we'd have to either have to use it as the default value in the callers (which means importing it in other modules) or make two separate calls, e.g.

```python
def caller_in_another_module(interval=None):
    ...
    if interval is None:
        task = PeriodicTask()
    else:
        task = PeriodicTask(interval=interval)
```

Using `None` everywhere is convenient, but there are a few big drawbacks: it's less readable since you can't tell the defaults by looking at the signature, you can't change the default value to something other than `None` without changing all the callers too, and it doesn't work for arguments that can actually have `None` as their value.

I'm on the fence here, if you have an opinion on this I'm happy to change it.

opencensus/metrics/transport.py
c24t: For trace exporters we want a queue that periodically exports batches of spans. It's possible that we'd reuse this, but I think we ought to move it when we do instead of now.

opencensus/metrics/transport.py
reyang: I don't have strong opinion here. Just observed that using None is a common pattern: https://docs.python-guide.org/writing/gotchas/. Feel free to make your call.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
c24t: Oh I see, yeah this shouldn't be mutable.

c24t: Fixed in 0dc919f.

opencensus/metrics/transport.py
reyang: Make sense.

PR #596, opened 2019-04-02
Remove vendor proto files and use opencensus-proto instead.

### LINKED ISSUE #318: Move proto-generated files to OpenCensus-Proto.

songy23:
~~DO NOT MERGE until https://github.com/census-instrumentation/opencensus-proto/pull/197 is merged and a new release is cut.~~

Fixes #318.


contrib/opencensus-ext-ocagent/setup.py
```diff
@@ -41,6 +41,7 @@
     install_requires=[
         'grpcio >= 1.0.0, < 2.0.0',
         'opencensus >= 0.4.dev0, < 1.0.0',
+        'opencensus-proto >= 0.0.dev0, < 1.0.0',
```
c24t: FYI we've been replacing "dev" in the release branches. If `opencensus-proto` follows the same versioning strategy the first release will be `0.1.0` and the version number on master after the release should be `0.2.dev0`, in which case this should be `0.1.dev0`.

reyang:
@songy23 just realized there was a lint rule which we can get rid off. You can leave it to me since we'll move from flake8 to pylint, up to you.

https://github.com/census-instrumentation/opencensus-python/blob/master/nox.py#L119

songy23:
Got it, I'll leave it to you then because I'm putting this PR on hold until we made the next release for opencensus-proto (currently it's 0.0.dev0).

songy23:
https://github.com/census-instrumentation/opencensus-proto/pull/197 is merged, now I think we can move forward with this PR.

PR #598, opened 2019-04-02
Fix closure binding bug

reyang:
Python works differently than Node.js, the function defined inside another function is not a real closure.

The previous version of `with_current_context` is buggy since the local variable used to hold the `caller_context` could be overwritten and cause side-effects.

I will add coverage test and add documentation in another PR.

### DIFF
--- a/context/opencensus-context/opencensus/common/runtime_context/__init__.py
+++ b/context/opencensus-context/opencensus/common/runtime_context/__init__.py
@@ -68,20 +68,24 @@ class _RuntimeContext(object):
         slot = self._slots[name]
         slot.set(value)
 
+    class Closure(object):
+        def __init__(self, context, func):
+            self.context = context
+            self.snapshot = context.snapshot()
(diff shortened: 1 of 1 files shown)


reyang:
@c24t turned out I was wrong.

This article explains how late binding works https://docs.python-guide.org/writing/gotchas/#late-binding-closures. This is not affecting the current implementation (as we are not mutating the local variable).

I confirmed that the current implementation is correct by examine the closure:

```
>>> from opencensus.common.runtime_context import RuntimeContext
>>> RuntimeContext.with_current_context(lambda: None).__closure__
(<cell at 0x0000022E9C2D3CA8: dict object at 0x0000022E9C39BC60>,
<cell at 0x0000022E9C2D3F18: function object at 0x0000022E9A37C268>,
<cell at 0x0000022E9C32B108: _AsyncRuntimeContext object at 0x0000022E9C340C50>)
```

Abandon this PR.

PR #602, opened 2019-04-04
Minor refactor Tags

reyang:
As I'm working on [TagMap](https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md) now, trying to do some clean up before adding features:
1. Instead of having to use Tag(TagKey('foo'), TagValue('value')), use Tag('foo', 'value').
2. Clean up `tags.execution_context`.
3. Fixing CI error introduced by PostgreSQL lib deprecation on old version of Python runtime.

@songy23 @c24t

PR #603, opened 2019-04-04
Prevent middleware error on older Django versions (<1.8)

wsabransky:


reyang:
@zoidbergwill FYI

reyang:
@wsabransky does this mean we can support much lower version at https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-django/setup.py#L42?

wsabransky:
@reyang Yep, I've tested this on 1.6.5 and it runs.

wsabransky:
Hmm, it seems like unit tests are failing for reasons potentially not related to this PR. Maybe due to this being moved into its own package?

reyang:
@wsabransky Great! Probably we can change the required Django version in `setup.py`. What would be your suggestion on the lowest version?

reyang:
The CI failure was introduced by the PostgreSQL lib update.
I have a fix in #602.

wsabransky:
@reyang I don't know enough about Django before 1.6.x to say with confidence that it would work. But 1.6.x is definitely safe.

reyang:
@wsabransky Great to know. Thank you!

reyang:
@wsabransky I have the CI fixed, would you rebase to the latest master and try?

wsabransky:
@reyang done

contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
```diff
@@ -99,7 +99,11 @@ def _set_django_attributes(span, request):
         return
 
     user_id = django_user.pk
-    user_name = django_user.get_username()
+    try:
+        user_name = django_user.get_username()
+    except Exception:
```
zoidbergwill: I think it'd be better checking the Django version or catching a more specific error then a general exception?

zoidbergwill:
@reyang Thanks for the ping, when we add support for testing multiple Django versions, we can support all the currently maintained versions then probably, which is great.

c24t:
FWIW we've been setting the minimum django version according to [django's support timeline](https://docs.djangoproject.com/en/2.2/internals/release-process/#supported-versions) even though the opencensus middleware is technically compatible with older versions. We do the same thing for flask.

Since we only depend on django to instrument it (and aren't e.g. running a django service ourselves), most of the usual security concerns don't apply. On the other hand, it makes our lives easier if we always set the minimum required versions of our dependencies such that it's impossible to accidentally install an old unsupported version.

contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
wsabransky: Good point. Changed this

wsabransky:
@c24t Makes sense. I've reverted the supported minimum version change. Anything else you'd like to see to get this merged?

c24t:
Looks like lint and coverage checks are still failing.

wsabransky:
@c24t Any chance we could get this merged into a minor version update for the django extension?

c24t:
@wsabransky in case you haven't seen it already: this was released in https://github.com/census-instrumentation/opencensus-python/releases/tag/v0.5.0.

Issue #544, opened 2019-03-06
Improve package installation


reyang:
Ideally we want to:
1. If user installs OpenCensus from PyPI, they should get packages from PyPI.
2. If user installs OpenCensus from [GitHub repo](/census-instrumentation/opencensus-python) master branch, they should get packages from the latest version of master branch.
3. If OpenCensus developers install `opencensus` from a local clone (with local modifications), they should get packages from the local folder (so they can test them properly).

Some context on how pip works:
https://github.com/pypa/pip/pull/6060
https://pip.pypa.io/en/stable/news/#deprecations-and-removals

This is part of #445 effort.

reyang:
@c24t I wonder if having development release (or nightly build) is a good idea.

https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/specification.html#development-releases

If pip could support installing package + dependencies from GitHub (without hack like #549), it'll be even better.

PR #604, opened 2019-04-05
Prepare for 0.4.0 release

### LINKED ISSUE #597: Release stackdriver stats and context changes
We've got two large breaking unreleased changes on master: making the stackdriver stats exporter consume stats as metrics (#593) and adding a context API that replaces threadlocals with contextvars on recent python versions (#588).

We'll have to release versions `0.4.0` of the core library, `0.3.0` of `opencensus-ext-stackdriver`, and `0.1.0` of `opencensus-context`. We can continue to release ext packages from the same release branch as the main library (in this case the new `v0.4.x` branch), or start using separate release branches for each package.

cc @reyang

c24t:
Update the version numbers, setup files, and changelogs for the `0.4.0` core library release and multiple ext package releases.

Ext packages with insignificant changes, micro version updates:
- `opencensus-ext-dbapi`
- `opencensus-ext-google-cloud-clientlibs`
- `opencensus-ext-grpc`
- `opencensus-ext-httplib`
- `opencensus-ext-mysql`
- `opencensus-ext-postgresql`
- `opencensus-ext-pymongo`
- `opencensus-ext-pymysql`
- `opencensus-ext-requests`
- `opencensus-ext-sqlalchemy`
- `opencensus-ext-threading`

Ext packages affected by package changes in the core library, minor version updates:
- `opencensus-ext-django`
- `opencensus-ext-flask`
- `opencensus-ext-jaeger`
- `opencensus-ext-pyramid`
- `opencensus-ext-prometheus`
- `opencensus-ext-zipkin`

Ext packages with significant changes, also minor version updates:
- `opencensus-ext-ocagent`
- `opencensus-ext-stackdriver`

Fixes #597.


contrib/opencensus-correlation/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.2.dev0'
+__version__ = '0.1.0'
```
c24t: @reyang do you want to keep releasing this package?

contrib/opencensus-ext-google-cloud-clientlibs/setup.py
```diff
@@ -39,9 +39,9 @@
     include_package_data=True,
     long_description=open('README.rst').read(),
     install_requires=[
-        'opencensus >= 0.4.dev0, < 1.0.0',
-        'opencensus-ext-grpc >= 0.2.dev0, < 1.0.0',
-        'opencensus-ext-requests >= 0.2.dev0, < 1.0.0',
+        'opencensus >= 0.3.0, < 1.0.0',
```
c24t: For ext packages that don't depend on changes in the core library I didn't change the required opencensus version. This could get confusing quickly since this requirement will be `0.5.dev0` on master to allow us to keep making breaking changes on the dev version.

We may want to keep these version numbers in sync across all the ext packages in the interest of our sanity, at the cost of backwards compatibility.

contrib/opencensus-correlation/version.py
reyang: Up to you (depending on which way is easier for you). We don't have dependency from opencensus and there is no actual functionality, no need to release it.

Once we heard back from @bogdandrutu, we can decide whether to put the correlation context work in this package, or remove it (and do the implementation in opencensus core lib).

c24t: OK, I won't touch it in this branch and it'll be released again as the same version.

c24t:
CI will fail for PRs like this that update the core library version and ext requirements to depend on that version in the same change. We may fix this as part of #544.

reyang:
> CI will fail for PRs like this that update the core library version and ext requirements to depend on that version in the same change. We may fix this as part of #544.

I think this is because we haven't updated `opencensus/common/version/__init__.py`?

c24t:
That'd do it, updated in b0fb1dd.

c24t:
Rebased to include 1689f98 and 294bc2c.

PR #611, opened 2019-04-09
Allow empty `MetricDescriptor.label_keys`

samn:
so metrics without tags can be used. A `View`'s columns are used as the
`MetricDescriptor`'s `label_keys`, and the
[docs](https://opencensus.io/stats/view/) say that they're optional.

This change preserves the behavior when the `label_keys` are `None`, or if
any values are `None`.

This matches the behavior of the [Java
API](https://github.com/census-instrumentation/opencensus-java/blob/08d09c48a40b7dc6b5dba8603795084fc6ea9695/api/src/main/java/io/opencensus/metrics/export/MetricDescriptor.java#L53-L54).


c24t:
@samn if you sign the CLA we can merge this.

PR #617, opened 2019-04-12
Add ocagent stats exporter.

prateekr:


contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/proto/__init__.py
```diff
@@ -0,0 +1 @@
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```
reyang: FYI - proto is being moved out #596.

prateekr:
Ah I was worried it wouldn't be ready at this point; will go ahead and update.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/utils/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+from google.protobuf.internal.well_known_types import ParseError
+from google.protobuf.timestamp_pb2 import Timestamp
+from opencensus.common.version import __version__
+from opencensus.proto.agent.common.v1 import common_pb2
+
+# Default agent endpoint
+DEFAULT_ENDPOINT = 'localhost:55678'
+
+# OCAgent exporter version
+EXPORTER_VERSION = '0.0.1'
```
reyang: The exporter package comes with a version `https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-ocagent/version.py`.
Probably consider using the same version instead of creating a new one (which is hard to maintain).

For example:
https://github.com/census-instrumentation/opencensus-python/pull/613/files#diff-cd21ef3212e9029163fd1e55142d543d

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/utils/__init__.py
```diff
@@ -0,0 +1,65 @@
+import datetime
+import os
+import socket
+
+from google.protobuf.internal.well_known_types import ParseError
+from google.protobuf.timestamp_pb2 import Timestamp
+from opencensus.common.version import __version__
```
reyang: Minor suggestion, give it an explicit name (e.g. opencensus_version) so it won't be confused with extension version.

reyang:
Please also update the README and CHANGELOG, thanks!

contrib/opencensus-ext-ocagent/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2018, OpenCensus Authors
```
reyang: One caveat about Python unittest, the base file name should be unique.
Consider something like "test_ocagent_utils.py".

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/proto/__init__.py
prateekr: Done

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/utils/__init__.py
prateekr: As version.py is currently located in `contrib/opencensus-ext-ocagent/version.py` rather than contrib/opencensus-ext-ocagent/opencensus/ext/ocagent`, it makes it difficult to import it as the import path itself stats at the opencensus dir.

I imagine this is why the existing trace exporter also defines the version directly: 
https://github.com/census-instrumentation/opencensus-python/blob/3b60af770da1a90e751cb94cefcc749eb24a5768/contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/trace_exporter/__init__.py#L34

I'd rather leave this out of scope of this PR and patch them together in a followup.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/utils/__init__.py
prateekr: Done

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/utils/__init__.py
reyang: Makes sense.

@c24t for awareness.

contrib/opencensus-ext-ocagent/tests/test_utils.py
prateekr: Done

prateekr:
One of the tests globally modifies channels created through grpc.insecure to always attach OpenCensusClientInterceptor.  This is causing test_stats_exporter to pull in this interceptor as well.

The grpc.StreamStreamClientInterceptor abstract class states stream interceptor method should return an object that's both a call (implementing the response iterator) and a future. 
 https://github.com/grpc/grpc/blob/8054a731d1486e439e6becb1987b1e97246e6476/src/python/grpcio/grpc/__init__.py#L529-L561

However, OpenCensusClientInterceptor just returns a python generator, thus not respecting the defined interface.  This in turn breaks all users (including api_core utils) that rely on `Future` and `Call` functionality.

https://github.com/census-instrumentation/opencensus-python/blob/c59a087a7c1f4c495b39d56fbdf3daebb421ffc9/contrib/opencensus-ext-grpc/opencensus/ext/grpc/client_interceptor.py#L228-L238

I've created a wrapper iterator that I now return here to ensure the current use cases work.  In a followup, it'd be good to pull out the `NotImplemented` methods in their entirety.
@reyang Could you take another pass at these changes?

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+from opencensus.metrics.export import metric_descriptor
+from opencensus.metrics.export import value
+from opencensus.proto.agent.metrics.v1 import metrics_service_pb2
+from opencensus.proto.agent.metrics.v1 import metrics_service_pb2_grpc
+from opencensus.proto.metrics.v1 import metrics_pb2
+from opencensus.stats import stats
+import grpc
+
+
+class StatsExporter(object):
```
c24t: Reminder that we'll need to rename these `MetricsExporter` eventually.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+def _create_stub(endpoint):
+    return metrics_service_pb2_grpc.MetricsServiceStub(
+        grpc.insecure_channel(endpoint))
+
+
+def new_stats_exporter(service_name,
+                       hostname=None,
+                       endpoint=None,
+                       interval=None):
+    """Create a ocagent exporter and attaches it to the exporter thread.
```
c24t: s/attaches/attach, but also note that this creates a new exporter thread now.

c24t:
FYI the coverage check is failing for WrappedResponseIterator:

`contrib/opencensus-ext-grpc/opencensus/ext/grpc/utils.py`
`75%   45-47, 93, 96, 99, 102, 105, 108, 111, 114, 117, 120, 126, 129, 132`

`contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py`
`98%   91, 79->91`

I don't see any problem with adding `# pragma: NO COVER` to the intentionally unimplemented methods.

contrib/opencensus-ext-grpc/tests/test_client_interceptor.py
```diff
@@ -12,13 +12,18 @@
...
+import threading
+import unittest
 
+from google.api_core import bidi
+from google.protobuf import proto_builder
+from grpc.framework.foundation import logging_pool
 from opencensus.ext.grpc import client_interceptor
 from opencensus.trace import execution_context
 from opencensus.trace.tracers.noop_tracer import NoopTracer
+import grpc
```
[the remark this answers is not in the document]
prateekr: Done.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+from opencensus.proto.metrics.v1 import metrics_pb2
+from opencensus.stats import stats
+import grpc
+
+
+class StatsExporter(object):
+    """ Stats exporter for an opencensus metrics grpc service.
+
+    :type rpc_handler: ExportRpcHandler
+    :param client: export rpc handler
```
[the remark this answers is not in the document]
prateekr: Fixed.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+from opencensus.metrics.export import value
+from opencensus.proto.agent.metrics.v1 import metrics_service_pb2
+from opencensus.proto.agent.metrics.v1 import metrics_service_pb2_grpc
+from opencensus.proto.metrics.v1 import metrics_pb2
+from opencensus.stats import stats
+import grpc
+
+
+class StatsExporter(object):
+    """ Stats exporter for an opencensus metrics grpc service.
```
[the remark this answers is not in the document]
prateekr: Woops, fixed.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
prateekr: Updated to note thread creation.

contrib/opencensus-ext-ocagent/opencensus/ext/ocagent/stats_exporter/__init__.py
prateekr: Ack.

prateekr:
> What are the odds that this integration is going to break with an update to grpc seeing as this is an experimental API?

Not sure, but the grpc interceptor has been around for over two years now, so I imagine it's maturing.  Also, I don't think the rpc stream return object class is experimental, so any potential changes will likely continue to return this.

reyang:
@prateekr sorry for my delayed response. The code looks good to me.

I have a general question @c24t, do we want to have a special context variable for exporters, so we don't intercept any operation (besides calculating some exporter specific metrics) while running exporter logic.

prateekr:
> Even if your class fixes test_stats_exporter it sounds like we need to stop the other test from patching this globally!

Agreed, looks like contrib/opencensus-ext-google-cloud-clientlibs/tests/test_google_cloud_clientlibs_trace.py is an offender.  I've patched it enough so the tests in this PR pass.  Leaving a wholistic cleanup of this test to a future CL.

c24t:
> do we want to have a special context variable for exporters, so we don't intercept any operation (besides calculating some exporter specific metrics) while running exporter logic

So you imagine checking for a context var in the integrations and -- if it's set -- passing calls through to the wrapped library (e.g. `grpcio`, `requests`) instead of intercepting them to attach trace data/collect metrics?

If it's possible I'd prefer not to monkey patch libraries like this at all, but I understand one of the reasons to do this (and why other projects/libraries do bytecode manipulation) is to instrument e.g. grpc calls even when they're buried inside another library. I don't have any suggestions to address this that aren't colossal hacks. 

At least in this specific case the problem is that a test isn't cleaning up after itself. Even if the integration acted as a no-op during the test it would still be a problem that the integration was loaded at all.

c24t:
@prateekr once coverage is fixed this looks good to merge.

reyang:
> So you imagine checking for a context var in the integrations and -- if it's set -- passing calls through to the wrapped library (e.g. `grpcio`, `requests`) instead of intercepting them to attach trace data/collect metrics?

Yes, at least we don't want to collect traces. We might need to collect metrics in some cases though.
Today we don't have a contract, so exporters ended up doing [hacks](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py#L117), imagine we have something MySQL exporter or PostgreSQL exporter, we will need to invent another hack. It'll be much cleaner if we have a flag to indicate whether the code is running inside an exporter context, so integrations and exporters can be decoupled through that contract.
 
> If it's possible I'd prefer not to monkey patch libraries like this at all, but I understand one of the reasons to do this (and why other projects/libraries do bytecode manipulation) is to instrument e.g. grpc calls even when they're buried inside another library. I don't have any suggestions to address this that aren't colossal hacks.

Me too! I think the long term direction is encourage 3rd party framework/libraries adding these probes, rather than having OpenCensus reverse engineering and hacking in the probes.
It will take time as you said, for now we will need to rely on community contribution to maintain such a big matrix of hacks.

> At least in this specific case the problem is that a test isn't cleaning up after itself. Even if the integration acted as a no-op during the test it would still be a problem that the integration was loaded at all.

Yup, this question is not specific to this PR. @prateekr's approach looks good to me within the PR context, and I've signed off.

c24t:
> It'll be much cleaner if we have a flag to indicate whether the code is running inside an exporter context, so integrations and exporters can be decoupled through that contract.

I think that's a good idea, and if we do implement a general purpose kill-switch for integrations it's probably something that we want in more than just the python client.

Issue #378, opened 2018-11-01
Probability sampler is broken for low sample rates


akoskaaa:
Hello everyone,

I've found an issue with the probability sampler not working for low sample rates. There are 2 piecens of code related to this:
https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/samplers/probability.py#L56-L66
As the code says, it fetches the last 8 bytes of the trace id, that will later be used to compare to a constant value calculated using the sample rate. This algorithm is, in fact, flawless and propagates traces perfectly.
The issue comes from a bit deeper:
https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span_context.py#L165-L173
As a matter of fact, `uuid4` is not an entirely random number according to the uuid specs.
Here's some empyrical data as proof:
https://gist.github.com/akoskaaa/a53c8b0201a8bffc1ae4df91e5965933

Source:
https://tools.ietf.org/html/rfc4122
https://en.wikipedia.org/wiki/Universally_unique_identifier

Solution could be to change the tracing id generation to a more random one. I've actually patched it in my systems with this function:
```py
def patched_generate_trace_id():
    hexdigits = '0123456789abcdef'
    return ''.join([hexdigits[random.randint(0, 0xF)] for _ in range(32)])
```
Do you have any better ideas?

Also, I'd just like to say thanks to the owners and everyone who put time into making this happen. Super useful!

reyang:
Should be fixed by #619, big thank you to @ewhauser!

akoskaaa:
Huge thanks for the fix!

PR #619, opened 2019-04-14
Change algorithm for span ID generation

ewhauser:
The span ID is currently generated using `uuid` which calls `/dev/urandom`. This is slow and can cause problems under load. Generating using non-cryptographic random is ~10x faster:

```console
python -m timeit "import uuid; uuid.uuid4().hex[:16]"
50000 loops, best of 5: 4.54 usec per loop
```

```console
> python -m timeit "import random; '{:016x}'.format(random.getrandbits(64))"
500000 loops, best of 5: 657 nsec per loop
```

The code used here to generate the ID is from [Zipkin's Python library](https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/util.py#L7). It meets the standard as defined under the [OpenCensus specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#traceid).

### DIFF
--- a/opencensus/trace/span_context.py
+++ b/opencensus/trace/span_context.py
@@ -17,7 +17,8 @@
 import logging
 import re
 import six
-import uuid
+import random
+import time
 
 from opencensus.trace import trace_options as trace_options_module
 
@@ -156,8 +157,7 @@ def generate_span_id():
     :rtype: str
     :returns: 16 digit randomly generated hex trace id.
     """
-    span_id = uuid.uuid4().hex[:16]
-    return span_id
+    return '{:016x}'.format(random.getrandbits(64))
 
 
 def generate_trace_id():
@@ -167,5 +167,6 @@ def generate_trace_id():
     :rtype: str
     :returns: 32 digit randomly generated hex trace id.
     """
-    trace_id = uuid.uuid4().hex
-    return trace_id
+    t = int(time.time())
+    lower_96 = random.getrandbits(96)
+    return '{:032x}'.format((t << 96) | lower_96)


ewhauser:
I signed it

reyang:
Thanks @ewhauser, this approach could potentially address #378 as well.

opencensus/trace/span_context.py
```diff
@@ -167,5 +167,6 @@ def generate_trace_id():
     :rtype: str
     :returns: 32 digit randomly generated hex trace id.
     """
-    trace_id = uuid.uuid4().hex
-    return trace_id
+    t = int(time.time())
+    lower_96 = random.getrandbits(96)
+    return '{:032x}'.format((t << 96) | lower_96)
```
reyang: Curious, why would we want to encode the time in the high bits? There could be side effect if sampler is using the high bits.
@bogdandrutu @songy23 Do we have a cross language guideline on the randomness of trace_id and span_id? (and how samplers should do)

opencensus/trace/span_context.py
```diff
@@ -156,8 +157,7 @@ def generate_span_id():
     :rtype: str
     :returns: 16 digit randomly generated hex trace id.
     """
-    span_id = uuid.uuid4().hex[:16]
-    return span_id
+    return '{:016x}'.format(random.getrandbits(64))
```
reyang: What if the user makes explicit call to random.seed? Will this generate the same sequence of ids under the same seed?
If yes, potentially we want a separate random seed / generator (which is separate from user code).

opencensus/trace/span_context.py
ewhauser: Zipkin does this for interop with Amazon X-Ray. See the notes here: https://github.com/apache/incubator-zipkin/issues/1754

opencensus/trace/span_context.py
ewhauser: `random.seed()` effectively modifies the global state of your program's use of `random`. If a user is intentionally setting the `seed`, then isn't the expected behavior that the characteristics of the generator would change? Could certainly isolate it.

c24t:
There's an issue in the specs repo for this: https://github.com/census-instrumentation/opencensus-specs/issues/64. We can't change the ID generation strategy in this client unilaterally, but we will need to give users the option to generate x-ray-compatible IDs when we add an x-ray exporter.

Replacing `uuid` with `random.getrandbits` seems like an obvious win though.

c24t:
Sorry @reyang, I meant that LGTM to be conditional on removing the x-ray specific logic. We can't do that in this client alone.

PR #620, opened 2019-04-16
Don't use time in trace ID creation

c24t:
As per https://github.com/census-instrumentation/opencensus-python/pull/619#issuecomment-483438349, unless the other client libraries do the same we don't want to use the current time in trace ID creation.


c24t:
No problem at all, my fault on the early approval.

Issue #394, opened 2018-11-09
Decouple exporter logic from framework integration


reyang:
As a follow up to #383.

Currently we have hard coded exporter endpoints in Flask/Django ext, need to refactor them.

reyang:
@c24t @songy23 here goes how I plan to solve this problem:
1. Introduce a class `Namespace` for dynamic import/evaluation under 'opencensus' namespace.
2. Refactor all existing configurations (which is going to be a break change).

```
import importlib

class Namespace(object):
    def __init__(self, name, parent=None):
        self.parent = parent
        self.name = name

    def __getattr__(self, name):
        return type(self)(name, self)

    def __str__(self):
        if self.parent is None:
            return self.name
        return '{!s}.{}'.format(self.parent, self.name)

    def __call__(self, *args, **kwargs):
        ctor = getattr(importlib.import_module(str(self.parent)), self.name)
        return ctor(*args, **kwargs)

    @classmethod
    def eval(cls, expr):
        return eval(expr, {}, {'opencensus': cls('opencensus')})


print(Namespace.eval("opencensus.ext.azure.trace_exporter.AzureExporter(options=opencensus.ext.azure.common.Options({}))"))
```

reyang:
Once the refactor work for Flask/Django/Pyramid are all done, need to do the following things:

1. Update documentation.
1.1. Fix the TBD docs.
1.2. Update the main README file.
2. Prepare release note on the breaking changes and how to migrate existing projects.
2.1. Configuration structure change.
2.2. Default value change for exporters.
2.3. Default propagator changed to W3C TraceContext.

reyang:
@c24t need your help on 2.2 and 2.3. Thanks!

PR #621, opened 2019-04-16
Decouple exporters from Flask integration

reyang:
This is part of the effort for #394.
1. Introduced a generic configuration evaluation class.
2. Decouple Flask integration and extension exporters.
3. Use W3C TraceContext as the default propagator. Following https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#propagation

_If user doesn't set any propagation methods explicitly, TraceContext is used._

@c24 @songy23 please take a brief look and see if you're okay with the direction.
Once I got okay from you on the direction, I'll work on Django and Pyramid as well.

### DIFF
--- a/contrib/opencensus-ext-azure/examples/custom.py
+++ b/contrib/opencensus-ext-azure/examples/custom.py
@@ -19,13 +19,15 @@ app = Flask(__name__)
 # TODO:
 # Replace 00000000-0000-0000-0000-000000000000 with your Instrumentation Key
 # https://docs.microsoft.com/en-us/azure/azure-monitor/app/create-new-resource
-app.config['OPENCENSUS_TRACE'] = {
-    'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
-    'EXPORTER': '''opencensus.ext.azure.trace_exporter.AzureExporter(
-        opencensus.ext.azure.common.Options(
-            instrumentation_key="00000000-0000-0000-0000-000000000000",
-            timeout=29.9,
-        ))''',
+app.config['OPENCENSUS'] = {
+    'TRACE': {
+        'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
+        'EXPORTER': '''opencensus.ext.azure.trace_exporter.AzureExporter(
+            opencensus.ext.azure.common.Options(
+                instrumentation_key="00000000-0000-0000-0000-000000000000",
+                timeout=29.9,
+            ))''',
+    },
 }
 middleware = FlaskMiddleware(app)
 
--- a/contrib/opencensus-ext-azure/examples/custom.py
+++ b/contrib/opencensus-ext-azure/examples/custom.py
@@ -1,0 +1,42 @@
+# Copyright 2019, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from flask import Flask
+from opencensus.ext.flask.flask_middleware import FlaskMiddleware
+
+app = Flask(__name__)
+# TODO:
+# Replace 00000000-0000-0000-0000-000000000000 with your Instrumentation Key
+# https://docs.microsoft.com/en-us/azure/azure-monitor/app/create-new-resource
+app.config['OPENCENSUS_TRACE'] = {
+    'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
+    'EXPORTER': '''opencensus.ext.azure.trace_exporter.AzureExporter(
+        opencensus.ext.azure.common.Options(
+            instrumentation_key="00000000-0000-0000-0000-000000000000",
+            timeout=29.9,
+        ))''',
+}
+middleware = FlaskMiddleware(app)
+
+
+@app.route('/')
+def hello():
+    return 'Hello World!'
+
+
+if __name__ == '__main__':
+    import logging
+    logger = logging.getLogger('werkzeug')
+    logger.setLevel(logging.ERROR)
+    app.run(host='localhost', port=8080, threaded=True)
--- a/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
+++ b/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import six
 import sys
 
 import flask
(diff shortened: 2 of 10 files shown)


opencensus/common/configuration/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+
+__all__ = ['Namespace', 'eval']
+
+
+class Namespace(object):
```
reyang: @c24t, need your help to pick a better name.

contrib/opencensus-ext-azure/examples/custom.py
```diff
@@ -0,0 +1,42 @@
...
+# limitations under the License.
+
+from flask import Flask
+from opencensus.ext.flask.flask_middleware import FlaskMiddleware
+
+app = Flask(__name__)
+# TODO:
+# Replace 00000000-0000-0000-0000-000000000000 with your Instrumentation Key
+# https://docs.microsoft.com/en-us/azure/azure-monitor/app/create-new-resource
+app.config['OPENCENSUS_TRACE'] = {
```
reyang: @c24t @songy23 I wish to know what's the long term configuration structure in your mind.

For example, today we put "PROPAGATOR" under "OPENCENSUS_TRACE", which means if we have tags propagation (e.g. we could have W3C correlation context protocol for tags propagation under HTTP, or a binary protocol for gRPC), either we need to introduce something like `OPENCENSUS_TAGS.PROPAGATOR`, or decouple propagator from traces (e.g. having a top level propagator concept, which takes the entire context, pick whatever is needed, and serialize/deserialize).
Similar question for log sampler, potentially we will need to have `OPENCENSUS_LOG.SAMPLER`.

reyang: Chatted with @songy23 offline, it seems gRPC integration is taking the approach of separated propagators.

So, the configuration might look like:

```
app.config['OPENCENSUS'] = {
    'TRACE': {
        'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
        'EXPORTER': 'opencensus.ext.stackdriver.trace_exporter.StackdriverExporter()',
        'PROPAGATOR': 'google_cloud_format.GoogleCloudFormatPropagator()'
    },
    'METRICS':  {
        'EXPORTER': 'opencensus.ext.stackdriver.metrics_exporter.StackdriverExporter()'
    },
    'LOGS': {
        'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
        'EXPORTER': 'opencensus.ext.stackdriver.log_exporter.StackdriverExporter()'
    },
    'TAGS': {
        'PROPAGATOR': 'CorrelationContextPropagator()'
    }
}
```

@songy23 please correct me if I'm wrong.

reyang: @c24t this also brings the question of where should samplers go, do we want to have ProbabilitySampler under `trace` or put it in the general place so both traces and logs can share them. Or should log respect the sampling decision if there is a trace context (and what if the customer is using log feature only, without using any of the trace features)?

opencensus/common/configuration/__init__.py
c24t: This is a pretty amazing little class, but I think I'm missing something here. Do you want to use `Namespace` objects to do some complicated config? Or do you expect only to use it via `load`?

As I understand it, this class lets you `eval` without having to import `opencensus` explicitly. It's a callable that (recursively creates callables that) lazily imports all the ancestor modules of the named target.

How is using a `Namespace` here different than using the imported module directly? E.g.

```python
import opencensus

def load(expr):
    return eval(expr, {}, {'opencensus': opencensus})
```

reyang: Import `opencensus` won't work in several conditions:

```
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import opencensus
>>> opencensus.ext.azure.trace_exporter.AzureExporter
  File "<stdin>", line 1, in <module>
AttributeError: module 'opencensus' has no attribute 'ext'
>>> import opencensus.ext.azure.trace_exporter
>>> opencensus.ext.azure.trace_exporter.AzureExporter
<class 'opencensus.ext.azure.trace_exporter.AzureExporter'>
>>>
```

c24t: That makes sense, turns out my example only worked because I'd already imported all the parent modules of my test class by calling your `load` first.

c24t: What about `Loader` or `AutoImporter`? Not obviously better than `Namespace`, this class my be too weird to name well.

contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
```diff
@@ -105,90 +87,28 @@ def init_app(self, app):
...
-                                        self.DEFAULT_SAMPLER))
-        self.exporter = (self.exporter
-                         or settings.get('EXPORTER',
-                                         self.DEFAULT_EXPORTER))
-        self.propagator = (self.propagator
-                           or settings.get('PROPAGATOR',
-                                           self.DEFAULT_PROPAGATOR))
-
-        # get params from app config
-        params = self.app.config.get('OPENCENSUS_TRACE_PARAMS', {})
```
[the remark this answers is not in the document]
reyang: Yep.
And probably instead of having `OPENCENSUS_TRACE`, use `OPENCENUS` as the configuration name (and having the second level `TRACE`, `METRICS`, etc.).

contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
```diff
@@ -68,24 +49,25 @@ class FlaskMiddleware(object):
     :type blacklist_paths: list
     :param blacklist_paths: Paths that do not trace.
 
-    :type sampler: :class: `type`
-    :param sampler: Class for creating new Sampler objects. It should extend
+    :type sampler: :class:`~opencensus.trace.samplers.base.Sampler`
+    :param sampler: Instance of sampler objects. It should extend
```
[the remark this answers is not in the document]
reyang: Good idea! Simplifying the wording now.

tests/system/trace/flask/main.py
```diff
@@ -41,10 +42,12 @@
...
 # transport.  This is intentionally different from the Django system test which
 # is using sync transport to test both.
-exporter = stackdriver_exporter.StackdriverExporter(
-    transport=async_.AsyncTransport)
-
-middleware = FlaskMiddleware(app, exporter=exporter)
+middleware = FlaskMiddleware(
+    app,
+    exporter=stackdriver_exporter.StackdriverExporter(
+                 transport=async_.AsyncTransport),
```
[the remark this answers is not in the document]
reyang: Good catch. Fixed.

reyang:
> Making config executable code may open up a can of worms, especially since some modules have import time effects... including reading config. I don't see any specific security risks here (since I wouldn't expect untrusted code to be used as config), but we should consider this before release.

I thought about this before making the change. It looks like Python by convention uses `*.py` file for configuration, which implies evaluation. I plan to fix all TBD docs after the refactor for Django/Pyramid. I will add a note in the document on the potential security consideration.
 
> This PR changes a lot of defaults, it'd be helpful to list these somewhere before the release. Users might reasonably expect `"JaegerExporter"` (old) and `"opencensus.ext.jaeger.trace_exporter.JaegerExporter()"` (new) to produce the same exporters in code.

Agree. I will write a section the release note.

opencensus/common/configuration/__init__.py
reyang: I will keep Namespace for now, as it is weird to have `Loader.parent` being another `Loader` (same applied for `AutoImporter`). Would be open for a better name.

PR #622, opened 2019-04-16
Respect default labels in StackdriverStatsExporter

### LINKED ISSUE #609: Override default Stackdriver monitoring labels
I am trying to override the default labels published to Stackdriver. 
In StackdriverStatsExporter, the exporter is initialized with Options including ``defaut_monitoring_labels``. The comments for this option say: ``
If unset, this defaults to a single label with key "opencensus_task" and value "py-<pid>@<hostname>".``
Later on in the function ``new_label_descriptors``, it appears that it always adds in the ``opencensus_task`` info to the label descriptors even if ``defaut_monitoring_labels`` is set:
```python
def new_label_descriptors(defaults, keys):
    """ create labels for the metric_descriptor
        that will be sent to Stackdriver Monitoring
    """
    label_descriptors = []
    for key, lbl in defaults.items():
        label = {}
        label["key"] = sanitize_label(key)
        label["description"] = lbl
        label_descriptors.append(label)

    for label_key in keys:
        label = {}
        label["key"] = sanitize_label(label_key.key)
        label["description"] = sanitize_label(label_key.description)
        label_descriptors.append(label)
    label_descriptors.append({"key": OPENCENSUS_TASK,
                              "description": OPENCENSUS_TASK_DESCRIPTION})
    return label_descriptors
```
This function is called with the defaults set to default_monitoring_labels option:
```python
    default_labels = self.options.default_monitoring_labels
    if default_labels is None:
        default_labels = {}
    desc_labels = new_label_descriptors(default_labels, oc_md.label_keys)
```
Am I misunderstanding this or is this the desired behavior? Overall I am just looking for a way to avoid publishing a label to Stackdriver with potential for very high cardinality

c24t:
This PR fixes two longstanding bugs affecting default labels in the `StackdriverStatsExporter`:
 - the "opencensus_task" label was attached to all exported metrics, regardless of `default_monitoring_labels` (#609)
- other custom default labels were never attached to metrics, only to metric descriptors

It also includes some opportunistic cleanup work, including refactoring `Options` and calling `get_task_value` once on exporter creation instead of on each export.

This behavior matches the [go](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stackdriver.go) [exporter](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stats.go) more closely, but still includes options that have been deprecated in go. There's still more work to do to get these exporters in sync.

Fixes #609.

### DIFF
--- a/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
+++ b/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
@@ -27,6 +27,8 @@ import google.auth
 from opencensus.common import utils
 from opencensus.common.monitored_resource import monitored_resource
 from opencensus.common.version import __version__
+from opencensus.metrics import label_key
+from opencensus.metrics import label_value
 from opencensus.metrics import transport
 from opencensus.metrics.export import metric as metric_module
 from opencensus.metrics.export import metric_descriptor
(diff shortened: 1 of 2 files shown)


contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -424,19 +421,12 @@ def new_label_descriptors(defaults, keys):
...
-        label["key"] = sanitize_label(key)
-        label["description"] = lbl
+        label["key"] = sanitize_label(lk.key)
+        label["description"] = lk.description
         label_descriptors.append(label)
 
-    for label_key in keys:
-        label = {}
-        label["key"] = sanitize_label(label_key.key)
-        label["description"] = sanitize_label(label_key.description)
```
c24t: As it's written `sanitize_label` shouldn't apply to descriptions.

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -65,70 +67,67 @@
...
+
+     Set `default_monitoring_labels` to `{}` to avoid getting the default
+     "opencensus_task" label. You should only do this if you know that
+     `resource` uniquely identifies this process.
+
+    :type project_id: str
+    :param project_id: The ID GCP project to export metrics to, fall back to
+        default application credentials if unset.
+
+    :type resource: str
```
songy23: Type should be `google.cloud.monitoring_v3.types.MonitoredResource`.

c24t: I thought so too, but it actually expects a string in `set_monitored_resource`. This is the monitored resource's (string-valued) type, not the resource itself.

That said, I think it would make sense to change this (and `set_monitored_resource`) to take a `MonitoredResource` instead.

songy23: Agree, that makes more sense and is consistent with Java and Go options.

PR #624, opened 2019-04-18
Refactor Django/Pyramid configuration

reyang:
This is part of the effort for #394, a follow up on #621.

Pyramid configuration is using explicit objects instead of string. So the change here is just to keep consistent with Flask configuration structure, but not providing any dynamic evaluation.

1. Change the configuration structure to align with Flask (#621).
2. Use W3C TraceContext as the default propagator.
3. Update README and CHANGELOG.


contrib/opencensus-ext-django/README.rst
```diff
@@ -35,29 +35,17 @@ And add this line to the ``INSTALLED_APPS`` section:
...
-        'SAMPLING_RATE': 0.5,
-        'SERVICE_NAME': 'my_service',
-        'ZIPKIN_EXPORTER_HOST_NAME': 'localhost',
-        'ZIPKIN_EXPORTER_PORT': 9411,
-        'ZIPKIN_EXPORTER_PROTOCOL': 'http',
+    OPENCENSUS = {
+        'TRACE': {
+            'SAMPLER': 'opencensus.trace.samplers.ProbabilitySampler(rate=1)',
+            'EXPORTER': '''opencensus.ext.ocagent.trace_exporter.TraceExporter(
+                service_name='foobar',
```
c24t: In the new scheme if we wanted a single config entry for the service name where would it go?

Based on [your comment on the flask PR](https://github.com/census-instrumentation/opencensus-python/pull/621/files#r276471411), it sounds like there's no top-level config namespace for values shared between `TRACE` and `METRICS`, is that right?

contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
```diff
@@ -118,79 +109,28 @@ class OpencensusMiddleware(MiddlewareMixin):
...
-
-        self.blacklist_hostnames = settings.params.get(
-            BLACKLIST_HOSTNAMES, None)
-
-        # Initialize the propagator
-        self.propagator = self._propagator()
+        settings = getattr(django.conf.settings, 'OPENCENSUS', {})
+        settings = settings.get('TRACE', {})
+
+        self.sampler = settings.get('SAMPLER', None) or \
```
c24t: I missed it in #621, but are you doing `get(..., None) or foo` instead of `get(..., foo)` because `settings['SAMPLER']` might be an empty string?

Also do you expect settings values to be a mix of objects and config strings? This pattern might also get confusing if we ever store plain old strings in config and read them alongside these config strings that represent objects.

tests/unit/trace/test_tracer.py
```diff
@@ -34,7 +35,7 @@ def test_constructor_default(self):
         assert isinstance(tracer.sampler, AlwaysOnSampler)
         assert isinstance(tracer.exporter, print_exporter.PrintExporter)
         assert isinstance(tracer.propagator,
```
c24t: newline!

contrib/opencensus-ext-django/README.rst
reyang: There is actually. I would imagine something like this:
```
'OPENCENSUS': {
    'TRACE': {},
    'SERVICE_NAME': 'foobar',
}
```
And this is the reason why I put a hierarchy so we have `OPENCENUS` instead of everything going to `OPENCENSUS_TRACE`.

contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
reyang: I use `or` to prevent evaluation:

Constructor is called only when setting evaluates to False:
```
settings.get('SAMPLER', None) or always_on.AlwaysOnSampler()
```
Constructor is always called, which could be costly (depending on the actual implementation):
```
settings.get('SAMPLER', always_on.AlwaysOnSampler())
```

reyang: Yeah, I was worried about this as well.
My current thinking is: for each configuration entry, we will have to design whether it is a string to be evaluated as a Python expression, or just a plain value. And this is why we have several `configuration.eval` calls (to make it explicit) instead of doing a `loop and eval`.

c24t: What about this?

```python
try:
    sampler = configuration.load(settings['SAMPLER'])
except KeyError:
    sampler = always_on.AlwaysOnSampler()
```

That way the type check is gone and it's clear from the code that the setting is a string.

In any case you can replace `get(foo, None)` with `get(foo)` if you do keep the `get`s.

tests/unit/trace/test_tracer.py
reyang: Sorry for these noisy format issues. Will try the vim automation as you suggested.

contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
reyang: Do we want the `SAMPLER` config to be a string, or it can be either a string or Exporter object?
I was following the original approach where the config could be either a string (where we need to construct the object based on the string) or a strong type.

The implementation in this PR is trying to support both:
```
'SAMPLER': 'opencensus.trace.samplers.always_on.AlwaysOnSampler()'
```
and
```
'SAMPLER': always_on.AlwaysOnSampler()
```

This gives us the benefit that if people use configuration in python source code, then can use the latter approach to benefit from static check (which seems to be common for Pyramid). If people loads the configuration from a JSON object or something similar, they can get dynamic evaluation (just install the ext, change config, no import required).
This comes with a cost of making the config/implementation a bit more complex. I'm open to take either approach.

reyang: @c24t please let me know your thoughts.

I'll merge this PR now and work on the documentation / notes on breaking changes. I'm open to adjust the current implementation to make it better. (And I'm thinking if it is a good idea to extract the configuration logic into core opencensus lib, given there are a lot of duplicated logic in Flask/Django/Pyramid)

PR #630, opened 2019-04-24
Prepare for 0.5.0 release

c24t:
Update versions and changelogs for the `0.5.0` release of `opencensus` and new releases of most ext packages.

Most ext packages got a micro version update for the docs changes in #625, packages affected by the config changes in #621 and #624 got a minor version update.

See [the diff against the `v0.4.x` branch](https://github.com/census-instrumentation/opencensus-python/compare/v0.4.x...c24t:v0.5.x?expand=1) for the full set of files changed since the last release.

### DIFF
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+## Unreleased
+
 ## 0.1.0
 Released 2019-04-23
 
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Changelog
 
-## Unreleased
-- Initial project skeleton
+## 0.1.0
+Released 2019-04-23
+
+- Initial release
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -3,6 +3,6 @@
 ## Unreleased
 
 ## 0.1.0
-Released 2019-04-23
+Released 2019-04-24
 
 - Initial release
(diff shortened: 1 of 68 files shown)


contrib/opencensus-ext-dbapi/setup.py
```diff
@@ -39,7 +39,7 @@
     include_package_data=True,
     long_description=open('README.rst').read(),
     install_requires=[
-        'opencensus >= 0.5.dev0, < 1.0.0',
+        'opencensus >= 0.3.0, < 1.0.0',
```
reyang: 0.5?

c24t: It actually doesn't need the newer version since nothing that it uses has changed since `0.3.0`. I changed the requirement to `0.5.dev0` on master for this and all other ext packages so it would be impossible to install a non-development version in development.

c24t: They'll still install a newer version via `pip`, this is just the minimum required version.

c24t: I'm willing to always require the most recent version instead. It costs us some flexibility, but would make releases simpler.

reyang: Got it, up to you :)

My personal preference would tend towards simplicity than flexibility (e.g. technically we can run some .NET 4 libraries on .NET 2, but why would we want to encourage that in the first place?).

I think the major outcome of having individual ext packages are:
1. People doesn't have to install everything if they don't need them.
2. It makes ext update (e.g. if there is a MySQL major version change which we want to support) easier without having to release the entire opencensus.

Using the ext package with an old version of opencensus wasn't our original goal, although one can argue there might be some benefits.

contrib/opencensus-ext-azure/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Changelog
 
-## Unreleased
-- Initial project skeleton
+## 0.1.0
```
songy23: Nit: `Unreleased` was removed by accident.

c24t: Good catch, thanks.

contrib/opencensus-ext-dbapi/setup.py
c24t: I agree the benefit isn't worth the maintenance cost here, changed all of these in e4201f7.

PR #634, opened 2019-04-30
Add issue templates

c24t:
Add issue templates for bug reports and feature requests.

These templates use the bullet points from https://github.com/census-instrumentation/opencensus-node/pull/186, but the text and formatting are changed. Unlike in node, these templates do not add the "triage-me" label.

We may decide that this is unnecessary boilerplate in the future, and should revisit these templates after using them for at least few weeks.

cc @dinooliva


c24t:
See also https://help.github.com/en/articles/about-issue-and-pull-request-templates.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Bug Report
+about: Create a report to help us improve
+labels: bug
+---
+
+**Describe your environment.** Describe any aspect of your environment relevant to the problem, including your Python version, [platform](https://docs.python.org/3/library/platform.html), cloud hosting provider, etc. If you're reporting a problem with a specific version of the library, please check to see whether the problem has been fixed on master.
+
+**Steps to reproduce.**
+Describe exactly how to reproduce the error. Include a code sample if applicable.
```
reyang: In case people use `pip install opencensus`, we might need to collect the version of packages.

c24t: Even if they installed from source they might have old versions of dependences because of other packages' dependencies in the same env. It may be useful sometimes, but I'm trying to avoid asking for too much info here, and I think this is already a lot of boilerplate.

c24t: Added a note in the env section above.

PR #635, opened 2019-04-30
Improve test process

reyang:
1. Instead of stick with the old version of test packages, always use the latest.
2. Remove tox.ini (still trying to figure out why do we need it).
3. Update test cases to resolve the deprecation warnings.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,1 @@
+OpenCensus Authors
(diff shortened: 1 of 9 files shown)


reyang:
> LGTM though I'm not sure if we need the AUTHOR file. All other repos have it so maybe leave it here for now.

Added AUTHORS file back since it is not really the purpose of this PR (will revisit later in another PR).

PR #639, opened 2019-05-01
Override sampling decision

### LINKED ISSUE #481: Override trace sampling decision

c24t:
Fixes #481.

This PR changes the sampler API so that `should_sample` now takes the span context, not just the trace ID. This moves the responsibility for checking the sampling bit from the tracer to the sampler.

This PR also:
- Changes the default sampler from `AlwaysOnSampler` to `ProbabilitySampler`
- Changes the default sampler in the django and flask extensions
- Changes `ProbabilitySampler`'s default sampling rate to match the other clients
- Moves the samplers into a single module

Note that we don't allow the user to override the sampler for a single span (as e.g. the java client does in `SpanBuilder`). Overriding the tracer's sampler solves the problem in #481, but we may want to add a span-specific sampling option later.


opencensus/trace/tracer.py
```diff
@@ -74,8 +74,7 @@ def should_sample(self):
         :rtype: bool
         :returns: Whether to trace the request or not.
         """
-        return self.span_context.trace_options.enabled \
-            or self.sampler.should_sample(self.span_context.trace_id)
+        return self.sampler.should_sample(self.span_context)
```
c24t: This is the important change: we used to short-circuit here, which meant we couldn't refuse to sample (from e.g. the `AlwaysOffSampler`) if the sampling bit was already set.

Note that we're reading `span_context.trace_options.enabled` in the sampler... and then using the sampling decision to set it. This works because we only use the sampler once at init time to set `tracer.tracer` (and the sampling bit). This is potentially confusing, but refactoring `Tracer` to fix this is outside the scope of this PR.

README.rst
```diff
@@ -116,8 +116,11 @@ There are several things you can customize in OpenCensus:
   other exporters are provided as `extensions <#trace-exporter>`__.
 
 * **Sampler**, which determines how traces are sampled.
-  The default sampler is ``AlwaysOnSampler``, other samplers include the
-  ``AlwaysOffSampler`` and ``ProbabilitySampler``.
+  The default sampler is the ``ProbabilitySampler``, which samples (i.e.
+  enables tracing for) a percentage of all requests. Sampling is deterministic
+  according to the trace ID. To force sampling for all requests, or to disable
+  prevent any request from being sampled, see ``AlwaysOnSampler`` and
```
reyang: "disable prevent" -> "prevent"

c24t: Oof, thanks.

opencensus/trace/samplers/__init__.py
```diff
@@ -12,11 +12,83 @@
...
+        :type span_context: :class:`opencensus.trace.span_context.SpanContext`
+        :param span_context: The span context.
+
+        :rtype: bool
+        :returns: Whether to sample the request according to the context.
+        """
+        raise NotImplementedError
+
+
+class AlwaysOnSampler(Sampler):
```
reyang: Having `samplers.FoobarSampler` is much much better and more intuitive than `samplers.foobar_sampler.FoobarSampler`! Wish we could have that for propagators and other stuff.

README.rst
```diff
@@ -116,8 +116,11 @@ There are several things you can customize in OpenCensus:
   other exporters are provided as `extensions <#trace-exporter>`__.
 
 * **Sampler**, which determines how traces are sampled.
```
reyang: One thing we probably need to consider - update the README file so first time users wouldn't be surprised when there is no trace dumped to the console due to the default low sample rate:

```
tracer = Tracer()
with tracer.span('foo'):
    pass
```

c24t:
> do we expect the majority of people would use 0.01% as the default sample rate?

I expect that a lot of people won't change the default settings no matter what they are (we have some evidence that people leave the `AlwaysOnSampler` on and run up huge monitoring bills), and this seems like a sane default even for large applications. It also matches the go and java clients.

> why do we need AlwaysOffSampler, or we can use ProbabilitySampler with rate=0

Without `AlwaysOffSampler` there's no option to disable sampling entirely. `AlwaysOnSampler` is redundant though, and AFAIK just exists for completeness.

> going forward we would need the agent to be able to pass in sampling policy, which could override the sampler.

We may need to allow setting per-span samplers for this.

README.rst
c24t: Added in 014b906, it does make the docs a bit uglier though.

PR #640, opened 2019-05-01
Update azure ctor, docs and samples

reyang:
* Updated samples and docs to reflect the sampler change in #639.
* Changed the AzureExporter constructor to take kwargs.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
@@ -41,8 +41,8 @@ class AzureExporter(base_exporter.Exporter):
     :param options: Options for the exporter. Defaults to None.
     """
 
-    def __init__(self, options=None):
-        self.options = options or Options()
+    def __init__(self, **options):
+        self.options = Options(**options)
         if not self.options.instrumentation_key:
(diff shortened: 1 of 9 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -41,8 +41,8 @@ class AzureExporter(base_exporter.Exporter):
     :param options: Options for the exporter. Defaults to None.
     """
 
-    def __init__(self, options=None):
-        self.options = options or Options()
+    def __init__(self, **options):
+        self.options = Options(**options)
```
c24t: Out of curiosity: why the change? And what does having a separate class for options give you now that you're splatting this?

reyang: @c24t Sorry I didn't notice your question yesterday!

> why the change?

Given all the parameters can be considered as options, it is simpler to use kwargs (which is more Pythonic).

> And what does having a separate class for options give you now that you're splatting this?

To provide default value, instead of putting the logic in ctor like:

```
if instrumentation_key is None:
    instrumentation_key = os.getenv('foo')
```

This gives the benefit of reusing the options for LogExporter and MetricsExporter.

PR #641, opened 2019-05-01
Minor cleanup

reyang:


AUTHORS
```diff
@@ -1 +0,0 @@
-OpenCensus Authors
```
songy23: I'm not sure how useful this file is, but since all other repos have it, I'd suggest to keep it.

reyang: I checked the other GitHub repos (e.g. https://github.com/nodejs/node/blob/master/AUTHORS), it seems the purpose is to keep a list or authors.
If we don't plan to put the actual names, might be good to just remove it?

songy23: I don't have a strong opinion on this - @c24t WDYT?

reyang: In C# repo we don't have it https://github.com/census-instrumentation/opencensus-csharp.
In Java repo we have "Google Inc.".

songy23: That one in Java is due to historical reason. Previously the license header is "Google Inc." instead of "OpenCensus Authors". Later we changed the header but looks like we forgot to update the AUTHORS file.

c24t: I think it's fine to remove it, we've got author info in the setup file too.

PR #644, opened 2019-05-06
Add details about aggregation pipelines & avoid get(default) parameter

angelini:
When issuing aggregations via Pymongo, they store the pipeline as a command attribute.

This PR adds support for storing the pipeline as a trace attribute.

We've also seen the following errors in production with a Mongo `3.6.12` instance an Pymongo `3.6.1`.

```
  File "<REDACTED>/.virtualenv/lib/python3.7/site-packages/pymongo-3.6.1-py3.7-linux-x86_64.egg/pymongo/monitoring.py", line 739, in publish_command_start 
    subscriber.started(event) 
  File "<REDACTED>/.virtualenv/lib/python3.7/site-packages/opencensus_ext_pymongo-0.1.2-py3.7.egg/opencensus/ext/pymongo/trace.py", line 54, in started 
    _attr = event.command.get(attr, default=None) 
TypeError: get() takes no keyword arguments 
```

It seems like not all `event.command` objects are `SON` instances (https://github.com/mongodb/mongo-python-driver/blob/master/bson/son.py#L32).

Since the `default` keyword argument isn't being used here, the default value is `None`, I opted to remove the use of it.

### DIFF
--- a/contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
+++ b/contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
@@ -44,6 +44,7 @@ class Test_pymongo_trace(unittest.TestCase):
             'filter': 'filter',
             'sort': 'sort',
             'limit': 'limit',
+            'pipeline': 'pipeline',
             'command_name': 'find'
         }
 
@@ -51,6 +52,7 @@ class Test_pymongo_trace(unittest.TestCase):
             'filter': 'filter',
             'sort': 'sort',
             'limit': 'limit',
+            'pipeline': 'pipeline',
             'request_id': 'request_id',
             'connection_id': 'connection_id'
         }
--- a/contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
+++ b/contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
@@ -103,9 +103,8 @@ class MockCommand(object):
     def __init__(self, command_attrs):
         self.command_attrs = command_attrs
 
-    def get(self, key, default=None):
-        return self.command_attrs[key] \
-            if key in self.command_attrs else default
+    def get(self, key):
+        return self.command_attrs.get(key)
 
 
 class MockEvent(object):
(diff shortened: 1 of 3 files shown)


reyang:
@tjzo please help to review.

contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
```diff
@@ -101,9 +103,9 @@ class MockCommand(object):
     def __init__(self, command_attrs):
         self.command_attrs = command_attrs
 
-    def get(self, key, default=None):
+    def get(self, key):
         return self.command_attrs[key] \
-            if key in self.command_attrs else default
+            if key in self.command_attrs else None
```
c24t: Nit: might as well use `self.command_attrs.get(key)`.

angelini:
> > It seems like not all event.command objects are SON instances

> It looks like this PR doesn't fix this?

As best I can tell the commands aren't `SON` objects, but they have a `get` method. This change has fixed it for our production use.

> if you've got a sample command to reproduce the error we can open another ticket to track this.

Yeah that would be great, working on a getting one but so far we've only seen this in production and haven't been able to reproduce it locally.

angelini:
@c24t I was just able to confirm from production that the command causing the error is an empty dictionary.

```
> print('command for (%s) %s' % (type(event.command), event.command))
command for (<class 'dict'>) {}
```

And this is the same error we were seeing.

```
In [1]: {}.get('a')

In [2]: {}.get('a', default=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-9332c12c9211> in <module>
----> 1 {}.get('a', default=1)

TypeError: get() takes no keyword arguments
```

Removing the use of `default=None` https://github.com/census-instrumentation/opencensus-python/pull/644/files#diff-bb0ccb334b607c4945ac2a1ed4ac9776R54 fixes it.

reyang:
@angelini would you update the `CHANGELOG.md` and rebase to the latest master branch? Once done, we're ready to merge.

angelini:
@reyang it's been rebased and the changelog has been updated

reyang:
Thanks for your contribution! @angelini

PR #646, opened 2019-05-08
Add contribute message for python versions

lzchen:
nox automation tests will fail if any of the python versions currently supported by OpenCensus is not installed. Currently, these are python 2.7, 3.4, 3.5 and 3.6

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ information on using pull requests.
 Prerequisites:
 
 * You need to have Python installed.
-* You must have every version of Python that OpenCensus supports installed for testing (see https://pypi.org/project/opencensus/ for which versions are supported).
+* You must have every major version of Python that OpenCensus supports installed for testing (see https://pypi.org/project/opencensus/ for which versions are supported).
 * You need to fork the project in GitHub.
 
 Clone the upstream repo:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -20,6 +20,7 @@ information on using pull requests.
 Prerequisites:
 
 * You need to have Python installed.
+* You must have every version of Python that OpenCensus supports installed for testing (see https://pypi.org/project/opencensus/ for which versions are supported).
```
reyang: "every version" -> "every major version" since we don't distinguish minor versions (e.g. 3.6.1 and 3.6.2).

CONTRIBUTING.md
```diff
@@ -20,6 +20,7 @@ information on using pull requests.
 Prerequisites:
 
 * You need to have Python installed.
```
reyang: Do we want to merge these two lines given they both talk about Python runtime requirements?

CONTRIBUTING.md
lzchen: Good point. We don't want them installing every single release!

CONTRIBUTING.md
lzchen: Makes sense. It's probably required to have all versions installed anyways regardless of whether you are only running tests.

Issue #478, opened 2019-01-31
Standardize import style


c24t:
The google python style guide [recommends against importing classes](http://google.github.io/styleguide/pyguide.html#22-imports). This makes it clear which namespace defines each class, but is cumbersome and leads to ugly workarounds for common namespace names (e.g. `... import view as view_module`).

Most modules have a consistent internal import style, but there's no consistent style across the library.  We should decide whether to follow the style guide everywhere or break from it everywhere.

reyang:
We might need to change the import path due to the refactoring work in #445 (potential breaking changes).

kshithijiyer:
I feel standardizing imports as per pep8 standards would be better.
https://www.python.org/dev/peps/pep-0008/

reyang:
@kshithijiyer would you explain why you would prefer pep8?

@c24t @lzchen @songy23 how do you think?

kshithijiyer:
Pep8 standard in general improves code readability in general. Apart from that it has now become the standard for python code. I would say that the standard mentioned in the pervious comments is good but creates and issue when we actually import everything into the namespace. For example if you wish to use `randint` function from 
 `random` module following the pervious standard says you can do an  `import random`  by doing so you actually import alot of unnecessary functions which will never be used inside the namespace as well. Where as pep8 says never to use `from <module> import *` which is same as `import <module>`. Hence looking at this I personally feel it's better to use pep8 instead of going with any other standard.

c24t:
We already follow PEP 8, this ticket is about consistently importing either classes or modules.

> you actually import a lot of unnecessary functions which will never be used inside the namespace as well

Whether you import the module or individual functions/classes the import machinery creates the module object and executes the code, there's no reason to prefer importing individual functions/classes except personal style.

> pep8 says never to use `from <module> import *` which is the same as `import <module>`

`import *` binds all names defined in the imported module, `import foo` binds the module object itself.

dineshkrishnareddy:
hey @c24t @kshithijiyer can I work on this?

reyang:
@dineshkrishnareddy definitely!
We kind of solved this issue in OpenTelemetry Python. It'll be great if you could help on OpenCensus.

FYI:
* https://github.com/open-telemetry/opentelemetry-python/blob/master/.isort.cfg
* https://github.com/open-telemetry/opentelemetry-python/blob/1968eedacdceec9856b75f60979391e38a8cb0e6/tox.ini#L58

PR #648, opened 2019-05-08
[WIP] Changing import style as per standards. (#478)

kshithijiyer:
Signed-off-by: kshithijiyer <kshithij.ki@gmail.com>

### DIFF
--- a/opencensus/common/backports/__init__.py
+++ b/opencensus/common/backports/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import six
+from six import raise_from
 import weakref
 
 
@@ -35,7 +35,7 @@ class WeakMethod(weakref.ref):  # pragma: NO COVER
         except AttributeError:
             error = TypeError("argument should be a bound method, not {}"
                               .format(type(meth)))
-            six.raise_from(error, None)
+            raise_from(error, None)
 
         def _cb(arg):
             # The self-weakref trick is needed to avoid creating a reference
--- a/opencensus/log/__init__.py
+++ b/opencensus/log/__init__.py
@@ -14,12 +14,13 @@
 
 from collections import namedtuple
 from copy import copy
-import logging
+from logging import (getLogger, LoggerAdapter,
+                     setLoggerClass, getLoggerClass)
 
 from opencensus.trace import execution_context
 
 
-_meta_logger = logging.getLogger(__name__)
+_meta_logger = getLogger(__name__)
 
 TRACE_ID_KEY = 'traceId'
 SPAN_ID_KEY = 'spanId'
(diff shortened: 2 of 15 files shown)


kshithijiyer:
Fixes for issue:  #478

reyang:
@kshithijiyer thanks for working on this!

Do you intend to fix #478 by making changes to all the imports? Or this is just for a single file?
Is there a way we can enforce the import style using some lint rule going forward?

kshithijiyer:
Yes, I do intend to work on the issue if it's still up-for-grab. Will edit and push it here as I make the changes. 
I am unaware if there is a way to configure pylint to check for it.

reyang:
@kshithijiyer got it. Please put "[WIP]" in the PR title for now, remove "[WIP]" when it is ready for review.

opencensus/common/backports/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import six
+from six import raise_from
```
reyang: @kshithijiyer would you clarify what's the desired state? Which "standard" are you trying to follow?

`import six` does follow the guidance http://google.github.io/styleguide/pyguide.html#224-decision, and now we are changing it to be against the guidance.

opencensus/log/__init__.py
```diff
@@ -14,12 +14,13 @@
 
 from collections import namedtuple
 from copy import copy
-import logging
+from logging import (getLogger, LoggerAdapter,
```
reyang: This seems to be against http://google.github.io/styleguide/pyguide.html#224-decision.

opencensus/common/backports/__init__.py
kshithijiyer: I am just importing what is needed and not the entire thing. Didn't understand how is this against the guidance can you please explain?

reyang: No I don't think so.

The guidance asks for importing the **module** instead of individual **class/function**. If you take a look at http://google.github.io/styleguide/pyguide.html#22-imports, you will see the statement "_Use import statements for packages and modules only, not for individual classes or functions_".

The guidance says do this:

```
from sound.effects import echo
...
echo.EchoFilter(input, output, delay=0.7, atten=4)
```

And don't do this:
```
from sound.effects.echo import EchoFilter
...
EchoFilter(input, output, delay=0.7, atten=4)
```

kshithijiyer: Could you point out where else this mistake has happened?

reyang: We cannot define mistakes without having clear rules.

My suggestion is to take a step back, clarify what is the desired state (e.g. having a set of rules, or simply saying we are going to follow http://google.github.io/styleguide/pyguide.html). Then we know what's "right" or "wrong" based on the rules.

kshithijiyer: I prersonally feel that the best practice is to import whatever is needed instead of importing everything. Say you have 10 functions in a module and you just wish to use just 2 functions out of those 10 functions. In this case what's the point of importing the entire module instead you can just import 2 functions into the namespace?

c24t:
I'm closing this PR doesn't solve the problem as described. @kshithijiyer feel free to open another PR with a different set of changes if you still want to work on this.

To fix #478 a PR should make imports consistent across all modules, either never importing classes (preferred) or always importing classes instead of their containing modules. It shouldn't change function imports (e.g. `import time` to `from time import sleep`).

kshithijiyer:
Acknowledged

Issue #459, opened 2019-01-18
Limit span attachments based on global config


c24t:
From [the spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#limits):

> We define limits on the number of attributes, annotations, message events and links on each span
in order to prevent unbounded memory increase for long-running spans.

We need to drop annotations (and etc.) at record time instead according to the global `TraceConfig`. This is only used in the [agent exporter](https://github.com/census-instrumentation/opencensus-python/blob/989470edce8fa7a82813e3846f774fd4927b5b07/opencensus/trace/exporters/ocagent/trace_exporter.py) now.

See for example the [java client implementation](https://github.com/census-instrumentation/opencensus-java/blob/c0f3a4cf4c437b9a4ac983f1bd44781f60d0182f/impl_core/src/main/java/io/opencensus/implcore/trace/RecordEventsSpanImpl.java#L384-L418).

PR #649, opened 2019-05-09
Limit span attachments

c24t:
This PR limits the number of span attachments (attributes, annotations, message events, and links) to the defaults listed on [the `TraceConfig` spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md), dropping the oldest attachment of each type when we reach the limit. I'm tracking the number of dropped attachments, but not yet emitting a stat.

I had to change `Span` and `SpanData` to replace `time_events` (each of which could be an annotation or message event) with new `annotations` and `message_events` attrs. This matches the [java client](https://github.com/census-instrumentation/opencensus-java/blob/f68f73b5926b97ada0c6952f31856e2a9e0e0de0/api/src/main/java/io/opencensus/trace/export/SpanData.java#L67-L68) more closely, but is a breaking API change.

Addresses #459


opencensus/trace/span.py
```diff
@@ -12,20 +12,107 @@
...
+from opencensus.trace import time_event
 from opencensus.trace.span_context import generate_span_id
 from opencensus.trace.tracers import base
 
 
+# https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md  # noqa
+MAX_NUM_ATTRIBUTES = 32
+MAX_NUM_ANNOTATIONS = 32
+MAX_NUM_MESSAGE_EVENTS = 128
+MAX_NUM_LINKS = 32
```
c24t: Hardcoded for now, I'll add them to the global traceconfig in another PR.

opencensus/trace/span.py
```diff
@@ -281,16 +387,26 @@ def format_span_json(span):
...
     if span.attributes:
-        span_json['attributes'] = attributes.Attributes(
+        span_json['attributes'] = attributes_module.Attributes(
             span.attributes).format_attributes_json()
 
     if span.stack_trace is not None:
         span_json['stackTrace'] = span.stack_trace.format_stack_trace_json()
 
-    if span.time_events:
+    formatted_time_events = []
```
songy23: Not sure why we'd want to combine annotations and message events?

opencensus/trace/span_data.py
```diff
@@ -125,10 +129,20 @@ def _format_legacy_span_json(span_data):
         span_json['stackTrace'] = \
             span_data.stack_trace.format_stack_trace_json()
 
-    if span_data.time_events:
+    formatted_time_events = []
```
songy23: Same question here.

opencensus/trace/span.py
c24t: Your guess is as good as mine, I did this to preserve the old behavior.

c24t: Actually `format_span_json` doesn't have any callers, and neither do the other json formatting methods it calls. `format_legacy_trace_json` is only used by the logging and file exporters. I'll remove them in another PR.

opencensus/trace/span.py
```diff
@@ -12,20 +12,113 @@
...
 
 
+# https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md  # noqa
+MAX_NUM_ATTRIBUTES = 32
+MAX_NUM_ANNOTATIONS = 32
+MAX_NUM_MESSAGE_EVENTS = 128
+MAX_NUM_LINKS = 32
+
+
+class BoundedList(Sequence):
```
songy23: Nit: consider splitting these helper data structures to another file.

opencensus/trace/span.py
songy23: SGTM

opencensus/trace/span.py
```diff
@@ -12,20 +12,113 @@
...
+        self.dropped = 0
+        self._dict = OrderedDict()
+
+    def __repr__(self):
+        return repr(self._dict)
+
+    def __getitem__(self, key):
+        return self._dict[key]
+
+    def __setitem__(self, key, value):
```
reyang: Probably worth to put it the comment: Python's built-in dict is thread-safe, here we don't have such promise.

opencensus/trace/span.py
```diff
@@ -12,20 +12,113 @@
...
+
+    def __iter__(self):
+        return iter(self._dict)
+
+    def __len__(self):
+        return len(self._dict)
+
+    @classmethod
+    def from_map(cls, maxlen, mapping):
+        if len(mapping) > maxlen:
```
reyang: Do we want to provide thread-safety for classmethod? If yes, probably consider a shallow copy before determine the size:
```
mapping = OrderedDict(mapping)
if len(mapping) > maxlen:
    ...
```

opencensus/trace/span.py
c24t: Good call, I made this threadsafe instead in 05cfd8e.

opencensus/trace/span.py
c24t: AFAICT that's not exactly threadsafe either since it's still possible to modify `mapping` while `MutableMapping.update` is iterating over it to make the copy, but that's certainly better than it being modified between the length check and the copy.

opencensus/trace/span.py
c24t: I know they're generic, but I left them in here for now since `span` is the only caller.

opencensus/trace/span.py
c24t: The logging and file exporters only incidentally call `format_legacy_trace_json`, the real reason for this odd formatting is stackdriver. It looks like the reason annotations and message events were combined is because that's how they're [reported to stackdriver](https://cloud.google.com/trace/docs/reference/v2/rest/v2/TimeEvents#TimeEvent).

The stackdriver data model leaks out into other classes too. Attributes, links, stack traces, and etc. have formatting methods specific to stackdriver, and we're using these methods to serialize `SpanData`s for the logging and file exporters.

I'm moving the stackdriver-specific formatting into the stackdriver exporter in #651 and adding generic serialization methods for the exporters that need them.

PR #652, opened 2019-05-14
Pin nox-automation, make noxfile explicit

c24t:
Travis builds are [failing](https://travis-ci.org/census-instrumentation/opencensus-python/builds/532095229) because it looks like the [`2019.5.13` `nox-automation` release](https://pypi.org/project/nox-automation/2019.5.13/) also installs `nox`, which gives us the wrong executable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ python:
   - '3.6'
 
 install:
-  - pip install --upgrade nox-automation
+  - pip install nox-automation==0.19.1
 
 script:
-  - nox -s docs
+  - nox -f nox.py -s docs
   - touch docs/.nojekyll
 
 branches:


.travis.yml
```diff
@@ -3,10 +3,10 @@ python:
   - '3.6'
 
 install:
-  - pip install --upgrade nox-automation
+  - pip install nox-automation==0.19.1
```
reyang: Going forward, it seems we only need:

```shell
pip install --upgrade nox
nox -f nox.py -s docs
```

Pin the tooling version has pros and cons, personally I'm leaning toward using the latest version.

reyang:
I guess `nox-automation` has been installing `nox`? This is what we do in the environment setup https://github.com/census-instrumentation/opencensus-python/blob/master/CONTRIBUTING.md#instructions:

```shell
$ pip install nox-automation  # Only first time.
$ nox
```

I suggest that we remove nox-automation, only install nox.

c24t:
> I guess `nox-automation` has been installing `nox`?

It looks like this is new behavior in [`2019.5.13`](https://pypi.org/project/nox-automation/2019.5.13/):

> This package is deprecated and merely installs the nox package instead.

.travis.yml
c24t: I agree, but that does mean making some changes to our tests and CI config. Any objection to merging this to unbreak the build and merging https://github.com/census-instrumentation/opencensus-python/pull/653 when it's working?

PR #653, opened 2019-05-14
Replace nox-automation with nox

reyang:
One possible way for #652. @c24t

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
       - run:
           name: Run tests - opencensus
           command: |
-            nox -f nox.py
+            nox -f noxfile.py
       - deploy:
           name: Push to PyPI (if this is a release tag).
           command: scripts/twine_upload.sh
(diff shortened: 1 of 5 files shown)


c24t:
I agree that this is the better approach (vs #652) in the long term, but it's still causing failing builds now. What do you think about merging #652 now and merging this one once CI is fixed?

reyang:
> I agree that this is the better approach (vs #652) in the long term, but it's still causing failing builds now. What do you think about merging #652 now and merging this one once CI is fixed?

Yes I agree. Let's unblock the CI first using #652.

.circleci/config.yml
```diff
@@ -40,7 +40,8 @@ jobs:
       - run:
           name: Run tests - opencensus
           command: |
-            nox -f nox.py
+            pip install --upgrade nox
```
reyang: This is a workaround to the preinstalled nox. We don't need this once we move to Travis.

reyang:
@c24t CI is fixed.

reyang:
FYI @lzchen

Issue #616, opened 2019-04-12
Logging integration design discussion


reyang:
@c24t wish to know your thoughts as you've done the pilot work in #536.

1. Currently the `use_oc_logging` works for all loggers created after the call. There is another way to work with loggers created before the call (although it could have wide impact), should we consider it (which seems to be useful for automatic capturing unhandled exceptions)?
```
class ContextLogRecord(logging.LogRecord):
    def __init__(self, name, level, pathname, lineno, msg, args, exc_info, func=None, sinfo=None):
        self.trace_id = 'foo'
        self.span_id = 'bar'
        return super(ContextLogRecord, self).__init__(name, level, pathname, lineno, msg, args, exc_info, func, sinfo)

logging.setLogRecordFactory(ContextLogRecord)
```

2. Do we want to move `use_oc_logging` to an ext package and use config_integration?
```
config_integration.trace_integrations(['logging'])
```

reyang:
We have a trace exporter which direct spans to logs `opencensus.trace.logging_exporter.LoggingExporter`.
Need to think about what's the story going forward. (Does it make sense to export spans to logging, then redirect logging to log exporter?)

c24t:
Using `setLogRecordFactory` in `use_oc_logging` sounds fine. Having `TraceLogger` and `TraceLoggingAdapter` is nice because you can still create loggers explicitly. I thought it was a feature -- not a bug -- that we didn't affect loggers created before loading the integration, but I think there's a good argument for either.

> which seems to be useful for automatic capturing unhandled exceptions

How so?

> 1. Do we want to move `use_oc_logging` to an ext package and use config_integration?

So leave `log/__init__.py` where it is and extract `use_oc_logging` into a package? That sounds fine to me, and would be consistent with other integrations. If we do separate it we may want to include formatters in the ext package.

c24t:
> Does it make sense to export spans to logging, then redirect logging to log exporter?

(From IRL conversation)

This seems fine since the logs themselves wouldn't change, the `LogRecord`s would just include the trace and span IDs, which would get dropped during formatting.

[...]

PR #655, opened 2019-05-15
Move logging integration into ext package

reyang:
This is part of the #616 effort.

With this change, the integration is more consistent:

```python
config_integration.trace_integrations(['logging'])
```

### DIFF
--- a/contrib/opencensus-ext-logging/README.rst
+++ b/contrib/opencensus-ext-logging/README.rst
@@ -25,7 +25,7 @@ Usage
     from opencensus.trace.tracer import Tracer
 
     config_integration.trace_integrations(['logging'])
-    logging.basicConfig(format='%(asctime)-15s traceId=%(traceId)s spanId=%(spanId)s %(message)s')
+    logging.basicConfig(format='%(asctime)s traceId=%(traceId)s spanId=%(spanId)s %(message)s')
     tracer = Tracer(sampler=AlwaysOffSampler())
 
     logger = logging.getLogger(__name__)
--- a/contrib/opencensus-ext-logging/README.rst
+++ b/contrib/opencensus-ext-logging/README.rst
@@ -1,0 +1,41 @@
+OpenCensus logging Integration
+============================================================================
+
+|pypi|
+
+.. |pypi| image:: https://badge.fury.io/py/opencensus-ext-logging.svg
+   :target: https://pypi.org/project/opencensus-ext-logging/
+
+Installation
+------------
+
+::
+
+    pip install opencensus-ext-logging
+
+Usage
+-----
+
+.. code:: python
+
+    import logging
+
+    from opencensus.trace import config_integration
+    from opencensus.trace.samplers import AlwaysOffSampler
+    from opencensus.trace.tracer import Tracer
+
+    config_integration.trace_integrations(['logging'])
+    logging.basicConfig(format='%(asctime)-15s traceId=%(traceId)s spanId=%(spanId)s %(message)s')
+    tracer = Tracer(sampler=AlwaysOffSampler())
+
+    logger = logging.getLogger(__name__)
+    logger.warning('Before the span')
+    with tracer.span(name='hello'):
+        logger.warning('In the span')
+    logger.warning('After the span')
+
+
+References
+----------
+
+* `OpenCensus Project <https://opencensus.io/>`_
--- a/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
+++ b/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
@@ -1,0 +1,27 @@
+# Copyright 2017, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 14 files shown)


contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def trace_integration(tracer=None):
+    """Replace the global default logging class with `TraceLogger`.
+
+    Loggers created after the integration will produce `LogRecord`s
+    with extra traceId, spanId, and traceSampled attributes from the opencensus
+    context.
+    """
+    logging.setLoggerClass(TraceLogger)
```
c24t: The whole integration is effectively just this line? I don't know that it's worth asking the user to install a separate package for this.

contrib/opencensus-ext-logging/README.rst
```diff
@@ -0,0 +1,41 @@
...
+.. code:: python
+
+    import logging
+
+    from opencensus.trace import config_integration
+    from opencensus.trace.samplers import AlwaysOffSampler
+    from opencensus.trace.tracer import Tracer
+
+    config_integration.trace_integrations(['logging'])
+    logging.basicConfig(format='%(asctime)-15s traceId=%(traceId)s spanId=%(spanId)s %(message)s')
```
c24t: Why -15s?

contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
reyang: I think eventually we might want to move the logger class into the extension (or consider switching to `setLogRecordFactory`).
The next step would be include formatters as what we've discussed in https://github.com/census-instrumentation/opencensus-python/issues/616#issuecomment-484992593.

contrib/opencensus-ext-logging/README.rst
reyang: Trying to follow the sample code in https://docs.python.org/3.1/library/logging.html#logging.debug.
I guess the reason is to align with some Linux console tools.

reyang:
> Is the goal here to keep big runtime changes to other libraries -- in this case, replacing `logging`'s default logging class -- quarantined in contrib?

Here goes my thinking:
1. Technically the logic here is similar to `opencensus-ext-requests`, we hook into the target library and do some integration with trace context.
2. It would be great if we can keep it consistent with other integrations, so we get benefit of 1) consistency 2) support for Flask/Django/Pyramid via configuration. For example, if we want to have integration with unhandled exceptions or other language features, we don't need to have to introduce another API like `use_oc_exception_handler`.

contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
reyang: Regarding whether it's worth asking the user, I guess we don't need to ask the user:
1. Library developers will just use `opencensus`, with no extensions.
2. Application developers will pick at least one exporter, that exporter can include the logging package (e.g. I would imagine Azure customers to use `pip install opencensus-ext-azure`, which implies `opencensus` and `opencensus-ext-logging`).

c24t: I see, I forgot about that comment. It looks like _I_ wanted to keep big monkey patches like this in contrib.

I think we've got different use cases in mind. I'd expect developers to use OC loggers from application code, either by patching logging globally with `use_oc_logging` or using `TraceLogger` for specific logs they want to capture alongside tracing info. It'd be surprising to get the logging integration from the exporter in this case.

OC logging adds trace metadata to regular logs by getting it from the context. Do we still have `traceId` (and etc.) for a given span in context when we export it? By the time it reaches the exporter this is data, not metadata. I'd expect `traceId` to show up in the body of an exporter log, not as a header-like field as it is here.

contrib/opencensus-ext-logging/README.rst
c24t: That's surprising to me since the other fields aren't padded, but feel free to leave it in if it looks standard to you.

c24t:
Both good reasons, LGTM.

contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
reyang: > I think we've got different use cases in mind. I'd expect developers to use OC loggers from application code, either by patching logging globally with `use_oc_logging` or using `TraceLogger` for specific logs they want to capture alongside tracing info. It'd be surprising to get the logging integration from the exporter in this case.

I think we're on the same page. Application developers will have to ask **enrich my logs with trace_id and span_id** explicitly. Exporters will not provide such functionality. The example I gave is that we don't have application developers to explicitly **install** the extension, as we can do it implicitly.
 
> OC logging adds trace metadata to regular logs by getting it from the context. Do we still have `traceId` (and etc.) for a given span in context when we export it?

No, context is unavailable in the exporter code.

> By the time it reaches the exporter this is data, not metadata. I'd expect `traceId` to show up in the body of an exporter log, not as a header-like field as it is here.

Yes, we're on the same page.

contrib/opencensus-ext-logging/README.rst
reyang: Given it is not the main purpose of the sample code, I've removed the `-15s` to avoid the confusion.

PR #656, opened 2019-05-15
Fix stackdriver k8s label keys

### LINKED ISSUE #647: k8s container attributes not included when exporting to Stackdriver
- Python 3.6.7, opencensus 0.5.0, opencensus-ext-stackdriver 0.3.0
- Instrumenting code running in a GKE Kubernetes cluster

**Steps to reproduce.**
Execute any tracing or stats collection code that exports to Stackdriver - the k8s metadata attributes will be missing from the results.

Results exported to Stackdriver should have associated attribute labels for project id, cluster name, pod name etc.

All k8s attributes are missing from the results I see in Stackdriver.

**Additional context.**
I think there is a bug in `contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/trace_exporter/__init__.py` within `set_monitored_resource_attributes`:
```py
def set_monitored_resource_attributes(span):
    """Set labels to span that can be used for tracing.
    :param span: Span object
    """
    resource = monitored_resource.get_instance()
    if resource is not None:
        resource_type = resource.get_type()
        resource_labels = resource.get_labels()

        if resource_type == 'gke_container':
            resource_type = 'k8s_container'
            ...
```
This code is still checking for resources of type `gke_container`, whereas `opencensus/common/monitored_resource/monitored_resource.py` now uses `k8s_container` when populating the resources to be automatically added to the Stackdriver results. Therefore, the above if block is entirely skipped and no k8s attributes are included.

The same bug appears to be present in the equivalent file for Stackdriver stats exporting (`gke_container` is also present there).

c24t:
This PR updates the expected label keys for kubernetes monitored resources in the stackdriver exporters to match #528 and does some incidental cleanup.

Fixes #647.


contrib/opencensus-ext-stackdriver/tests/test_stackdriver_exporter.py
```diff
@@ -678,12 +678,12 @@ def test_monitored_resource_attributes_gke(self, gmr_mock):
...
-            'pod_id': 'pod',
-            'cluster_name': 'cluster',
-            'namespace_id': 'namespace',
-            'container_name': 'c1',
+            'k8s.io/pod/name': 'pod',
+            'k8s.io/cluster/name': 'cluster',
+            'k8s.io/namespace/name': 'namespace',
+            'k8s.io/container/name': 'c1',
             'project_id': 'my_project',
             'instance_id': 'instance',
```
songy23: Nit: `instance_id` shouldn't be part of `k8s_container`

bogdandrutu: Is instance_id not part of the Host?

songy23: It is, but in Stackdriver definition `k8s_container` resource doesn't have this label. Setting `instance_id` for `k8s_container` will result in errors (e.g https://github.com/census-instrumentation/opencensus-java/issues/1888).

bogdandrutu: ACK

contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py
```diff
@@ -300,62 +303,65 @@ def register_metric_descriptor(self, oc_md):
...
+    resource_type = resource.get_type()
+    resource_labels = resource.get_labels()
+
+    def set_attribute_label(attribute_key, label_key, label_value_prefix=''):
+        """Set a label to timeseries that can be used for monitoring.
+
+        :param series: TimeSeries object based on view data
+        :param resource_labels: collection of labels
+        :param attribute_key: actual label key
+        :param label_key: exporter specific label key, Optional
```
reyang: curious, is the capitalization in `Optional` intentional?

c24t: Nope, it's copypasta from L351.

PR #657, opened 2019-05-20
Skeleton for Azure Log Exporter

reyang:
Part of the #616 effort, using Azure log exporter for the pilot work.

I wish to get early feedback on the design and usage. Currently I haven't added the actual export logic, all the exporter does is to print logs to the STDOUT.

We can explore if it makes sense to move the following classes to opencensus core library:
- `opencensus.ext.azure.log_exporter.BaseLogHandler`
- `opencensus.ext.azure.log_exporter.Worker`

The design principle:
1. Align with Python logging practice, avoid introducing another pattern.
2. The export logic should run asynchronously, instead blocking the hot path.
3. Try to align with the queue/worker direction for trace exporter.

This is what I expect customers to use:
```python
import logging

logger.warning('Hello, World!')
```

A more complex scenario with multiple handlers:
```python
import logging

# create azure log handler
oclh = AzureLogHandler()
oclh.setFormatter(logging.Formatter('%(message)s (%(pathname)s:L%(lineno)s)'))
logger.addHandler(oclh)

# create console log handler
ch = logging.StreamHandler()
ch.setFormatter(logging.Formatter('%(asctime)s %(message)s'))
logger.addHandler(ch)

# create stackdriver log handler
# sdlh = StackdriverLogHandler()
# logger.addHandler(sdlh)

logger.warning('Hello, World!')
```

@c24t @lzchen


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+                trace_id = getattr(item, 'traceId', 'N/A')
+                span_id = getattr(item, 'spanId', 'N/A')
+                print('{levelname} {trace_id} {span_id} {pathname}:L{lineno} {msg}'.format(trace_id=trace_id, span_id=span_id, **vars(item)))
+        if event:
+            event.set()
+
+
+class LogHandler(logging.Handler):
+    def __init__(self, exporter):
+        logging.Handler.__init__(self)
```
c24t: Why not `super`?

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    :param options: Options for the exporter. Defaults to None.
+    """
+
+    def __init__(self, **options):
+        self.options = Options(**options)
+        if not self.options.instrumentation_key:
+            raise ValueError('The instrumentation_key is not provided.')
+        super(AzureExporter, self).__init__(**options)
+
+    def emit(self, batch, event=None):
```
c24t: It looks like this relies on `BaseExporter` having `emit` and `export`. I thought the goal was to remove `emit` as in https://github.com/census-instrumentation/opencensus-python/pull/642#discussion_r281860719.

I don't think we should plan to move `BaseExporter` into the core library in its current state.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        if batch:
+            for item in batch:
+                trace_id = getattr(item, 'traceId', 'N/A')
+                span_id = getattr(item, 'spanId', 'N/A')
+                print('{levelname} {trace_id} {span_id} {pathname}:L{lineno} {msg}'.format(trace_id=trace_id, span_id=span_id, **vars(item)))
```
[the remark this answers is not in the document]
reyang: Yep, we're on the same page.
We don't expect the user to go through OpenCensus in order to print/format logs, users will just have to follow standard Python approach.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
reyang: `super` won't work in this case (I haven't looked into why), (and this is probably why) all the built-in handlers are using `logging.Handler.__init__(self)`.
I'll take a closer look tomorrow and update here.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
reyang: That's right, we only want to keep `export`, and remove `emit`. And we also discussed that probably we don't even need a `BaseExporter` at all.

In this PR I try to focus on making the log exporter work end-to-end, without doing too much refactor.

reyang:
> Exposing the exporter as a log handler and having the exporter accept `LogRecord`s seem like good choices to me.

After a bit more exploration, I think it makes more sense to combine the handler and exporter logic, this aligns better with Python practice (e.g. having formatter associated with handler). Each handler will have its own queue.
 
> This looks good as WIP, but I think there are still API changes to make (remove `emit`, consider losing the shared event, etc.) before moving other classes out of the contrib package.

Yes, I agree. This is the direction we're moving towards.

> We should also consider adding logging (vs. just log correlation) to the spec if we're going to expose a general purpose log exporter API.

After spending more time playing around logging, I think it makes more sense to align logging APIs with language/runtime instead of trying to align across SDKs, thoughts? @c24t

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
reyang: Problem solved for logs (traces will take more effort, probably in a separate PR).

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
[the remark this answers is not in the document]
reyang: Here goes the proposal:

```python
import logging

# create azure log handler
oclh = AzureLogHandler()
oclh.setFormatter(logging.Formatter('%(message)s (%(pathname)s:L%(lineno)s)'))
logger.addHandler(oclh)

# create console log handler
ch = logging.StreamHandler()
ch.setFormatter(logging.Formatter('%(asctime)s %(message)s'))
logger.addHandler(ch)

# create stackdriver log handler
# sdlh = StackdriverLogHandler()
# logger.addHandler(sdlh)

logger.warning('Hello, World!')
```

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+    def flush(self, timeout=None):
+        self._queue.flush(timeout=timeout)
+
+
+class Worker(threading.Thread):
+    daemon = True
+
+    def __init__(self, src, dst):
+        self._src = src
+        self._dst = dst
```
c24t: Note the circular reference here, we've been bitten by this before trying to clean up exporters. See https://github.com/census-instrumentation/opencensus-python/blob/04c111f99dc122a079b589942b61d22c23979afa/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/stats_exporter/__init__.py#L372-L406.

reyang: My original understanding is that we have `Handler.close` which can be used explicitly or implicitly (by Python logging library), and the circular reference will be collected by the GC. This shouldn't cause problem? Is this for preventing memory leak, or to reduce GC overhead?

Regarding memory leak, I run the following app for an hour and see a flat memory usage:
```
import logging

while True:
    handler = AzureLogHandler()
    logger.addHandler(handler)
    logger.warning('Hello, World!')
    logger.removeHandler(handler)
    handler.close()
```

For GC overhead, given the circle is pretty small, I guess there is no noticeable difference?

c24t: > the circular reference will be collected by the GC. This shouldn't cause problem?

I thought there were more general problems cleaning up circular references in python, but it may only be a problem for classes that define `__del__` (see https://stackoverflow.com/a/2428888), in which case this is fine.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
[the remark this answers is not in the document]
reyang: > Simplifying this LGTM, but later we may still want to
bas> 
> * revisit the worker/queue classes as per [#642 (comment)](https://github.com/census-instrumentation/opencensus-python/pull/642#discussion_r283962128) and related comments

Yep, I'll merge this PR as is. And work on the next PR to have the actual Azure log exporter logic (currently it is just console print) today. Starting from next week, I'll have time to revisit/refactor worker/queue out.

> * consider separating the exporter from the log handler when we e.g. send logs to the agent

We can explore the following options:
1. Align across OC/OT SDKs - having an explicit exporter concept across all languages.
2. Align more with language convention (e.g. log appender for Java, log handler for Python), and provide base/helper class to support configuration/policy.

PR #658, opened 2019-05-21
Integrate travis ci

lzchen:
As per #402, we want to use TravisCI for our CI pipeline instead of CircleCI.

Currently it mimics the CI steps of CircleCI exactly except system tests and deployment.
In my opinion, we should get this merged and see if most builds pass the new CI. Then we can implement the deployment step once we see that it is stable and gradually phase out CircleCI.

Another point to note is TravisCI runs all Python environments separately so with the current config, the coverage data covers each env individually. I believe it is more worthwhile to get TravisCI implemented before spending time on researching how to get aggregated code coverage.

@c24t @reyang


tox.ini
```diff
@@ -0,0 +1,55 @@
...
+  py36-lint: flake8
+  py36-setup: docutils
+  py36-setup: pygments
+  py{27,34,35,36}-cover: coverage
+  py{27,34,35,36}-cover: pytest-cov
+  py36-docs: setuptools >= 36.4.0
+  py36-docs: sphinx >= 1.6.3
+
+commands = 
+  py{27,34,35,36}-unit: py.test --quiet --cov={envsitepackagesdir}/opencensus --cov=context --cov=contrib --cov-report= --cov-config=.coveragerc --cov-fail-under=97 tests/unit/ context/ contrib/
```
reyang: Instead of using the destination folder `{envsitepackagesdir}/opencensus`, can we use the source folder?
The reason for this asking is to avoid run coverage with packages that don't belong to this repo but still under `opencensus` namespace (e.g. opencensus-proto, which is being covered by the coverage report).

reyang:
Great to see this!
Could you describe how to run test locally with tox?

tox.ini
```diff
@@ -0,0 +1,55 @@
...
+  ; TODO: System tests
+  py36-lint: flake8 context/ contrib/ opencensus/ tests/ examples/
+  py36-setup: python setup.py check --restructuredtext --strict
+  ; TODO: aggregate coverage data
```
c24t: It seems fine to me not to aggregate the coverage. We probably want to know if tests don't have full coverage under a specific version of python.

We might have to annotate conditional imports to skip the check, but any other reason not to run coverage separately for each version?

.travis.yml
```diff
@@ -1,12 +1,16 @@
...
+  - '3.5'
   - '3.6'
 
 install:
-  - pip install --upgrade nox
+  - pip install tox-travis
 
 script:
-  - nox -f noxfile.py -s docs
+  # - nox -f noxfile.py -s docs
```
c24t: Please open a ticket or leave a TODO instead of the commented line for docs here.

.travis.yml
```diff
@@ -1,12 +1,16 @@
 language: python
 python:
+  - '2.7'
+  - '3.4'
+  - '3.5'
   - '3.6'
```
c24t: What's the story on 3.7 (and maybe [3.8!](https://docs.python.org/3.8/whatsnew/3.8.html)) support on travis?

lzchen: 3.7 will be supported in another ticket. This one is simply just to introduce the migration to travis CI.

lzchen:
How to run test locally (as per @reyang )
1. Install tox: pip install --upgrade tox
2. Install all versions of python that are supported (currently 2.7, 3.4,3.5 and 3.6)
3. Run "tox -e py{27,34,35,36}-unit" to run all unit tests

tox.ini
lzchen: I am okay with not aggregating the coverage. Specific version coverage seems more important at the moment :)

tox.ini
lzchen: This is a good point. I will change this to use {envdir} instead to point to the exact source folder in the virtual environment

lzchen: This is a good point. I will use {envdir} instead to point to the virtual environment source directory instead.

.travis.yml
lzchen: Docs step is already included. Removing comment.

lzchen:
Seems like the circle ci status isn't being reflected in Github.
It might be due to protection settings for the master branch?
https://circleci.com/docs/2.0/workflows-waiting-status/

@c24t 
@reyang

reyang:
Would you try to remove the circle ci from your fork and see if it mitigates the problem?

image

reyang:
This PR has been superseded by #662.

PR #659, opened 2019-05-23
Use httplib's host property instead of urllib3's _dns_host

tpyo:
The httplib extension references `_dns_host` which is typically used with urllib3, not httplib.  This means `HTTPConnection` mixins that do not implement a `_dns_host` property will fail with an `AttributeError` (encountered this with [boto](https://github.com/boto/botocore/blob/14e0eab5c1e4aec437c3e558e6899de00fd5e98e/botocore/awsrequest.py#L98)).  

From the [urllib3 source code](https://github.com/urllib3/urllib3/blob/master/src/urllib3/connection.py#L137-L144):
> We assume that only urllib3 uses the `_dns_host` attribute; httplib itself only uses `host`, and it seems reasonable that other libraries follow suit.

The urllib3 implementation provides [getters and setters](https://github.com/urllib3/urllib3/blob/master/src/urllib3/connection.py#L117-L144) for `host` that use `_dns_host`, so switching to `host` should have no effect on working integrations.


reyang:
Thanks for your contribution @tpyo!

PR #661, opened 2019-05-23
Test Circle CI trigger

reyang:
It is strange that @lzchen's PR won't trigger Circle CI.
Trying to send the same change from my own fork and see if it works.


reyang:
@lzchen this one seems to be working.

PR #449, opened 2019-01-15
Remove well_known_types.Error and well_known_types.ParseError.

songy23:

PR #667, opened 2019-05-28
Pylint for linting

lzchen:
Add pylint functionality for linting. 
We keep flake8 for now for even more robust linting checks.

Pylint has no support for linting all files within a certain directory (https://github.com/PyCQA/pylint/issues/352), so one workaround is to use a script to output all ".py" files as arguments. Adjustments can be made to ignore future .py files within those directories.

I also made it so that the return code from pylint is successful so it does not fail the build. Currently, there are hundreds of lint errors across our files if pylint is introduced. Any suggestions on how to enforce standards for the future but also fix our legacy code? I can begin to fix them module by module if need be.

@c24t  @reyang

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -19,7 +19,7 @@ ignore-patterns=
 
 # Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
 # number of processors available to use.
-jobs=1
+jobs=0
 
 # Control the amount of potential inferred values when inferring a single
 # object. This can help the performance when dealing with large functions or
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,563 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Specify a configuration file.
+#rcfile=
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED.
+confidence=
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=print-statement,
+        parameter-unpacking,
+        unpacking-in-except,
+        # old-raise-syntax,
+        # backtick,
+        long-suffix,
+        old-ne-operator,
+        old-octal-literal,
+        import-star-module-level,
+        non-ascii-bytes-literal,
+        raw-checker-failed,
+        bad-inline-option,
+        locally-disabled,
+        file-ignored,
+        suppressed-message,
+        useless-suppression,
+        deprecated-pragma,
+        use-symbolic-message-instead,
+        apply-builtin,
+        basestring-builtin,
+        buffer-builtin,
+        cmp-builtin,
+        coerce-builtin,
+        execfile-builtin,
+        file-builtin,
+        long-builtin,
+        raw_input-builtin,
+        reduce-builtin,
+        standarderror-builtin,
+        unicode-builtin,
+        xrange-builtin,
+        coerce-method,
+        delslice-method,
+        getslice-method,
+        setslice-method,
+        no-absolute-import,
+        old-division,
+        dict-iter-method,
+        dict-view-method,
+        next-method-called,
+        metaclass-assignment,
+        indexing-exception,
+        raising-string,
+        reload-builtin,
+        oct-method,
+        hex-method,
+        nonzero-method,
+        cmp-method,
+        input-builtin,
+        round-builtin,
+        intern-builtin,
+        unichr-builtin,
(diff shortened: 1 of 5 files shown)


lzchen:
It seems builds are failing due to Google making a release last night to their protocol buffers library. The ParseError type is missing from well_known_types.py.
https://github.com/protocolbuffers/protobuf/blob/v3.8.0/python/google/protobuf/internal/well_known_types.py

songy23:
@lzchen https://github.com/census-instrumentation/opencensus-python/pull/449 should resolve the failure.

tox.ini
```diff
@@ -47,6 +47,7 @@ commands =
   py{27,34,35,36,37}-unit: py.test --quiet --cov={envdir}/opencensus --cov=context --cov=contrib --cov-report= --cov-config=.coveragerc --cov-fail-under=97 tests/unit/ context/ contrib/
   ; TODO: System tests
   py37-lint: flake8 context/ contrib/ opencensus/ tests/ examples/
+  py37-lint: bash ./scripts/pylint.sh
```
reyang: This assumes all developers are using Linux/Mac?

reyang: Please refer to #361, where metrics didn't work on Windows.
Having the CI running on both Linux and Windows could help to catch such issues.

lzchen: I don't expect developers to run the lint environment directly from their machines. 
Travis default environment is Ubuntu Linux so the scripts should run under builds as long as we don't change the vm. 
https://docs.travis-ci.com/user/reference/overview/

reyang: Would you update https://github.com/census-instrumentation/opencensus-python/blob/master/CONTRIBUTING.md#instructions in this PR to show how local test would work?

CONTRIBUTING.md
```diff
@@ -37,8 +37,10 @@ $ git remote add fork https://github.com/YOUR_GITHUB_USERNAME/opencensus-python.
 Run tests:
 
 ```sh
-$ pip install nox  # Only first time.
-$ nox
+# Make sure you have all supported versions of Python installed
+# Currently these are 2.7, 3.4, 3.5, 3.6 and 3.7
+$ pip install tox  # Only first time.
+$ tox -e py27-unit,py34-unit,p35-unit,py36-unit,py37-unit # unit tests
```
reyang: This has different semantic, previously local `nox` would run lint and now we intend to remove it?

c24t:
FWIW the default pylint checker set is probably much stricter than we actually want. I've disabled some checkers in my `.pylintrc` (see https://gist.github.com/c24t/bc71a75bf33d4ab0e5754166ae7befeb), but even with this config still see many spurious lint errors.

lzchen:
@c24t 
Yeah I agree that the pylint checker is quite strict. Adding the .pylintrc file actually makes the code have a negative rating by pylint. Do we even want to follow this rating? How should we handle the legacy code?

.pylintrc
```diff
@@ -0,0 +1,563 @@
...
+
+# Minimum number of public methods for a class (see R0903).
+min-public-methods=2
+
+
+[EXCEPTIONS]
+
+# Exceptions that will emit a warning when being caught. Defaults to
+# "Exception".
+overgeneral-exceptions=Exception
```
reyang: nit: add a new line here

reyang:
Here goes my suggestion, which takes three steps:
1. Add pylint and start with a reasonable rule set (we can use @c24t's as a starting point). Instead of blocking the CI on lint failure, show warning instead.
2. Fix existing lint error, make the CI warning free.
3. Enforce lint rule, block CI if lint failed.

.pylintrc
```diff
@@ -0,0 +1,563 @@
...
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
```
reyang: seems 0 is better for us?

lzchen:
> Here goes my suggestion, which takes three steps:
> 
> 1. Add pylint and start with a reasonable rule set (we can use @c24t's as a starting point). Instead of blocking the CI on lint failure, show warning instead.
> 2. Fix existing lint error, make the CI warning free.
> 3. Enforce lint rule, block CI if lint failed.

lzchen:
@reyang  Okay we can start number 1 in this PR. I believe the fixing of the existing warnings needs to happen iteratively in other PRs.

c24t:
Another option is to add the linter to https://github.com/open-telemetry/opentelemetry-python instead. It may not be worth the effort to change all the files in this repo to satisfy pylint. It'll be easier to start from scratch and fix each section of the code separately as we move it into the new repo.

lzchen:
@c24t 
I think we can start off by applying the linting rules to a smaller subset of the files in the repo, most noticeably the core-sdk. There might be different lint requirements and logic for the external libraries.
That way we can iteratively fix the lint warnings.

scripts/pylint.sh
```diff
@@ -0,0 +1,32 @@
...
+function pylint_dir {
+  python -m pip install --upgrade pylint
+  find context/ -type f -name "*.py" > output
+  find contrib/ -type f -name "*.py" >> output
+  find opencensus/ -type f -name "*.py" >> output
+  find tests/ -type f -name "*.py" >> output
+  find examples/ -type f -name "*.py" >> output
+  # TODO fix lint errors
+  cat output | xargs pylint || true # ignores errors
+  rm -rf output
```
c24t: I know I'd be unhappy if running lint clobbered my files. Why not use a tempfile or do something like this?

`pylint $(find context/ contrib opencensus/ tests/ examples/ -type f -name "*.py")`

tox.ini
c24t: @reyang what do you recommend for cross-platform scripts? I agree that our tox targets shouldn't assume the platform is linux.

reyang: I would write a small Python script in this case.

scripts/pylint.sh
lzchen: I was looking for something like this. Clearly you are the bash expert :)

lzchen:
@reyang 
Regarding your question about having the CI run on both windows. To clarify, you mean running tox on developers' local machines right? If so, I agree that we should support this. I think for now we should focus on migrating the CI from CircleCI. We can create another pr to change OS specific logic (like docs)

reyang:
> Regarding your question about having the CI run on both windows. To clarify, you mean running tox on developers' local machines right?

Yes, we don't want to give contributors an impression that OpenCensus only works on Linux environments.

> If so, I agree that we should support this. I think for now we should focus on migrating the CI from CircleCI. We can create another pr to change OS specific logic (like docs)

This works.

lzchen:
@reyang @c24t 
Any more reviews for this?

CONTRIBUTING.md
lzchen: Changed to just use tox

tox.ini
lzchen: Yes that's a good idea. Will make a separate PR to address this issue.

reyang:
> Any more reviews for this?

With this PR, pylint would dump a lot of warning messages and we won't block the CI when there is any lint issue. We will probably need to lower the bar and make it blocking (on lint error) before we can migrate flake8.

Not meant to block this PR, @lzchen you make the call.

lzchen:
> > Any more reviews for this?
> 
> With this PR, pylint would dump a lot of warning messages and we won't block the CI when there is any lint issue. We will probably need to lower the bar and make it blocking (on lint error) before we can migrate flake8.
> 
> Not meant to block this PR, @lzchen you make the call.

Flake8 is still in place. We can merge this PR to introduce Pylint without any of it blocking the build as it is (only flake8 will block the build). We can gradually lower the bar in subsequent PRs.

CONTRIBUTING.md
```diff
@@ -37,8 +37,10 @@ $ git remote add fork https://github.com/YOUR_GITHUB_USERNAME/opencensus-python.
 Run tests:
 
 ```sh
-$ pip install nox  # Only first time.
-$ nox
+# Make sure you have all supported versions of Python installed
+# Currently these are 2.7, 3.4, 3.5, 3.6 and 3.7
```
reyang: Probably no need to mention that all supported versions of Python again in the since we've already mentioned it in the "Prerequisites".

PR #668, opened 2019-05-30
Azure log exporter

reyang:
1. Implemented the actual log conversion and exporting logic.
2. Moved the network/storage transmission logic from trace exporter to a shared mixin, applied to both traces and logs.
3. Avoid common names (`Object` and `prototype`), use `BaseObject` and `_default` instead. (Addressing the comments left in https://github.com/census-instrumentation/opencensus-python/pull/613#issuecomment-483432780).

### DIFF
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 - Added log exporter
-  ([#657](https://github.com/census-instrumentation/opencensus-python/pull/657))
-  ([#668](https://github.com/census-instrumentation/opencensus-python/pull/668))
+  ([#657](https://github.com/census-instrumentation/opencensus-python/pull/657),
+  [#668](https://github.com/census-instrumentation/opencensus-python/pull/668))
 - Added persistent storage support
   ([#640](https://github.com/census-instrumentation/opencensus-python/pull/640))
 - Changed AzureExporter constructor signature to use kwargs
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 - Added log exporter
   ([#657](https://github.com/census-instrumentation/opencensus-python/pull/657))
+  ([#668](https://github.com/census-instrumentation/opencensus-python/pull/668))
 - Added persistent storage support
   ([#640](https://github.com/census-instrumentation/opencensus-python/pull/640))
 - Changed AzureExporter constructor signature to use kwargs
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
@@ -36,10 +36,10 @@ class BaseObject(dict):
     def __getattr__(self, name):
         try:
             return self[name]
-        except KeyError as ex:
+        except KeyError:
             raise AttributeError("'{}' object has no attribute {}".format(
                 type(self).__name__,
-                ex,
+                name,
             ))
 
     def __getitem__(self, key):
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 
-class Object(dict):
+class BaseObject(dict):
     def __init__(self, *args, **kwargs):
-        super(Object, self).__init__(*args, **kwargs)
+        super(BaseObject, self).__init__(*args, **kwargs)
         for key in kwargs:
             self[key] = kwargs[key]
 
@@ -25,9 +25,9 @@ class Object(dict):
             for item in self.items():
                 if item[0] not in tmp:
                     tmp[item[0]] = item[1]
-            if self.prototype == self:
+            if self._default == self:
                 break
-            self = self.prototype
+            self = self._default
         return repr(tmp)
 
     def __setattr__(self, name, value):
@@ -43,18 +43,18 @@ class Object(dict):
             ))
 
     def __getitem__(self, key):
-        if self.prototype is self:
-            return super(Object, self).__getitem__(key)
+        if self._default is self:
+            return super(BaseObject, self).__getitem__(key)
         if key in self:
-            return super(Object, self).__getitem__(key)
-        return self.prototype[key]
+            return super(BaseObject, self).__getitem__(key)
+        return self._default[key]
 
 
-Object.prototype = Object()
+BaseObject._default = BaseObject()
 
 
-class Data(Object):
-    prototype = Object(
+class Data(BaseObject):
+    _default = BaseObject(
         baseData=None,
         baseType=None,
     )
@@ -65,8 +65,8 @@ class Data(Object):
         self.baseType = self.baseType
 
 
-class Envelope(Object):
-    prototype = Object(
+class Envelope(BaseObject):
+    _default = BaseObject(
         ver=1,
         name='',
         time='',
@@ -84,8 +84,8 @@ class Envelope(Object):
         self.time = self.time
 
 
-class Event(Object):
-    prototype = Object(
+class Event(BaseObject):
+    _default = BaseObject(
         ver=2,
         name='',
         properties=None,
@@ -98,8 +98,8 @@ class Event(Object):
         self.name = self.name
 
 
-class Message(Object):
-    prototype = Object(
+class Message(BaseObject):
+    _default = BaseObject(
         ver=2,
         message='',
         severityLevel=None,
@@ -113,8 +113,8 @@ class Message(Object):
         self.message = self.message
 
 
-class RemoteDependency(Object):
-    prototype = Object(
+class RemoteDependency(BaseObject):
+    _default = BaseObject(
         ver=2,
         name='',
         id='',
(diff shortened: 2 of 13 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -97,7 +106,7 @@ def stop(self, timeout=None):
             return time.time() - start_time  # time taken to stop
 
 
-class AzureLogHandler(BaseLogHandler):
+class AzureLogHandler(TransportMixin, BaseLogHandler):
```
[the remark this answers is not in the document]
reyang: You're right, that's our goal. For more info please take a look at #642 #657.

Today the trace exporter has legacy issue that needs to be refactored (e.g. trace exporter has `export` method which sends data to the queue, and `emit` which actually export the data. what we want is to only have `export` method, move queue and worker to the core opencensus library).

In this logging effort, we're trying to have a cleaner design, and later we'll refactor traces to follow this design.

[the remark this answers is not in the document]
reyang: Here I took the mixin approach, so the log handler owns all the resources, transport mixin just provides the method on top of these resources.

reyang:
@c24t @songy23 need your help to review, thanks!

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -0,0 +1,133 @@
...
+            text = response.text
+        except Exception as ex:
+            logger.warning('Error while reading response body %s.', ex)
+        else:
+            try:
+                data = json.loads(text)
+            except Exception:
+                pass
+        if response.status_code == 200:
+            logger.info('Transmission succeeded: %s.', text)
```
songy23: Nit: I feel there's a bit too much logs in this class. Maybe consider reducing logs or using traces with a low sampling rate.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -0,0 +1,133 @@
...
+    def _transmit(self, envelopes):
+        """
+        Transmit the data envelopes to the ingestion service.
+        Return a negative value for partial success or non-retryable failure.
+        Return 0 if all envelopes have been successfully ingested.
+        Return the next retry time in seconds for retryable failure.
+        This function should never throw exception.
+        """
+        # TODO: prevent requests being tracked
+        blacklist_hostnames = execution_context.get_opencensus_attr(
```
songy23: Not sure if it's a good idea to put this in the trace `execution_context`? /cc @c24t WDYT?

reyang: Long term goal is to avoid exporters to access execution_context directly, for more info, refer to #573.
In this PR, the code was extracted from the existing trace exporter.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
reyang: This is at info level, which is intended for debugging only and off by default.
The default log level is WARNING.

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
```diff
@@ -0,0 +1,130 @@
...
+        handler = log_exporter.AzureLogHandler(
+            instrumentation_key='12345678-1234-5678-abcd-12345678abcd',
+            storage_path=os.path.join(TEST_FOLDER, self.id()),
+        )
+        logger.addHandler(handler)
+        try:
+            return 1 / 0  # generate a ZeroDivisionError
+        except Exception:
+            logger.exception('Captured an exception.')
+        handler.close()
```
c24t: Isn't the point of the test to check that you got the stack trace here?

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_export_empty(self, request_mock):
+        handler = log_exporter.AzureLogHandler(
+            instrumentation_key='12345678-1234-5678-abcd-12345678abcd',
+            storage_path=os.path.join(TEST_FOLDER, self.id()),
+        )
+        handler._export([])
+        self.assertEqual(len(os.listdir(handler.storage.path)), 0)
+        handler.close()
+
+    @mock.patch('opencensus.ext.azure.log_exporter.AzureLogHandler.log_record_to_envelope')  # noqa: E501
```
c24t: Nit, feel free to ignore if you disagree:

```suggestion
    @mock.patch('opencensus.ext.azure.log_exporter'
                '.AzureLogHandler.log_record_to_envelope')

```

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_export_failure(self, log_record_to_envelope_mock):
+        log_record_to_envelope_mock.return_value = ['bar']
+        handler = log_exporter.AzureLogHandler(
+            instrumentation_key='12345678-1234-5678-abcd-12345678abcd',
+            storage_path=os.path.join(TEST_FOLDER, self.id()),
+        )
+        with mock.patch('opencensus.ext.azure.log_exporter.AzureLogHandler._transmit') as transmit:  # noqa: E501
```
c24t: Same thing

```suggestion
        with mock.patch('opencensus.ext.azure.log_exporter'
                        '.AzureLogHandler._transmit') as transmit:
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -0,0 +1,133 @@
...
+            'blacklist_hostnames',
+        )
+        execution_context.set_opencensus_attr(
+            'blacklist_hostnames',
+            ['dc.services.visualstudio.com'],
```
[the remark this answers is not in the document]
reyang: This will go away when we don't have exporters explicitly maintain the blacklist.

contrib/opencensus-ext-azure/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 - Added log exporter
   ([#657](https://github.com/census-instrumentation/opencensus-python/pull/657))
+  ([#668](https://github.com/census-instrumentation/opencensus-python/pull/668))
```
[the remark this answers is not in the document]
reyang: Thanks for catching this!

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
reyang: Taken, thanks.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -0,0 +1,133 @@
...
+
+from opencensus.trace import execution_context
+
+logger = logging.getLogger(__name__)
+
+
+class TransportMixin(object):
+    def _transmit_from_storage(self):
+        for blob in self.storage.gets():
+            if blob.lease(self.options.timeout + 5):
```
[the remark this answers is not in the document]
reyang: Good feedback. I've added comments here.

The purpose is to reduce the chance of blob lease timeout during transmission (`options.timeout` is for networking timeout, so we added a few seconds here). This would reduce the chance where blob is being sent to the network, while lease timed out, and blob is picked up and sent again by another process (which has extra cost and introduce duplicated data).

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
```diff
@@ -25,9 +25,9 @@ def __repr__(self):
             for item in self.items():
                 if item[0] not in tmp:
                     tmp[item[0]] = item[1]
-            if self.prototype == self:
+            if self._default == self:
                 break
-            self = self.prototype
+            self = self._default
```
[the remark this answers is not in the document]
reyang: This is trying to mimic the JavaScript prototype behavior, traveling through the prototype chain and get all the key-value pairs.
Do you think assigning to `self` is evil and we should have another variable?

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
reyang: I added one assert to verify the exception type name shows up in the post body.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
reyang: I'll try to address it in a separate PR, there are similar scenarios which we need a consistent solution:
1. How to prevent traces generated by trace exporters (e.g. during sending data via HTTP) being sent to trace exporter and cause infinite loop.
2. How to prevent logs/exceptions generated by log exporters/handlers being emitted to the same log handlers and cause infinite loop.

My current thinking is to have the entire worker thread marked as "do not track", so all the spans/logs will have a special tag, and based on this tag we can filter these data out before sending them to exporters.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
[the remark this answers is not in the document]
c24t: Maybe not evil, but certainly confusing.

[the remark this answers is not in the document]
reyang: Got it, will give a better name in the next PR.

PR #670, opened 2019-06-01
Prepare for 0.6.0 release

reyang:


contrib/opencensus-ext-grpc/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.3.dev0'
+__version__ = '0.3'
```
songy23: `0.3.0`?

context/opencensus-context/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.2.dev0'
+__version__ = '0.1.0'
```
songy23: Should this be `0.2.0` or `0.1.1`?

contrib/opencensus-correlation/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '0.2.dev0'
+__version__ = '0.1.0'
```
songy23: Same here.

context/opencensus-context/version.py
reyang: Current version is 0.1.0, and there is no change, so we don't need a new release.

contrib/opencensus-ext-grpc/version.py
reyang: You're right, fixed. Thanks!

contrib/opencensus-correlation/version.py
reyang: There is no change, probably no need to release this package.

reyang: Just noticed this one is special, the version should be 0.3.0 although it hasn't been changed at all.

reyang: Changed to 0.3.0.

PR #680, opened 2019-06-07
add compatibility check badges to README

ylil93:
Hello google package maintainer,
The badges being added to the README in this PR will indicate your compatibility with other google packages. This addresses a set of user bugs which have happened when a user depends on two Cloud libraries (or runtimes that bundle libraries), say A and B, which both depend on library C. If the two libraries require different versions of C, the users can run into issues both when they pip install the libraries, and when they deploy their code. Our compatibility server checks that all libraries we make including this one are self and pairwise compatible as well as not having any deprecated dependencies. The two badges will mark the build for your project green when the latest version available on PyPI and github HEAD respectively meet all compatibility checks with itself and all other libraries. The badge target will link to a details page that elaborates on the current status. This should help you fix issues pre-release, to avoid user surprises. For more information, please take a look at our project charter at go/python-cloud-dependencies-project-charter and the badging PRD https://docs.google.com/document/d/1GYRFrfUou2ssY71AtnLkc8Sg1SD4dxqN4GzlatGHHyI/edit?ts=5c6f031d

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,6 +4,8 @@ OpenCensus - A stats collection and distributed tracing framework
 |gitter|
 |circleci|
 |pypi|
+|compat_check_pypi|
+|compat_check_github|
 
 .. |circleci| image:: https://circleci.com/gh/census-instrumentation/opencensus-python.svg?style=shield
    :target: https://circleci.com/gh/census-instrumentation/opencensus-python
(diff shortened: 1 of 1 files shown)


reyang:
@ylil93 would you suggest that we add this badge to other packages under `contrib` folder?

ylil93:
@reyang Yes that would be great! I will go ahead and add them to our whitelist.

ylil93:
@reyang 

TLDR: Actually it seems like the extensions won't be needing to undergo compatibility checks after all.

Since all of the extension packages are installing opencensus package plus the extension dependency, (eg django for opencensus-ext-django), it's probable that these packages would lag behind the latest versions of these dependencies by several months before being able to migrate support as in the case of django for example, it would need to look deep into the middleware implementation in order to fulfill trace functionality. And the same thing is probably true of other integrations. Given this, every other google python package's compatibility status would be affected if we required the extension packages to undergo the same compatibility testing, this could wind up blocking the releases of those packages. To avoid this, we should only have the main opencensus package undergo compatibility checks and display the compatibility badges.

reyang:
@ylil93 I see your point, thanks for the elaboration.

reyang:
@ylil93 I noticed that one badge seems to be broken, would you help to shed some lights? Thanks.

image

ylil93:
@reyang 
This seems to be a well known issue with shields.io (which we use to build the badges) and github caching: https://github.com/badges/shields/issues/1568

I believe the direct link to the image itself works:
https://python-compatibility-tools.appspot.com/one_badge_image?package=git%2Bgit%3A//github.com/census-instrumentation/opencensus-python.git

reyang:
@ylil93 got it, thanks!

Issue #677, opened 2019-06-05
opencensus.ext.grpc has no "trace" module


allenjohnashton:
**Describe your environment.** Python 2.7

```  'opencensus==0.6.0' \                                                                                                                                                                                                                                                              
  'opencensus-ext-flask==0.3.1' \                                                                                                                                                                                                                                                     
  'opencensus-ext-gevent==0.1.0' \                                                                                                                                                                                                                                                    
  'opencensus-ext-google-cloud-clientlibs==0.1.2' \                                                                                                                                                                                                                                   
  'opencensus-ext-grpc==0.3.0' \                                                                                                                                                                                                                                                      
  'opencensus-ext-logging==0.1.0' \                                                                                                                                                                                                                                                   
  'opencensus-ext-postgresql==0.1.2' \                                                                                                                                                                                                                                                
  'opencensus-ext-stackdriver==0.4.0' \                                                                                                                                                                                                                                               
  'opencensus-ext-requests==0.1.2' \                                                                                                                                                                                                                                                  
  'opencensus-ext-sqlalchemy==0.1.2' \                                                                                                                                                                                                                                                
  'opencensus-ext-threading==0.1.2' \  ```

**Steps to reproduce.**
Follow instructions for trace integration.

**What is the expected behavior?**
Trace integration working

**What is the actual behavior?**
 trace_sampled=False Failed to integrate module: opencensus.ext.grpc.trace
No module named trace

**Additional context.**
Looking at other modules I see a trace.py.  Looks like grpc does not have one for now.

reyang:
This seems to be a doc issue, the grpc integration is done by explicitly using the interceptors (refer to the examples) instead of through `config_integration`.

c24t:
@allenjohnashton which instructions are you following? It'd help if you could post the code that gives you the error.

allenjohnashton:
@c24t Sorry I did not update because @reyang  was correct.  The docs were incorrect.  I see he has linked cl so I am assuming no further clarification was needed.

PR #687, opened 2019-06-17
Fix gRPC doc issue

reyang:
Fixing doc issue discovered in #677.

PR #688, opened 2019-06-17
gRPC integration docs usage fix

c24t:
This updates the gRPC readme to remove an outdated usage example. See #677.

@reyang let me know if you'd rather inline some of the hello world example than remove the usage section completely.


c24t:
Closing in favor of #687.

PR #693, opened 2019-06-21
Implement Azure Metrics Exporter

lzchen:
This PR implements the Metrics Exporter for Azure Monitor. The underlying logic uses the Stats API.

- Uses asynchronous polling of metrics on a different thread periodically to be exported to Azure Monitor.
- Recording of measures is done synchronously
- Histograms (ValueDistributions) are not supported since to view metrics in Azure Monitor as histograms, you use queries to construct the summary view that you want
- Label Key/Value are inserted into Azure Monitor as "properties"


contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
```diff
@@ -128,6 +128,29 @@ def __init__(self, *args, **kwargs):
...
+        metrics=[],
+        properties=None,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(MetricData, self).__init__(*args, **kwargs)
+        self.ver = self.ver
+        self.metrics = self.metrics
+
+class DataPoint(BaseObject):
```
reyang: sort

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
```diff
@@ -128,6 +128,29 @@ def __init__(self, *args, **kwargs):
...
+    _default = BaseObject(
+        name='',
+        value=0.0,
```
reyang: Should have all the fields defined according to https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/DataPoint.bond.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
```diff
@@ -128,6 +128,29 @@ def __init__(self, *args, **kwargs):
...
+        ver=2,
+        metrics=[],
+        properties=None,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(MetricData, self).__init__(*args, **kwargs)
+        self.ver = self.ver
+        self.metrics = self.metrics
+
```
reyang: minor: lint - two blank lines here

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
```diff
@@ -65,6 +65,25 @@ def __init__(self, *args, **kwargs):
...
+        value=0.0,
+        count=None,
+        min=None,
+        max=None,
+        stdDev=None,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(DataPoint, self).__init__(*args, **kwargs)
+        self.ns = self.ns
```
reyang: `ns` is not required here?

lzchen: Can you explain a little bit about why it is not required?

contrib/opencensus-ext-azure/examples/metrics/sum.py
```diff
@@ -25,13 +25,13 @@
...
 
-CHIPS_EATEN_MEASURE = measure_module.MeasureFloat("chips_eaten",
-                                                  "number of chips eaten",
-                                                  "chips")
-CHIPS_EATEN_VIEW = view_module.View("chips_eaten_view",
-                                    "number of chips eaten",
-                                    [],
-                                    CHIPS_EATEN_MEASURE,
+REQUEST_MEASURE = measure_module.MeasureFloat("Requests",
+                                                "number of requests",
```
reyang: minor, clean up the indentation / lint.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,89 @@
...
 
 __all__ = ['MetricsExporter', 'new_metrics_exporter']
 
 
-class MetricsExporter(object):
+class MetricsExporter(TransportMixin):
     """Metrics exporter for Microsoft Azure Monitor."""
 
-    def __init__(self, options):
+    def __init__(self, options=None):
```
reyang: Is there a reason to prefer `options` other than `**options`?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,89 @@
...
+    def metric_to_data_points(self, metric):
+        """Convert an metric's OC time series to a list of Azure data points."""
+        data_points = []
+        # Each time series will be uniquely identified by it's label values
+        for time_series in metric.time_series:
+            # Using stats, time_series should only have one point
+            # which contains the aggregated value
+            for point in time_series.points:
+                if point.value is not None:
+                    data_point = DataPoint(ns=metric.descriptor.name,
```
reyang: Would the name allow something like a namespace? E.g. https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md#measures
If yes, we might consider to split the name string to namespace and the actual name?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
reyang: According to https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/DataPoint.bond, ns is not required.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,89 @@
...
+            return
+        envelopes = []
         for metric in metrics:
-            # TODO: implement export to Azure Monitor functionality
-            print(repr(metric))
+            # No support for histogram aggregations
+            if metric.descriptor.type == MetricDescriptorType.CUMULATIVE_DISTRIBUTION:
+                continue
+            envelopes.append(self.metric_to_envelope(metric))
+            self._transmit(envelopes)
```
reyang: What if transmission failed?

reyang: Should this be inside the for loop or outside?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/protocol.py
lzchen: Yes. After figuring out the behavior of the injection service, I removed ns as a required field.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: How to control the max batch size (number of items) that we export?

lzchen: Outside. Fixed.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: Interesting solution. Given the max length of name is 1024 characters, it is possible to encode the namespace and actual name inside the "name" field. Having a namespace field is useful as it categorizes what kind of metric its for, and the name serves as the specific type of metric. However there must be some documentation on how to use the encoding and for corner cases as well. (e.g. if we do name=<namespace>:<actual_name>, what happens when either of them is blank or the ":" is missing?)

lzchen: I'd say we leave it as it is for now, I'd rather change the DataPoint API than having to encode something in name.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: The users should create metric exporter instances via new_metrics_exporter(**options) instead of calling the initializer directly. This is a design choice, as I want the starting of the exporter thread to be done outside.

reyang: Got it.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: Since we do not have a defined retry logic for Metrics yet, the exporter simply ignores the transmissions that have failed and logs an error message with the given status code. I do not want to throw and exception and stop the application if a single transmission fails.

lzchen: We have not yet defined what "number of items" means in terms of metrics (# of metrics, time series or points?). The injestion service also only accepts a single metric at a time, the list passed in is simply for convention (so match with stats design). I say we can keep it as it is for now (sends all time series passed in) until we can define what this means.

reyang: `_transmit` has [dependency on the storage](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py#L106), we need to decouple it (or make storage optional if we don't use it in metrics exporter).

lzchen: Yes you are correct. Leaving the logic to run into the exception is not a good practice. I fixed it so that it will only pass to storage if storage is being used (not being used for Metrics exporter). This means that for now, any partial success response we get back from the ingestion service will be discarded.

lzchen: The ingestion service only accepts a single data point at a time, so there is no need for batching.

https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/MetricData.bond

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -102,7 +102,7 @@ def _transmit(self, envelopes):
                                 error['message'],
                                 envelopes[error['index']],
                             )
-                    if resend_envelopes:
+                    if resend_envelopes and self.storage is not None:
```
reyang: This seems to be hack, need to clean up.

contrib/opencensus-ext-azure/tests/test_azure_metrics_exporter.py
```diff
@@ -0,0 +1,158 @@
...
+from opencensus.ext.azure import metrics_exporter
+from opencensus.metrics import label_key
+from opencensus.metrics import label_value
+from opencensus.metrics.export import metric
+from opencensus.metrics.export import metric_descriptor
+from opencensus.metrics.export.metric_descriptor import MetricDescriptorType
+from opencensus.metrics.export import value
+from opencensus.metrics.export import point
+from opencensus.metrics.export import time_series
+from opencensus.ext.azure.common.protocol import DataPoint
```
reyang: sort

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,30 +12,94 @@
...
+        return properties
+
+    def create_envelope(self, data_point, time_stamp, properties):
+        envelope = Envelope(
+            iKey=self.options.instrumentation_key,
+            tags=dict(utils.azure_monitor_context),
+            time=time_stamp.isoformat(),
+        )
+        envelope.name = "Microsoft.ApplicationInsights.Metric"
+        # The ingestion service for Azure Monitor only takes a single
```
reyang: I don't think the ingestion service only supports single data point per request.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,30 +12,94 @@
...
+                        data_point = self.create_data_points(
+                                     time_series, md)[0]
+                        # The timestamp is when the metric was recorded
+                        time_stamp = time_series.points[0].timestamp
+                        # Get the properties using label keys from metric
+                        # and label values of the time series
+                        properties = self.create_properties(time_series, md)
+                        envelope = self.create_envelope(
+                                   data_point, time_stamp, properties)
+                        self._transmit(envelope)
```
reyang: This wouldn't work.
`_transmit` takes a list of envelopes, passing a single envelope would always fail with illegal request body.
This also shows how important it is to check the result instead of silently ignore the error.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: > The ingestion service only accepts a single data point at a time, so there is no need for batching.

It is hard to believe that the ingestion service would be able to serve real customer in production without batching.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: I tested this and it works fine. _transmit merely calls json.dumps and the ingestion service is able to accept individual envelopes. I am able to receive a 200 status whether I input envelope or a list of envelopes. However, given that the name is "envelopes" with a plural, I can change it to pass in a list. Also this me be appropriate if I was mistaken about the ingestion service taking only one data point at a time (I will look into this).

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: You are correct. Fixed to use batch in form of passing in list of envelopes.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: I refactored the "rules" for what _transmit will return and introduced a "NoopStorage" that is used for metrics exporter.

contrib/opencensus-ext-azure/examples/metrics/batch.py
```diff
@@ -26,38 +25,43 @@
...
+                            SUM_MEASURE,
+                            aggregation_module.SumAggregation())
 
 
 def main():
     # Enable metrics
     # Set the interval in seconds in which you want to send metrics
-    exporter = metrics_exporter.new_metrics_exporter(export_interval=5)
+    exporter = metrics_exporter.new_metrics_exporter(
+                                export_interval=2.0,
```
reyang: I suggest that we give a reasonable exporter interval comparable to the ingestion and processing latency.

contrib/opencensus-ext-azure/examples/metrics/simple.py
```diff
@@ -38,7 +38,7 @@
 def main():
     # Enable metrics
     # Set the interval in seconds in which you want to send metrics
-    exporter = metrics_exporter.new_metrics_exporter(export_interval=2)
+    exporter = metrics_exporter.new_metrics_exporter(export_interval=2.0)
```
reyang: I suggest that we give a reasonable exporter interval comparable to the ingestion and processing latency.

contrib/opencensus-ext-azure/examples/metrics/sum.py
```diff
@@ -41,13 +41,14 @@ def main():
     exporter = metrics_exporter.new_metrics_exporter(export_interval=5)
```
reyang: I suggest that we give a reasonable exporter interval comparable to the ingestion and processing latency.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -21,6 +21,12 @@
 logger = logging.getLogger(__name__)
 
 
+class ResponseType(object):
```
reyang: This is confusing. Technically it is not a response type.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -63,7 +69,7 @@ def _transmit(self, envelopes):
         except Exception as ex:  # TODO: consider RequestException
             logger.warning('Transient client side error %s.', ex)
             # client side error (retryable)
-            return self.options.minimum_retry_interval
+            return ResponseType.RETRY
```
reyang: Here we might not even received any reponse (e.g. network failure while sending request), it is very confusing to return something called `ResponseType`.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -134,7 +135,7 @@ def _export(self, batch, event=None):  # pragma: NO COVER
             if batch:
                 envelopes = [self.log_record_to_envelope(x) for x in batch]
                 result = self._transmit(envelopes)
-                if result > 0:
+                if result == ResponseType.RETRY:
                     self.storage.put(envelopes, result)
```
reyang: This is buggy, we end up putting 1 second retry period.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -125,11 +125,11 @@ def _transmit(self, envelopes):
                 text,
             )
             # server side error (retryable)
-            return self.options.minimum_retry_interval
+            return ResponseType.RETRY
```
reyang: merge conflict?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
```diff
@@ -170,3 +170,25 @@ def put(self, data, lease_period=0, silent=False):
             ),
         ))
         return blob.put(data, lease_period=lease_period, silent=silent)
+
+
+class LocalNoopStorage(object):
```
reyang: This is like a mock class for testing, do we want to hack this way?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: Right it does seem confusing. I think it would be better to not refactor transport in this PR.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
lzchen: Refactored to exporter to include _transmit functionality without retry policy.

contrib/opencensus-ext-azure/README.rst
```diff
@@ -109,7 +109,7 @@ Using the Metrics exporter
         def main():
             # Enable metrics
             # Set the interval in seconds in which you want to send metrics
-            exporter = metrics_exporter.new_metrics_exporter(export_interval=2)
+            exporter = metrics_exporter.new_metrics_exporter(export_interval=30.0)
```
reyang: Use default (implicit).

contrib/opencensus-ext-azure/examples/metrics/sum.py
```diff
@@ -25,29 +25,30 @@
...
-CHIPS_EATEN_VIEW = view_module.View("chips_eaten_view",
-                                    "number of chips eaten",
-                                    [],
-                                    CHIPS_EATEN_MEASURE,
-                                    aggregation_module.SumAggregation())
+REQUEST_MEASURE = measure_module.MeasureFloat("Requests",
+                                              "number of requests",
+                                              "requests")
+NUM_REQUESTS_VIEW = view_module.View("Number of Requests",
+                                     "View for number of requests made",
```
reyang: Remove "View of".

contrib/opencensus-ext-azure/examples/metrics/sum.py
```diff
@@ -25,29 +25,30 @@
...
     # Set the interval in seconds in which you want to send metrics
-    exporter = metrics_exporter.new_metrics_exporter(export_interval=5)
+    exporter = metrics_exporter.new_metrics_exporter()
     view_manager.register_exporter(exporter)
 
-    view_manager.register_view(CHIPS_EATEN_VIEW)
+    view_manager.register_view(NUM_REQUESTS_VIEW)
     mmap = stats_recorder.new_measurement_map()
     tmap = tag_map_module.TagMap()
+    tmap.insert("url", "website.com")
```
reyang: Need to be concise here.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,219 @@
...
+        return data_points
+
+    def create_properties(self, time_series, metric_descriptor):
+        properties = {}
+        # We construct a properties map from the
+        # label keys and values
+        # We assume the ordering is already correct
+        for i in range(len(metric_descriptor.label_keys)):
+            if time_series.label_values[i].value is None:
+                value = "None"
```
reyang: Think about how to make it generic.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,219 @@
...
+        )
+        envelope.name = "Microsoft.ApplicationInsights.Metric"
+        data = MetricData(
+            metrics=[data_point],
+            properties=properties
+        )
+        envelope.data = Data(baseData=data, baseType="MetricData")
+        return envelope
+
+    def _transmit_without_retry(self, envelopes):
```
reyang: Leave a todo comment here.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,219 @@
...
+                        else:
+                            logger.error(
+                                'Data drop %s: %s %s.',
+                                error['statusCode'],
+                                error['message'],
+                                envelopes[error['index']],
+                            )
+                    # show the envelopes that can be
+                    # retried manually for visibility
+                    if resend_envelopes:
```
reyang: probably rename

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,31 +12,219 @@
...
+                        logger.warning(
+                            'Error while processing data. Data dropped.' +
+                            'Consider manually retrying for indices at: %s.',
+                            ', '.join(resend_envelopes)
+                        )
+                except Exception as ex:
+                    logger.error(
```
reyang: consider `logger.exception`

contrib/opencensus-ext-azure/README.rst
```diff
@@ -118,7 +118,7 @@ Using the Metrics exporter
 
             mmap.measure_int_put(CARROTS_MEASURE, 1000)
             mmap.record(tmap)
-            time.sleep(10)
+            time.sleep(60)
```
reyang: If the application exits without sleeping, what's the expected result?
What's our recommendation to the users?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,30 +12,222 @@
...
+                        # The timestamp is when the metric was recorded
+                        time_stamp = time_series.points[0].timestamp
+                        # Get the properties using label keys from metric
+                        # and label values of the time series
+                        properties = self.create_properties(time_series, md)
+                        envelopes.append(self.create_envelope(data_point,
+                                                              time_stamp,
+                                                              properties))
+                        # Send data in batches of max_batch_size
+                        if len(envelopes) == self.max_batch_size:
```
reyang: If there are 10 metrics, each contains only one data point, do we end up sending 10 requests or just 1?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -12,30 +12,222 @@
...
+                            self._transmit_without_retry(envelopes)
+                            del envelopes[:]
+            # If leftover data points in envelopes, send them all
+            if envelopes:
```
reyang: Need to optimize the logic flow here.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: The TODO should cover:
1. consolidate with transport logic (instead of duplicating code)
2. handle failures properly

contrib/opencensus-ext-azure/README.rst
lzchen: It is possible that an application will exit before the exporter thread has a chance to exporter the metrics based on exporter_interval.

It is recommended for the user to keep their applications running, or at least long enough for the exporter to meet the interval.

For applications that terminate before the default interval (15s), it is fine because we only have up to one minute of granularity for the ingestion service (soon to be 30s), so the data seen prior to one minute does not have much value.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: If the max_batch_size value is 10,  one request is sent with a body serialized from a list with 10 envelopes.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: The logic here is to send a request as soon as "max_batch_size" amount of metrics have been processed instead of waiting for all of the metrics to be processed.

I think it is minimal optimization, it might be cleaner to just iterate through all the metrics, convert them to envelopes, and then send the envelopes batch by batch. What do you think?

contrib/opencensus-ext-azure/README.rst
reyang: > It is recommended for the user to keep their applications running, or at least long enough for the exporter to meet the interval.

This is vague, what does "long enough" mean? We probably want to document it.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: I see.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: I was thinking about latter, `del envelopes[:]` and having a separate logic handling the residues outside several nested scope could be hard to maintain.

reyang: Up to you. I don't have strong opinion here.

contrib/opencensus-ext-azure/README.rst
reyang: The minimum requirement, we don't want to tell the user to "pick some lucky number, wait and pray".

lzchen: Right. Pit of success.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: Done.

Issue #679, opened 2019-06-06
Invalid point value type error. MeasureInts being silently converted to ValueDouble


mcnamam:
**Describe your environment.** Describe any aspect of your environment relevant to the problem, including your Python version, [platform]
Linux, Python 2.7, opencensus 0.6.0, opencensus-ext-stackdriver 0.4.0

I recently upgraded my opencensus client from 0.3.1 to 0.6.0 to get the fix for https://github.com/census-instrumentation/opencensus-python/issues/350
Now I am encountering the following error when running my script that publishes metrics via opencensus:
ERROR:opencensus.metrics.transport:Error handling metric export
Traceback (most recent call last):
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/metrics/transport.py", line 57, in func
    return self.func(*aa, **kw)
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/metrics/transport.py", line 98, in export_all
    export(get())
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/ext/stackdriver/stats_exporter/__init__.py", line 156, in export_metrics
    metrics = list(metrics)
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/stats/measure_to_view_map.py", line 152, in get_metrics
    metric = metric_utils.view_data_to_metric(vd, timestamp)
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/stats/metric_utils.py", line 132, in view_data_to_metric
    return metric.Metric(md, ts_list)
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/metrics/export/metric.py", line 43, in __init__
    self._check_type()
  File "/python_venv/local/lib/python2.7/site-packages/opencensus/metrics/export/metric.py", line 69, in _check_type
    raise ValueError("Invalid point value type")
ValueError: Invalid point value type

I added some print lines to metric.py to get a better idea of what type mismatch it was encountering and here's what I see:
view bytes_avail_view has metric desc MetricDescriptor(name="bytes_avail_view", description="bytes available under /storage across all hosts", unit=By, type=ValueLong)
ts is TimeSeries([ValueDouble(1.22790525338e+12)], label_values=(), start_timestamp=2019-06-06T21:28:29.757175Z)
check type is <class 'opencensus.metrics.export.value.ValueLong'>

This is my declaration for my bytes_avail measure and view:
[...]

PR #696, opened 2019-06-27
Support int typed aggregations

### LINKED ISSUE #565: Support int export for sum and lastvalue aggregations
`LastValueAggregationData` and `SumAggregationDataFloat` are always converted to a float-valued metrics. This makes metric conversion impossible for views with `MeasureInt` measures.

The java client uses multiple classes (e.g. [`MutableSumLong` and `MutableSumDouble`](https://github.com/c24t/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java)) to solve this problem.

colincadams:
This makes Aggregation a factory of AggregationData, using the type of the underlying measure to determine which value type to use.

fixes #565

Currently just a first pass, looking for feedback on the approach that we started discussing in #565.


reyang:
@lzchen please help to review.

opencensus/stats/aggregation.py
```diff
@@ -78,16 +61,25 @@ class SumAggregation(BaseAggregation):
...
-        """The sum of the current aggregation"""
-        return self._sum
+    def __init__(self, sum=None):
+        super(SumAggregation, self).__init__()
+        self._initial_sum = sum or 0
+
+    def new_aggregation_data(self, measure):
+        value_type = MetricDescriptorType.to_type_class(
+            self.get_metric_type(measure))
+        # TODO: do we need to type cast `_initial_sum`?
```
songy23: I don't think that's necessary.

colincadams: Done.

opencensus/stats/aggregation.py
```diff
@@ -51,8 +35,7 @@ class BaseAggregation(object):
     :param aggregation_type: represents the type of this aggregation
 
     """
-    def __init__(self, buckets=None, aggregation_type=Type.NONE):
-        self._aggregation_type = aggregation_type
+    def __init__(self, buckets=None):
         self._buckets = buckets or []
 
     @property
```
lzchen: You still have an aggregation_type property.

opencensus/stats/aggregation.py
```diff
@@ -78,16 +61,24 @@ class SumAggregation(BaseAggregation):
     :param aggregation_type: represents the type of this aggregation
```
lzchen: Also the comment

opencensus/stats/aggregation.py
```diff
@@ -101,16 +92,16 @@ class CountAggregation(BaseAggregation):
...
```
lzchen: Fix param comments

opencensus/stats/aggregation_data.py
```diff
@@ -52,16 +52,21 @@ def to_point(self, timestamp):
...
-class SumAggregationDataFloat(BaseAggregationData):
+class SumAggregationData(BaseAggregationData):
     """Sum Aggregation Data is the aggregated data for the Sum aggregation
 
-    :type sum_data: float
+    :type value_type: class that is either
+        :class:`opencensus.metrics.export.value.ValueDouble` or
+        :class:`opencensus.metrics.export.value.ValueLong`
+    :param value_type: the type of value to be used when creating a point
+    :type sum_data: meh
```
lzchen: meh?

opencensus/stats/aggregation.py
lzchen: "Sum Aggregation DESCRIBES"

lzchen:
This is great! The factory design is very slick. Just a couple of comments and should be good to go!

opencensus/stats/aggregation_data.py
```diff
@@ -52,16 +52,21 @@ def to_point(self, timestamp):
         raise NotImplementedError  # pragma: NO COVER
 
 
-class SumAggregationDataFloat(BaseAggregationData):
+class SumAggregationData(BaseAggregationData):
```
c24t: Why the decision to make `value_type` an instance attr instead of keeping separate aggregation classes for each value type here?

lzchen:
Also fixes [#679]

opencensus/stats/aggregation.py
colincadams: I ended up removing `BaseAggregation` altogether, per @c24t's comments.

opencensus/stats/aggregation.py
colincadams: Same as above.

opencensus/stats/aggregation.py
colincadams: Done.

opencensus/stats/aggregation_data.py
colincadams: loll 🤸‍♂

opencensus/stats/aggregation.py
colincadams: Done.

opencensus/stats/aggregation.py
```diff
@@ -51,8 +35,7 @@ class BaseAggregation(object):
...
 
     """
-    def __init__(self, buckets=None, aggregation_type=Type.NONE):
-        self._aggregation_type = aggregation_type
+    def __init__(self, buckets=None):
```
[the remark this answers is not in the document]
colincadams: Done, just removed the class.

opencensus/stats/aggregation_data.py
colincadams: Initially, I had this private class with `value_type` and then bound classes like:
```
SumAggregationDataFloat = functools.partial(SumAggregationData, value_type=value.ValueDouble)
SumAggregationDataInt = functools.partial(SumAggregationData, value_type=value.ValueInt)
```

But then in `SumAggregation.new_aggregation_data`, we needed an additional set of conditionals to check the type of the measure or metric and decide which to use. Instead doing it this way, we just need the one set of conditionals in .`get_metric_type` which we need anyway, and can reuse `MetricDescriptor.to_type_class`. It seemed a little cleaner, but really it isn't that different.

We could also duplicate the classes entirely, instead of the `functools.partial`, but that seemed unnecessary.

I guess the other option is to force the user to decide and provide Int or Float, but since we can determine it that seems easier and less error prone.

opencensus/stats/aggregation_data.py
```diff
@@ -52,16 +52,21 @@ def to_point(self, timestamp):
         raise NotImplementedError  # pragma: NO COVER
```
colincadams: The `BaseAggregationData` class seems a little odd to me as well, each implementation supplies `aggregation_data` upon construction but then doesn't update it, and stores there data in a separate instance variable that they update and return.

Is there a reason for this class to exist?

opencensus/stats/aggregation.py
```diff
@@ -14,125 +14,79 @@
...
+            self.get_metric_type(measure))
+        return aggregation_data.SumAggregationData(
+            value_type=value_type, sum_data=self._initial_sum)
+
+    @staticmethod
+    def get_metric_type(measure):
+        """Get the MetricDescriptorType for the metric produced by this
+        aggregation and measure.
+        """
+        if isinstance(measure, measure_module.MeasureInt):
```
reyang: I don't have strong opinion here, it seems a more "Pythonic" style is:
```python
        if isinstance(measure, measure_module.MeasureInt):
            return MetricDescriptorType.CUMULATIVE_INT64
        if isinstance(measure, measure_module.MeasureFloat):
            return MetricDescriptorType.CUMULATIVE_DOUBLE
        raise ValueError
```
Up to you.

colincadams: Good catch, surprised the linter didn't yell about this (or maybe it was a warning I didn't see :grin:)

opencensus/stats/aggregation_data.py
c24t: Not that I can tell, besides to show that aggregations should implement `to_point`, but in that case there should be an unimplemented `add_sample` in the class too.

In any case it looks like you can lose the `aggregation_data` attr and property.

colincadams: I just removed `BaseAggregationData` entirely

colincadams:
Thanks all! And thanks for updating the changelog @c24t, totally forgot!

PR #701, opened 2019-06-30
More robust header decoding and encoding

### LINKED ISSUE #465: Cloud format propagator parsing drops trace options
See https://github.com/census-instrumentation/opencensus-python/pull/464#issuecomment-457015893.

```python
>>> middleware.propagator.to_headers(SpanContext(trace_options=TraceOptions(0)))
{'X-Cloud-Trace-Context': 'd5e525e45bbe4d25b571fa0988749117/None;o=0'}

>>> middleware.propagator.from_headers({'X-Cloud-Trace-Context': 'd5e525e45bbe4d25b571fa0988749117/None;o=0'})
SpanContext(trace_id=d5e525e45bbe4d25b571fa0988749117, span_id=None, trace_options=TraceOptions(enabled=True), tracestate=None)
```

mhindery:
Fixes #465

When serializing to string in `to_header`, omit the slash and the span_id alltogether if there is no span_id. Previously this would include the literal string `/None`. I'm not entirely sure if that is the right way to do it, or if you prefer it with a slash and `0` as span_id? With the current libraries around, I am able to set `X-Cloud-Trace-Context` in requests with a value that does not include any 'slash + something'-part at all and that works across everything, so I based the omission of it in the `to_header` based on that.

When deserializing an incoming header in `from_header`, now check if the span_id is either the literal `None` string or just the empty string and in that case set it to None. This makes it more robust, as it now can take all of the formats `d5e525e45bbe4d25b571fa0988749117/d5e525e45bbe4d25;o=0`, `d5e525e45bbe4d25b571fa0988749117/None;o=0`, `d5e525e45bbe4d25b571fa0988749117/;o=0` and `d5e525e45bbe4d25b571fa0988749117;o=0`


c24t:
> Previously this would include the literal string `/None`. I'm not entirely sure if that is the right way to do it, or if you prefer it with a slash and `0` as span_id?

It's certainly a bug that we were including "None" in the header here. If there's no span ID in the span context I think it's reasonable to use `INVALID_SPAN_ID` (`0000000000000000`) here.

I agree with making `from_header` more robust, but I think we should fix `to_header` and refuse to parse "None" here.

What are you doing that you're regularly parsing traces without span IDs?

mhindery:
It's not that I regularly work with them without span ID's, more that it made sense to fallback to a none value and let a span id be autogenerated.

The way I use these; when I want something to be traced, I add a header in a request with a traceID in it (which I later can look up in StackDriver), but I don't care about the span ID, that can be whatever it wants for me, so I omit it. My header value always is the format of `6e0c63257de34c92bf9efcd03927272e;o=1`. Isn't being required to generated and add a span id every time quite cumbersome?

mhindery:
I see that the [SpanContext._check_span_id](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span_context.py#L118) returns None when you give it a non-matching, random value, while the [SpanContext._check_trace_id](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span_context.py#L138) function returns an auto-generated one in that case. Is that intentional? Both are claimed to be optional in the class docstring, yet the span_id is in fact not generated. There is a `generate_span_id()` function in the file, but it's never called.

c24t:
> Isn't being required to generated and add a span id every time quite cumbersome? When a span_id is omited, I interpret it as 'make one for me'.

I'd expect most people to create the span via the `Span` class, which is where `generate_span_id` is called: https://github.com/census-instrumentation/opencensus-python/blob/ed286275b8a854ca4b7cd496e789697d4422dfd9/opencensus/trace/span.py#L237-L238

> Both are claimed to be optional in the class docstring, yet the span_id is in fact not generated

It looks like the docstring was changed in #147 without changing the actual logic: https://github.com/census-instrumentation/opencensus-python/pull/147/files#diff-b15d3f1cb2307e569c217a11c197c335R48.

opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -19,7 +19,8 @@
 from opencensus.trace.trace_options import TraceOptions
 
 _TRACE_CONTEXT_HEADER_NAME = 'X-Cloud-Trace-Context'
-_TRACE_CONTEXT_HEADER_FORMAT = r'([0-9a-f]{32})(\/([0-9a-f]{16}))?(;o=(\d+))?'
+_TRACE_CONTEXT_HEADER_FORMAT = r'([0-9a-f]{32})'\
+    r'(\/([0-9a-f]{16}|(None|)))?(;o=(\d+))?'
```
reyang: It makes more sense to use `0000000000000000` than `None` unless:
1. `None` is a well documented format for `X-Cloud-Trace-Context`, here we're just trying to follow the spec.
2. We need to support `None` for backward compatibility (as there are existing services which send `None` already, and we need to interop with them).

mhindery: Regarding 2; the old behaviour wrongly parsed the trace options but silently ignored the wrong span_id. With this new regex there is still span_id=None, but a warning is logged and the trace options are correct. See a header being parsed with the old and the new function below:

```python

In [131]: header3_on                                                                                                                                                                                                      
Out[131]: '6e0c63257de34c92bf9efcd03927272e/None;o=1'

In [132]: orig_from_header(header3_on)                                                                                                                                                                                    
Out[132]: SpanContext(trace_id=6e0c63257de34c92bf9efcd03927272e, span_id=None, trace_options=TraceOptions(enabled=True), tracestate=None)

In [133]: new_from_header(header3_on)                                                                                                                                                                                     
WARNING:root:Span_id None does not the match the required format
Out[133]: SpanContext(trace_id=6e0c63257de34c92bf9efcd03927272e, span_id=None, trace_options=TraceOptions(enabled=True), tracestate=None)
```

```python
In [134]: header3_off                                                                                                                                                                                                     
Out[134]: '6e0c63257de34c92bf9efcd03927272e/None;o=0'

In [135]: orig_from_header(header3_off)                                                                                                                                                                                   
Out[135]: SpanContext(trace_id=6e0c63257de34c92bf9efcd03927272e, span_id=None, trace_options=TraceOptions(enabled=True), tracestate=None)

In [136]: new_from_header(header3_off)                                                                                                                                                                                    
WARNING:root:Span_id None does not the match the required format
Out[136]: SpanContext(trace_id=6e0c63257de34c92bf9efcd03927272e, span_id=None, trace_options=TraceOptions(enabled=False), tracestate=None)
```

c24t: Do you think it's still worth handling the "None" case here now that `to_header` is fixed? One argument in favor is that this will work with older versions of the library that included the bug, but we wouldn't check for "None" here if the bug had never existed.

mhindery: I'm happy to remove it, the argument would indeed be backwards compatibility with 'old' incoming headers, but seeing as the current behaviour isn't correct anyways, and it's only a small and non-breaking difference in behaviour, I would agree that it might not be worth including it.

Related, what's your opinion on handling headers with no span id at all (so e.g. `6e0c63257de34c92bf9efcd03927272e/;o=0` or without the slash `6e0c63257de34c92bf9efcd03927272e;o=0`)? If that should not be handled, then the extra group can be removed from the regex. To support it and lt resulting in a spancontext without span_id but with the right trace_options, an empty group can be kept where now the `None|` is.

mhindery:
Upon further debugging some strange behaviour, [this different cross-langauge behaviour](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/169) came up as an issue, it seems relevant to here as well:

> As a related but important note, which I'll raise in the Python repo as well: there's a difference in behaviour in parsing the headers. Python uses [a regex](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/propagation/google_cloud_format.py#L22) which only matches exactly 16 characters as span id. The Go code doesn't use that regex (or any other), but works with splitting on separators and happily accepts a span id which is shorter than 16 characters. It will also propagate that shorter span ID further. The result is that the trace propagation fails between services in the different langauges. Should this behaviour be equalized between langauges?

mhindery:
I've started another PR related to the headers on, if that one goes through then this one is no longer relevant I think: https://github.com/census-instrumentation/opencensus-python/pull/719

c24t:
Closing in favor of #719.

Issue #695, opened 2019-06-25
Implement "out-of-the-box" (OOTB) metrics for exporters


lzchen:
It will be valuable for users to automatically send "out-of-the-box" (OOTB) metrics easily as soon as they instrument their applications with a metrics exporter. The SDK can expose a simple way to turn these on. Necessary measures and views will be created automatically depending on the specific metric required.

We need to first define which OOTB metrics we are going to implement. IMO a good place to start would be with CPU related, memory related and requests related metrics.

reyang:
Maintain the thread #724, #722, #720, #718.

lzchen:
[#708]

lzchen:
Referenced are all the "out-of-the-box" metrics currently scoped for Azure exporter. Closing this and opening an issue to address latency concerns for monkey patching request/http server libraries.

PR #708, opened 2019-07-01
Implement Standard Metrics - Free Memory

lzchen:
Initial implementation of standard metrics. [#695]
Standard Metrics are defined as metrics that will be exported by default using the metrics exporter.
Standard metrics are enabled by default as suggested by @c24t 

The first one that I am including is Free Memory **in bytes**.
Free Memory is defined as: memory not being used at all (zeroed) that is readily available; note that this doesn’t reflect the actual memory available (use available instead). total - used does not necessarily match free. Defined in https://psutil.readthedocs.io/en/latest/

In Azure Monitor, there are special namespace names to denote a metric as being "standard". For Free Memory, you can view the telemetry in metrics explorer by going to Standard Metrics (preview) namespace and Available Memory under Performance Counters.

As for architectural decisions, the periodic recording of the standard metrics must be non-blocking to the main thread. So I chose to use a separate recording  thread (initiated by MetricExporterTask) to perform this. The exporter and recorder threads should be decoupled to perform their specific jobs (the simpler the task, the better on cpu and memory). 

There is a possible race condition in which the exporting is called before the recording, but the metric will then be simply delayed by one interval, which I don't think is a huge issue.


lzchen:
image

lzchen:
image

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -239,11 +245,57 @@ def _transmit_without_retry(self, envelopes):
...
+            "Amount of free memory in bytes",
+            [],
+            free_memory_measure,
+            aggregation_module.LastValueAggregation())
+        view_manager.register_view(free_memory_view)
+
+    def record_standard_metrics(self):
+        # Function called periodically to record standard metrics
+        vmem = psutil.virtual_memory()
+        free_memory = vmem.free
```
reyang: Do we need "Available Memory" for Azure Monitor, or "Free Memory"?
It seems what Azure Monitor has is Available Memory.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -239,11 +245,57 @@ def _transmit_without_retry(self, envelopes):
...
 def new_metrics_exporter(**options):
-    options = Options(**options)
-    exporter = MetricsExporter(options=options)
+    options_ = Options(**options)
+    exporter = MetricsExporter(options=options_)
+    if options_.enable_standard_metrics:
+        exporter.enable_standard_metrics()
+        # Separate thread for recording of metrics
+        transport.get_recorder_thread(exporter.record_standard_metrics,
+                                      interval=options_.export_interval)
```
reyang: Not sure if the recording interval should be the same as export interval. What are the other SDKs doing today?

contrib/opencensus-ext-azure/setup.py
```diff
@@ -41,6 +41,7 @@
     install_requires=[
         'opencensus >= 0.7.dev0, < 1.0.0',
         'requests >= 2.19.0',
+        'psutil >= 5.6.3',
```
reyang: nit, sort

opencensus/metrics/transport.py
```diff
@@ -100,3 +100,24 @@ def export_all():
...
+    :param record_metrics: The function called to record metrics.
+
+    :type interval: int or float
+    :param interval: Seconds between record calls.
+
+    :rtype: :class:`PeriodicTask`
+    :return: A running thread responsible for executing record_metrics().
+
+    """
+    tt = MetricExporterTask(interval, record_metrics)
```
reyang: Maybe we should find another name for `MetricExporterTask` as we started to use it for non-exporting tasks.

reyang: Maybe we should find another name for `MetricExporterTask` as we started to use it for non-exporting tasks.

lzchen: Good idea. I will use something more appropriate.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: You are correct. I was following Node SDK which uses os.freemem() but apparently that returns the available memory. I will change this.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: In the Node SDK, the intervals for recording and exporting are different and can be configured. I was using the same interval because I was thinking of simplicity for the user so they do not have to configure anything in options. Perhaps derive the recording interval from the export interval (maybe interval-%age), as long as the recording happens prior.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -239,11 +245,57 @@ def _transmit_without_retry(self, envelopes):
                 text,
             )
 
+    def enable_standard_metrics(self):
+        # Sets up stats related objects to begin
+        # recording standard metrics
+        stats_ = stats.stats
```
reyang: Curious, why not use `stats_module.stats` for consistency?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -239,11 +245,57 @@ def _transmit_without_retry(self, envelopes):
...
+            [],
+            available_memory_measure,
+            aggregation_module.LastValueAggregation())
+        view_manager.register_view(available_memory_view)
+
+    def record_standard_metrics(self):
+        # Function called periodically to record standard metrics
+        vmem = psutil.virtual_memory()
+        available_memory = vmem.available
+        available_measure = self.standard_metrics_measure_map[StandardMetricType.AVAILABLE_MEMORY]
```
reyang: The 'available_measure' name is confusing.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -239,11 +245,57 @@ def _transmit_without_retry(self, envelopes):
...
+
+        # Standard metrics uses a separate instance of MeasurementMap
+        # from regular metrics but still shares the same underlying
+        # map data in the context. This way, the processing during
+        # record is handled through a separate data structure but
+        # the storage only uses one single data structure
+        # Uniqueness should be based off the name of the view and
+        # measure. A warning message in the logs will be shown
+        # if there are duplicate names
+        self.standard_metrics_map = stats_recorder.new_measurement_map()
```
reyang: Consider decoupling the actual logic from the exporter class. We will need to add several default metrics, need to control the size of this file and have a better organized way.

lzchen: Yes I will do this.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: @reyang We can also add another option in Options. What do you think?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: I'll use stats_module.stats for consistency.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: Good catch. I will match with the previous naming.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
reyang: I wonder how we plan to do this in OpenCensus/OpenTelemetry, it is not supposed to be exporter specific I guess?

lzchen: If it won't be exporter specific, we can abstract out Options to have OpenCensus default behaviour and then implement exporter specific behaviour. Just an idea.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+        stats_recorder = stats_.stats_recorder
+
+        # Standard metrics uses a separate instance of MeasurementMap
+        # from regular metrics but still shares the same underlying
+        # map data in the context. This way, the processing during
+        # record is handled through a separate data structure but
+        # the storage only uses one single data structure
+        # Uniqueness should be based off the name of the view and
+        # measure. A warning message in the logs will be shown
+        # if there are duplicate names
```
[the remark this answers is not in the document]
lzchen: It's to pass the lint checks for maximum characters. I just eyeballed the length.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+from opencensus.stats import stats as stats_module
+from opencensus.stats import view as view_module
+
+
+# Definitions taken from psutil docs
+# https://psutil.readthedocs.io/en/latest/
+class StandardMetricsType(object):
+    # Memory that can be given instantly to
+    # processes without the system going into swap
+    AVAILABLE_MEMORY = "\\Memory\\Available Bytes"
```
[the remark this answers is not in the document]
lzchen: This is the namespace that is used in Azure Monitor that uniquely identifies certain standard metrics.

lzchen:
@c24t The gauges look really cool! DerivedGauge seems to fit exactly what we need. We can also get rid of the recorder thread this way. If this is implemented for standard metrics, then the exporter would essentially be "getting metrics" from two different places, the gauges registry and the measure to view map. So before they are exported, these two lists should be merged. There is a case where measure_to_view_map produces metrics that have the same name as the certain metrics produces from the gauges registry. But since we are in control of the naming for standard metrics, I don't think this is a huge issue. What do you think?

lzchen:
Implementation now uses Gauges instead of measures and views. StandardMetricProducer and BaseStandardMetric are also abstracted into the core sdk. Developers can make their own Producers and StandardMetrics that extend from these classes.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,65 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import psutil
+
+from opencensus.metrics.export.gauge import DerivedLongGauge
+from opencensus.metrics.export.gauge import Registry
+from opencensus.metrics.export.standard_metrics import StandardMetricsProducer
+from opencensus.metrics.export.standard_metrics import BaseStandardMetric
```
reyang: minor, `BaseStandardMetric` before `StandardMetricsProducer`?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,65 @@
...
+
+class StandardMetricsType(object):
+    AVAILABLE_MEMORY = "\\Memory\\Available Bytes"
+
+
+# Definitions taken from psutil docs
+# https://psutil.readthedocs.io/en/latest/
+class AvailableMemoryStandardMetric(BaseStandardMetric):
+    """ Metric for Available Memory
+        Avaliable memoru is defined as memory that can be given
```
reyang: typo: memoru -> memory

contrib/opencensus-ext-azure/tests/test_standard_metrics.py
```diff
@@ -0,0 +1,33 @@
...
+from opencensus.stats import execution_context
+from opencensus.stats import stats as stats_module
+
+
+class TestStandardMetrics(unittest.TestCase):
+    def setUp(self):
+        execution_context.set_measure_to_view_map({})
+        stats_module.stats = stats_module._Stats()
+
+    def test_constructor_and_setup(self):
```
reyang: Is this intentionally empty (since the purpose is to call `setUp`)?

opencensus/metrics/transport.py
```diff
@@ -61,19 +62,23 @@ def func(*aa, **kw):
...
 
 
-def get_exporter_thread(metric_producer, exporter, interval=None):
+def get_exporter_thread(metric_producers, exporter, interval=None):
     """Get a running task that periodically exports metrics.
 
     Get a `PeriodicTask` that periodically calls:
 
-        exporter.export_metrics(metric_producer.get_metrics())
+        exporter.export_metrics(all_metrics)
```
reyang: Consider putting the actual code example here (instead of the description below).
Consider `itertools.chain`.

contrib/opencensus-ext-azure/tests/test_standard_metrics.py
lzchen: My mistake I should have commented this out. I tend to leave writing tests once the design changes have been approved.

opencensus/metrics/transport.py
```diff
@@ -61,19 +62,23 @@ def func(*aa, **kw):
             except Exception:
                 logger.exception("Error handling metric export")
 
-        super(MetricExporterTask, self).__init__(interval, func, args, kwargs)
+        super(PeriodicMetricTask, self).__init__(interval, func, args, kwargs)
 
 
-def get_exporter_thread(metric_producer, exporter, interval=None):
+def get_exporter_thread(metric_producers, exporter, interval=None):
```
reyang: I have a design question, do we intend to have one `MetricProducer` taking multiple metrics, or multiple `MetricProducer` (which I guess is the case)? If it is the latter, what's the main consideration of having multiple producers (e.g. just to categorize by ownership, or ext lib, or we expect each producer to export individually?)

lzchen: Yes we intend to take multiple MetricProducer. The idea is simply as you noted to categorize ownership of the types of metrics that each Producer produces. My thought was to separate these based on logic, so if a developer would like to produce some other set of metrics using a different architecture (stats, gauges, etc.), they can simply create their own MetricProducer and implement "get_metrics".

reyang: Got it, thanks.

tests/unit/metrics/test_transport.py
```diff
@@ -114,9 +114,30 @@ def test_producer_deleted(self, mock_logger):
...
+            producer1.get_metrics.assert_not_called()
+            producer2.get_metrics.assert_not_called()
+            exporter.export_metrics.assert_not_called()
+            time.sleep(INTERVAL + INTERVAL / 2.0)
+            producer1.get_metrics.assert_called_once_with()
+            producer2.get_metrics.assert_called_once_with()
+            exporter.export_metrics.assert_called_once_with(metrics)
+        finally:
+            task.cancel()
+            task.join()
```
reyang: nit, new line before EOF

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -241,9 +242,12 @@ def _transmit_without_retry(self, envelopes):
 
 
 def new_metrics_exporter(**options):
-    options = Options(**options)
-    exporter = MetricsExporter(options=options)
-    transport.get_exporter_thread(stats.stats,
+    options_ = Options(**options)
```
reyang: I don't have a strong opinion here, but I was curious on why do we need `options_` instead of using `options = Options(**options)`.

lzchen: It was to differentiate the parameter from the local variable. I didn't want to name it something completely different.

tests/unit/metrics/export/test_standard_metrics.py
```diff
@@ -0,0 +1,45 @@
...
+        producer.register_metrics(metrics)
+
+        metric.register.assert_called_once()
+
+    def test_producer_get_metrics(self):
+        test_gauge = DerivedLongGauge('test gauge',
+                                      'test',
+                                      'test',
+                                      [])
+        test_gauge.create_default_time_series(lambda x: (1))
```
reyang: why `(1)` instead of `1`?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class StandardMetricsType(object):
+    AVAILABLE_MEMORY = "\\Memory\\Available Bytes"
+
+
+# Definitions taken from psutil docs
+# https://psutil.readthedocs.io/en/latest/
+class AvailableMemoryStandardMetric(BaseStandardMetric):
+    """ Metric for Available Memory
```
reyang: minor, the comments are using different styles, it'll be good to keep a consistent style.

opencensus/metrics/transport.py
reyang: @c24t would you take a look at the changes to the core library? Thanks.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+        available_memory_gauge = DerivedLongGauge(
+            StandardMetricsType.AVAILABLE_MEMORY,
+            'Amount of available memory in bytes',
+            'byte',
+            [])
+        available_memory_gauge.create_default_time_series(
+            self.get_available_memory)
+        registry.add_gauge(available_memory_gauge)
+
+    def get_available_memory(self):
```
c24t: You might want to lift this out of the class too (or make it a static/classmethod) since it doesn't touch class state.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+        return psutil.virtual_memory().available  # pragma: NO COVER
+
+
+class AzureStandardMetricsProducer(StandardMetricsProducer):
+    """Implementation of the producer of standard metrics
+
+    :type standard_metrics: list
+    :param standard_metrics: the list of standard metrics
+        :class:`BaseStandardMetric`
```
c24t: Same deal, looks like the wrong line.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+            StandardMetricsType.AVAILABLE_MEMORY,
+            'Amount of available memory in bytes',
+            'byte',
+            [])
+        available_memory_gauge.create_default_time_series(
+            self.get_available_memory)
+        registry.add_gauge(available_memory_gauge)
+
+    def get_available_memory(self):
+        return psutil.virtual_memory().available  # pragma: NO COVER
```
c24t: You probably do want a test here since this line getting called is the whole point of the module. Just mock out `psutil`.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+from opencensus.metrics.export.standard_metrics import StandardMetricsProducer
+
+
+class StandardMetricsType(object):
+    AVAILABLE_MEMORY = "\\Memory\\Available Bytes"
+
+
+# Definitions taken from psutil docs
+# https://psutil.readthedocs.io/en/latest/
+class AvailableMemoryStandardMetric(BaseStandardMetric):
```
c24t: Like `StandardMetricsProducer` below, since the only state in this class in the registry -- and I don't think you need the registry here -- I don't see a reason to have this class.

Flattening this class could look like:

```python
def get_available_memory():
    return psutil.virtual_memory().available


def get_available_memory_metric():
    gauge = DerivedLongGauge(
        AVAILABLE_MEMORY,
        'Amount of available memory in bytes',
        'byte',
        [])
    gauge.create_default_time_series(get_available_memory)


class AzureStandardMetricsProducer(MetricProducer):
    def __init__(self):
        self.registry = Registry()
        self.registry.add_gauge(get_available_memory_metric())

    def get_metrics(self):
        return self.registry.get_metrics()


producer = AzureStandardMetricsProducer()
```

This is just one option -- it'd be fine to have separate classes for each metric still instead -- but I don't see the point of passing the registry to the metric.

contrib/opencensus-ext-azure/tests/test_azure_standard_metrics.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_producer_ctor(self):
+        producer = standard_metrics.AzureStandardMetricsProducer()
+
+        attributes = inspect.getmembers(
+            standard_metrics.StandardMetricsType,
+            lambda a: not(inspect.isroutine(a)))
+        types = [a for a in attributes
+                 if not(a[0].startswith('__') and
+                        a[0].endswith('__'))]
+        self.assertEquals(len(producer.metrics), len(types))
```
c24t: It'd be a lot clearer just to list these out (especially since there's only one metric at the moment) than call `inspect` in the test.

opencensus/metrics/transport.py
```diff
@@ -61,19 +62,23 @@ def func(*aa, **kw):
...
     """Get a running task that periodically exports metrics.
 
     Get a `PeriodicTask` that periodically calls:
 
-        exporter.export_metrics(metric_producer.get_metrics())
+        export(itertools.chain(*all_gets))
 
-    :type metric_producer:
+        where all_gets is the concatenation of all metrics produced by
```
c24t: Dedent this to get it to render as text, and consider wrapping the code blocks in backticks. You might want to generate the docs to see how this looks.

opencensus/metrics/transport.py
```diff
@@ -61,19 +62,23 @@ def func(*aa, **kw):
...
+        the metric producers in metric_producers, each calling
+        metric_producer.get_metrics()
+
+    :type metric_producers: list
+    :param metric_producers: The list of metric producers to use to get metrics
```
c24t: Same thing, move down a line.

opencensus/metrics/transport.py
c24t: This change looks fine, but the other option is to add a metric producer class that calls other metric producers under the hood. Since we're likely to make a similar change for exporters we might want to revisit this later. But note that this change changes the API, the other way doesn't.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 
 ## Unreleased
+- Refactored PeriodicMetricTask
```
c24t: The changelog should generally describe user-facing changes.

lzchen:
@c24t Thanks for reviewing. My initial design idea was to abstract the metric types and producers so that developers can write their own standard metrics and producers. However I realize that Metric_producer class does essentially this for ALL metric producers; there is no need to have an interface for STANDARD metric producer.

lzchen:
@c24t I removed the standard metrics file from the core sdk as per your suggestion. It doesn't make a lot of sense to include it. I have also made changes to the docstrings to adhere to the recommended practices for python docstrings. Can you take a look when you have the time?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -54,18 +55,17 @@ def export_metrics(self, metrics):
                 type_ = metric.descriptor.type
                 if type_ != MetricDescriptorType.CUMULATIVE_DISTRIBUTION:
                     md = metric.descriptor
-                    # Each time series will be uniquely
-                    # identified by it's label values
+                    # Each time series will be uniquely identified by it's
```
c24t:
```suggestion
                    # Each time series will be uniquely identified by its
```

contrib/opencensus-ext-azure/tests/test_azure_metrics_exporter.py
```diff
@@ -450,11 +450,24 @@ def test_create_envelope(self):
...
+    @mock.patch('opencensus.ext.azure.metrics_exporter'
+                '.transport.get_exporter_thread')
     def test_new_metrics_exporter(self, exporter_mock):
         iKey = '12345678-1234-5678-abcd-12345678abcd'
         exporter = metrics_exporter.new_metrics_exporter(
             instrumentation_key=iKey)
 
         self.assertEqual(exporter.options.instrumentation_key, iKey)
+        self.assertEqual(len(exporter_mock.call_args_list), 1)
+        self.assertEqual(len(exporter_mock.call_args[0][0]), 2)
```
c24t: Why not check that the the actual metric producer is in the arg list here?

contrib/opencensus-ext-azure/examples/metrics/standard.py
```diff
@@ -0,0 +1,36 @@
...
+
+    print(_exporter.max_batch_size)
+    for i in range(100):
+        print(psutil.virtual_memory())
+        time.sleep(5)
+
+    print("Done recording metrics")
+
+
+if __name__ == "__main__":
```
c24t: More for @reyang than @lzchen: if you expect users to actually run the azure example scripts you might want to make them executable and add a `#!/usr/bin/env python` line up top. Not for this PR, but to consider later.

reyang: And `chmod 755`.

lzchen:
@reyang Ready to merge!

PR #718, opened 2019-07-13
Standard Metric - Private Bytes

lzchen:
Add another standard metric for Azure exporter - Private Bytes.

Private bytes for the current process is measured by the Resident Set Size, which is the non-swapped physical memory a process has used.


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -45,6 +49,34 @@ def get_available_memory_metric():
     return gauge
 
 
+def get_process_private_bytes():
+    try:
+        process = psutil.Process()
+        return process.memory_info().rss
+    except psutil.NoSuchProcess as ex:
+        logger.error('Error: Process does not exist %s.', ex)
```
reyang: Use `logger.exception`.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -45,6 +49,34 @@ def get_available_memory_metric():
...
+    except psutil.AccessDenied as ex:
```
reyang: What about other [exceptions](https://psutil.readthedocs.io/en/latest/#exceptions) that might be added in the future?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -45,6 +49,34 @@ def get_available_memory_metric():
     return gauge
 
 
+def get_process_private_bytes():
+    try:
+        process = psutil.Process()
+        return process.memory_info().rss
+    except psutil.NoSuchProcess as ex:
```
reyang: Curious, when would `NoSuchProcess` get triggered?

contrib/opencensus-ext-azure/tests/test_azure_standard_metrics.py
```diff
@@ -46,3 +47,38 @@ def test_get_available_memory(self, psutil_mock):
         mem = standard_metrics.get_available_memory()
 
         self.assertEqual(mem, 100)
+
+    def test_get_process_private_bytes_metric(self):
+        gauge = standard_metrics.get_process_private_bytes_metric()
+
+        self.assertEqual(gauge.descriptor.name,
```
reyang: Add a TODO here. Need to evolve the backend to be more generic (currently it is "Windows focus" when one look at the names).

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
lzchen: Good point. I'll add a generic catch.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
lzchen: `os.getpid` is called when constructing a Process with no pid. `NoSuchProcess` is raised if the pid does not exist. When accessing methods of the Process class, it is good to always catch `NoSuchProcess` and `AccessDenied`

reyang: I'm very confused - are we getting the info from the current process? How could we get `NoSuchProcess`?

lzchen: Yes we are getting the info from the current process so there is no way to get `NoSuchProcess`. I included the exception based off of psutil docs to always include these two exception checks.
https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info

lzchen:
@reyang 
Reasoning behind exception logic:
1. Using `except Exception` instead of `except`
 - `except Exception` will only handle exceptions derived from `Exception` class which is fine
- Pylint also complains about using bare `except`
2. Removing `NoSuchProcess` and `AccessDenied` 
- These can be caught with except, there is no specific logic to do for these exceptions
3. Writing to logs for every exception
- Simple.

reyang:
> Using `except Exception` instead of `except`
> 
> * `except Exception` will only handle exceptions derived from `Exception` class which is fine
> * Pylint also complains about using bare `except`

Great.
 
> Removing `NoSuchProcess` and `AccessDenied`
> 
> * These can be caught with except, there is no specific logic to do for these exceptions

Good.

> Writing to logs for every exception
> 
> * Simple.

I'm worried about customers getting repetitive error logs which overwhelmed their console outputs or log files. It is okay to leave it for now. We might need to revisit later.

reyang:
We might need to update the document (under the section "Below is a list of standard metrics that are currently available") as well.

PR #719, opened 2019-07-16
Span id encoding

mhindery:
The span ID is currently a hex-encoded 16-char string, which shouldn't be the case. It should be a decimal, see https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/169#issuecomment-511620825
> @mhindery according this [doc](https://cloud.google.com/trace/docs/troubleshooting) spanId should be decimal.
> For first span it should be 0. example copied from the doc
> 
> ```
> curl "http://www.example.com" --header "X-Cloud-Trace-Context:
>   105445aa7843bc8bf206b120001000/0;o=1"
> ```

This breaks compatibility between languages as (e.g. in my case) python and go don't propagate traces correctly.

I changed this by the modifying the `check_span_id` and `generate_span_id` in the SpanContext class to not generate a hexadecimal representation, but just a decimal one. If that would be a problem for other parts of the code which require hex representation, it could be kept, and converted in the to_header() method of the `GoogleCloudFormatPropagator`, however, if it isn't necessary to actually have a hex representation then the conversion can be omitted altogether as it's simpler to just have it as decimal everywhere.


opencensus/trace/span_context.py
```diff
@@ -149,13 +146,13 @@ def _check_trace_id(self, trace_id):
 
 
 def generate_span_id():
-    """Return the random generated span ID for a span. Must be a 16 character
-    hexadecimal encoded string
+    """Return the random generated span ID for a span. Must be a 64 bit
```
reyang: I have a different understanding.
`SpanContext` is the in-memory representation, where we prefer span_id as byte array for sake of performance and interoperability with different protocols. https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#spanid

The actual propagator will be in charge of converting the incoming request headers into the in memory representation, or converting the in-memory representation to outgoing request headers. Examples are https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/propagation/b3_format.py#L61 and https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/propagation/trace_context_http_header_format.py#L55.

In order to have the correct implementation for Stackdriver format, what you need to change is the serialization and deserialization logic inside `google_cloud_format.py`, rather than the `SpanContext` data structure.

mhindery: I understand, out of curiosity, what is the advantage of having it as hexadecimal representation internally? Is it because most other representations use hex, and Stackdriver is the exception?

reyang: I believe it was chosen due to the performance consideration + being neutral to the actual protocols.

You can refer to https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L60.

opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -107,7 +110,7 @@ def to_header(self, span_context):
 
         header = '{}/{};o={}'.format(
             trace_id,
-            span_id,
+            int(span_id, 16),
```
reyang: We probably need to update `from_header` as well.

@c24t FYI - this is a potential breaking change (e.g. if customers mix use old version and the latest version of OpenCensus Python SDK).

mhindery: the from_header has been updated, it contains

```python
            if span_id:
                span_id = '{:016x}'.format(int(span_id))
```

to do the conversion, or do you mean something additional?

reyang: @mhindery please update `CHANGELOG.md` if this is a breaking change. Thanks.

reyang: Adding @c24t to comment from Google side.

I cannot find the official spec for `X-Cloud-Trace-Context`, and I'm a bit worried that this change might break existing stuff for Google (considering there are system tests running as part of the CI, it is a bit surprising to me that we've been sending the wrong format to Stackdriver).

opencensus/trace/propagation/google_cloud_format.py
```diff
@@ -19,7 +19,7 @@
 from opencensus.trace.trace_options import TraceOptions
 
 _TRACE_CONTEXT_HEADER_NAME = 'X-Cloud-Trace-Context'
-_TRACE_CONTEXT_HEADER_FORMAT = r'([0-9a-f]{32})(\/([0-9a-f]{16}))?(;o=(\d+))?'
+_TRACE_CONTEXT_HEADER_FORMAT = r'([0-9a-f]{0,32})(\/([\d]{0,20}))?(;o=(\d+))?'
```
reyang: `{0,32}` -> `{32}` according to the [doc](https://cloud.google.com/trace/docs/troubleshooting):
```TRACE_ID is a 32-character hexadecimal value representing a 128-bit number. It should be unique between your requests, unless you intentionally want to bundle the requests together. You can use UUIDs.```

**Comment by mhindery** [comment]
I don't get why the coverage claims the lines 65-68 of `opencensus/trace/propagation/google_cloud_format.py` are not covered. There is a system test `TestBasicTrace` which calls the `from_header` method on the propagator with a valid header, and those lines it claims are missing, are handled in that method.

**Comment by reyang** [comment]
> I don't get why the coverage claims the lines 65-68 of `opencensus/trace/propagation/google_cloud_format.py` are not covered. There is a system test `TestBasicTrace` which calls the `from_header` method on the propagator with a valid header, and those lines it claims are missing, are handled in that method.

I guess the coverage is complaining about a missing branch coverage:
```
L65         if span_id:
L66             span_id = '{:016x}'.format(int(span_id))
L67
L68         span_context = SpanContext(
```

What we need is two test cases to cover the branch statement:
1. `if span_id` evaluates to True.
2. `if span_id` evaluates to False.

mhindery:
Added some extra tests, now all checks pass.

Issue #715, opened 2019-07-10
FileExporter should print each span new line


tjzo:
**Describe your environment.**
mac os
Python 3.6.5
opencensus==0.3.0

**Steps to reproduce.**
Just use FileExporter

**What is the expected behavior?**
Each span should be printed in separated line

**What is the actual behavior?**
All spans are printed in one line

**Additional context.**

c24t:
I'm surprised to see the file exporter overwrite the file on export, but it looks like all spans in the same trace should write to the same line. I'd expect the exporter to append to the output file and add linebreaks between traces.

PR #721, opened 2019-07-17
File exporter format fixes

c24t:
Quick fix for #715: make file exporter append instead of overwrite, and put each trace on a separate line.


opencensus/trace/file_exporter.py
```diff
@@ -37,13 +37,13 @@ class FileExporter(base_exporter.Exporter):
...
     :param file_mode: The file mode to open the output file with.
-                      Defaults to w+
+                      Defaults to a.
 
     """
 
     def __init__(self, file_name=DEFAULT_FILENAME,
                  transport=sync.SyncTransport,
-                 file_mode='w+'):
+                 file_mode='a'):
```
reyang: Nice, this actually aligns better with log handler `class logging.FileHandler(filename, mode='a', encoding=None, delay=False)`.

PR #722, opened 2019-07-18
Standard Metrics - Process CPU Usage

lzchen:
Add another standard metric for Azure exporter - Process CPU Usage.

Represents CPU Usage for the current process as a percentage.


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -77,6 +77,38 @@ def get_process_private_bytes_metric():
     return gauge
 
 
+def get_process_cpu_usage():
+    try:
+        # In the case of a process running on multiple threads on different CPU
+        # cores, the returned value of cpu_percent() can be > 100.0. The actual
+        # value that is return from this function should be capped at 100.
+        return min(PROCESS.cpu_percent(), 100.0)
```
reyang: Please check with Cijo Thomas on how we should handle this in a consistent way, probably we need to normalize against the cores.

lzchen: Right. AI-Node SDK does this too.

lzchen:
@reyang 
I talked to Cijo Thomas and also referred to the AI-Node SDK.
They both conform to the normalizing of the process cpu usage by dividing the value by the number of cores.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics.py
```diff
@@ -77,6 +77,39 @@ def get_process_private_bytes_metric():
     return gauge
 
 
+def get_process_cpu_usage():
+    try:
+        # In the case of a process running on multiple threads on different CPU
+        # cores, the returned value of cpu_percent() can be > 100.0. We
+        # normalize the cpu process using the number of logical CPUs
+        cpu_count = psutil.cpu_count(logical=True)
```
reyang: I'm not sure if we want to get cpu_count each time. Do we optimize for plug-and-play scenario?

lzchen: Good point. I can store the cpu_count information.

lzchen: @reyang  I will be refactoring the standard metrics into their own files in the next PR, so I can do this change then. I think this one is good to merge!

PR #724, opened 2019-07-22
Refactor Standard Metrics

lzchen:
This work involves moving different standard metrics types to their own files. It will allow easiness for developers to add their own standard metrics by simply creating a file and adding the module to the registry in the producer. It also allows for modularity of the standard metrics, where data exists only within the scope and calculated within those metrics that need them.


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/process.py
```diff
@@ -0,0 +1,83 @@
...
+            'Amount of memory process has used in bytes',
+            'byte',
+            [])
+        gauge.create_default_time_series(ProcessMemoryMetric.get_value)
+        return gauge
+
+
+class ProcessCPUMetric(object):
+    NAME = "\\Process(??APP_WIN32_PROC??)\\% Processor Time"
+    CPU_COUNT = psutil.cpu_count(logical=True)
```
reyang: Not sure about the PnP scenario https://www.kernel.org/doc/html/v4.14/core-api/cpu_hotplug.html.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/process.py
```diff
@@ -0,0 +1,83 @@
...
+    @staticmethod
```
reyang: Would it be better to make this a `@classmethod` then use `cls.CPU_COUNT` instead of hardcoding the class name again?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from opencensus.metrics.export.gauge import Registry
+from opencensus.metrics.export.metric_producer import MetricProducer
+from opencensus.ext.azure.metrics_exporter.standard_metrics.available_memory import AvailableMemoryMetric
```
reyang: The modules are using different naming rules, it will be great to have some consistency.
1. `available_memory` is adj. + noun.
2. `cpu_usage` is noun. + noun.
3. `process` is noun.

I don't have strong preference here, up to you.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/process.py
lzchen: Now looking more into when to use `@classmethod` I think it will be a good idea to use that instead of `@staticmethod` for all of the metrics, since `get_value` will be working with class specific data (and not isolated functions like `@staticmethod` is normally used for).

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/process.py
lzchen: Doing a bit more research, I actually think it will make sense to query the cpu_count for every function call.

1. psutil.cpu_count has very low overhead (it calls `os.sysconf("SC_NPROCESSORS_ONLN")` with no other overhead, see https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py under `cpu_count_logical()`. I also ran some preliminary timers on my local machine and the call times are instantaneous.

2. With cpu_count being constantly queried, the PnP scenario will be covered because we are asking the os for the cpu information.

3. Theoretically, the same argument for PnP can be made for any of the functions that utilize querying system information (like memory allocated). With that being said, the latency added for calling these functions every time is so minimal that it should be okay leaving them as is.

PR #725, opened 2019-07-23
Update Flask middleware to not use full URL with query params as a span name

Kami:
This pull request updates Flask middleware so it doesn't use full URL with query params as a spam name.

A lot of times query parameters can contain sensitive data (things such as API keys etc, since not all the services support sending such information via headers so query params are often used as a fallback in such scenarios).

We could perhaps add config option to include query parameters as an attribute with an option for blacklisted attributes (this does increase the complexity of the code and adds additional processing time and overhead to the middleware though).

## TODO

- [x] Tests

### DIFF
--- a/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
+++ b/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
@@ -146,7 +146,7 @@ class FlaskMiddleware(object):
             tracer.add_attribute_to_current_span(
                 HTTP_METHOD, flask.request.method)
             tracer.add_attribute_to_current_span(
-                HTTP_URL, str(flask.request.base_url))
+                HTTP_URL, str(flask.request.url))
             execution_context.set_opencensus_attr(
                 'blacklist_hostnames',
                 self.blacklist_hostnames)
--- a/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
+++ b/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
@@ -142,11 +142,11 @@ class FlaskMiddleware(object):
             # Set the span name as the name of the current module name
             span.name = '[{}]{}'.format(
                 flask.request.method,
-                flask.request.url)
+                flask.request.base_url)
             tracer.add_attribute_to_current_span(
                 HTTP_METHOD, flask.request.method)
             tracer.add_attribute_to_current_span(
-                HTTP_URL, str(flask.request.url))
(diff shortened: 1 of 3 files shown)


Kami:
I've added a test case and a changelog entry.

If others think it would be reasonable to add a config option (which defaults to ``False``) which allows query parameters to be stored as a span attribute (with an optional blacklist), I can add that.

Either as part of this PR or as a separate PR once this one is approved and merged.

contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
```diff
@@ -142,11 +142,11 @@ def _before_request(self):
             # Set the span name as the name of the current module name
             span.name = '[{}]{}'.format(
                 flask.request.method,
-                flask.request.url)
+                flask.request.base_url)
             tracer.add_attribute_to_current_span(
                 HTTP_METHOD, flask.request.method)
             tracer.add_attribute_to_current_span(
-                HTTP_URL, str(flask.request.url))
+                HTTP_URL, str(flask.request.base_url))
```
reyang: Using `method + " " + base_url` as span name looks good.

Using `base_url` in HTTP span attribute is **against** the [specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes).
I suggest creating a GitHub issue in the [spec repo](https://github.com/census-instrumentation/opencensus-specs) and send PR to update the specification, then implement the spec in Python SDK.

Kami: Thanks, I will update the code it will follow the specification.

reyang: Awesome! FYI we're also open to making the spec better (including privacy & security concerns), any recommendations/ideas are very welcomed.

Kami:
You are welcome.

Here is the revert - 55447e21da6fd9532e01df4ba5967e1271ca7f2c.

I will submit feedback against specification. Thanks.

reyang:
> I will submit feedback against specification. Thanks.

Great! And @Kami you're very welcomed to help us to design/review the [OpenTelemetry](https://github.com/open-telemetry/community) APIs. You can read more from https://opentelemetry.io, in a nutshell OpenCensus and OpenTracing are now merging into the same OpenTelemetry project under CNCF, and we'll be porting the majority of extension functionalities to OpenTelemetry.

reyang:
@Kami please help to rebase, thanks!

reyang:
This one has been superseded by #746.

PR #729, opened 2019-07-24
Standard Metrics - Outgoing Requests Per Second

lzchen:
Part of [#695]. Standard metric for outgoing requests per second. Outgoing requests is under "dependencies".
Tracks requests made from application using the [requests](https://2.python-requests.org/en/master/#) library. Other http libraries include urllib2 (python2 exclusive), urllib(python3 exclusive) and urllib3 (both) which might be introduced in the future.

Outgoing request rate is not part of Azure Monitor's standard metrics, so to see the data, view them under "customMetrics" in Kusto.

image


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -135,7 +135,10 @@ def _transmit_without_retry(self, envelopes):
             ['dc.services.visualstudio.com'],
         )
         try:
-            response = requests.post(
+            session = requests.Session()
+            # Ensure this request is not captured by standard metrics
+            session.disable_collection = True
```
reyang: This would only eliminate the requests from azure exporter, and applies to metrics (but not traces/logs) only?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/dependency.py
```diff
@@ -0,0 +1,94 @@
...
+                elapsed_seconds = current_time - last_time
+                interval_count = current_count - last_count
+                result = interval_count / elapsed_seconds
+            else:
+                result = 0
+            dependency_map['last_time'] = current_time
+            dependency_map['last_count'] = current_count
+            return result
+        except ZeroDivisionError:
+            logger.exception('Error handling get outgoing request rate. '
```
reyang: An alternative way is to return the previous value.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: This is an interesting topic. We need to define exactly which requests from what sources will be collected. The sources are, the user's application, only the metrics exporter in azure, everything in azure and everything in OpenCensus. With this we can ignore collections in certain ways:

1. Ignore only requests in metrics exporter
2. Ignore requests in Azure Exporter
3. Ignore all requests from OpenCensus (including other ext)

IMO as a user, when I instrument with OC, I would want none of the requests from OC to be tracked. So if I were to instrument with multiple exporters (stackdriver, azure, etc.) I wouldn't want those requests (if any) to be tracked in my metrics. With that being said, opencensus-ext-azure is the only extension that utilizes "requests", so as you mentioned, adding logic for disabling checks for everything in that extension would be needed. 

In the future, if other http libraries need to be supported, the same logic should apply to those.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/dependency.py
lzchen: That is true.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
markwolff: What are your thoughts on this having a more descriptive name with less chance for conflicts? e.g. prefixing with `oc_`?

lzchen: Possible. But I think we are going with the solution of using the RuntimeContext to track whether the request is coming from an exporter thread or not. The variable name will be changed appropriately.

lzchen:
This PR also includes the implementation of a thread context, named "is_exporter_thread", which is a flag set whenever the thread being run is one that has an exporter in it. This is to prevent spans and metrics being generated if the requests are being sent from an exporter.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -30,6 +31,8 @@
 from opencensus.ext.azure.common.transport import TransportMixin
 
 logger = logging.getLogger(__name__)
+if 'is_exporter_thread' not in RuntimeContext.snapshot().keys():
```
reyang: I guess there is a better place for this instead of the `log_exporter`.
Going forward in OpenTelemetry we're trying to remove the requirement of having to register slots.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/__init__.py
```diff
@@ -24,8 +26,9 @@
     import ProcessMemoryMetric
 
 # List of standard metrics to track
-STANDARD_METRICS = [AvailableMemoryMetric, ProcessCPUMetric,
-                    ProcessMemoryMetric, ProcessorTimeMetric]
+STANDARD_METRICS = [AvailableMemoryMetric, DependencyRateMetric,
+                    ProcessCPUMetric, ProcessMemoryMetric,
```
reyang: Personally I think it could be better to put each item on its own line, on strong opinion though.

contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
```diff
@@ -93,6 +102,15 @@ def call(url, *args, **kwargs):
 
 def wrap_session_request(wrapped, instance, args, kwargs):
     """Wrap the session function to trace it."""
+    try:
+        # Check if request was sent from an exporter. If so, do not wrap.
+        disable_from_exporter = RuntimeContext.is_exporter_thread
+        if disable_from_exporter:
```
reyang: Minor, would it be better to put `if RuntimeContext.is_exporter_thread:` directly?

lzchen: This is true.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/standard_metrics/__init__.py
lzchen: I like it better as well. Makes it less annoying to reformat and alphabetize.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
lzchen: Right. Instead of checking and changing the RuntimeContext for every exporter, we use a slot in the execution context now instead and simply import the module in each exporter.

opencensus/trace/execution_context.py
```diff
@@ -17,8 +17,14 @@
 
 _attrs_slot = RuntimeContext.register_slot('attrs', lambda: {})
 _current_span_slot = RuntimeContext.register_slot('current_span', None)
+_exporter_slot = RuntimeContext.register_slot('is_exporter_thread', False)
 _tracer_slot = RuntimeContext.register_slot('tracer', noop_tracer.NoopTracer())
 
+def is_exporter_thread():
```
reyang: Curious, why do we need this instead of using `RuntimeContext.is_exporter_thread` directly?

lzchen: 1. There should only be a single way to read/modify the same piece of data. It would be quite awkward to read using `RuntimeContext` and write using `execution_context`

2. Consistency with the other `execution_context` slots.

reyang: > There should only be a single way to read/modify the same piece of data. It would be quite awkward to read using RuntimeContext and write using execution_context

This single way would be `RuntimeContext.is_exporter_thread`.

> Consistency with the other execution_context slots.

We're in the direction to remove execution_context.

lzchen: It would be strange to have execution_context imported but not used.

reyang: I see. Okay for this PR, we should clean this up once we say goodbye to execution_context.

reyang: Please clean up https://github.com/census-instrumentation/opencensus-python/blob/1f92449067e2799b299eaa4015ba6eb73795ae34/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py#L45.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -148,11 +140,6 @@ def _transmit_without_retry(self, envelopes):
             # No retry policy, log output
             logger.warning('Transient client side error %s.', ex)
             return
-        finally:
```
reyang: It's great that we can say goodbye to this :)

PR #731, opened 2019-07-26
Support server performance breakdown by operation name [Do Not Merge]

### LINKED ISSUE #730: Azure Exporter performance endpoints not showing

Python 3.6 inside docker image ubuntu:16.04
Running with gunicorn

**Steps to reproduce.**

server.py
```
import os

from flask import Flask
from flask_cors import CORS

from api.bp1 import setup_blueprint as bp1_blueprint
from api.bp2 import setup_blueprint as bp2_blueprint
from helpers import CustomJSONEncoder

from logger import new

from opencensus.ext.flask.flask_middleware import FlaskMiddleware
from opencensus.ext.azure.trace_exporter import AzureExporter
from opencensus.trace.samplers import ProbabilitySampler

LOG = new(__name__)

config_integration.trace_integrations(['requests'])

app = Flask(__name__)
app.json_encoder = CustomJSONEncoder
CORS(app)
middleware = FlaskMiddleware(
    app,
    exporter=AzureExporter(),
    sampler=ProbabilitySampler(rate=1.0),
)


def register_blueprints(app):
    LOG.info("Registering blueprints")
    app.register_blueprint(bp1_blueprint())
    app.register_blueprint(bp2_blueprint())
    LOG.info("Blueprints registered")


if __name__ == "__main__":
    port = int(os.environ.get("PORT"))

    register_blueprints(app)

    LOG.info("Starting app")
    app.run(host='0.0.0.0', threaded=False, port=port)
```

wsgi.py
```
from server import app, register_blueprints

register_blueprints(app)

if __name__ == "__main__":
    app.run()
```

Running with command: `gunicorn --bind 0.0.0.0:$PORT --timeout 600 --chdir src wsgi:app`

Azure app insights to show each endpoint inside the performance tab.

image

reyang:
This should fix #730. @victoraugustolls

### DIFF
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
 ## Unreleased
+- Supported server performance breakdown by operation name
+  ([#731](https://github.com/census-instrumentation/opencensus-python/pull/731))
 
 ## 0.3.1
+Released 2019-06-30
 - Added metrics exporter
   ([#678](https://github.com/census-instrumentation/opencensus-python/pull/678))
 
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
@@ -79,6 +79,7 @@ class AzureExporter(TransportMixin, BaseExporter):
             if 'http.url' in sd.attributes:
                 data.name = data.name + ' ' + sd.attributes['http.url']
                 data.url = sd.attributes['http.url']
+                envelope.tags['ai.operation.name'] = data.name
             if 'http.status_code' in sd.attributes:
                 data.responseCode = str(sd.attributes['http.status_code'])
         else:


contrib/opencensus-ext-azure/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Changelog
 
 ## Unreleased
+- Supported server performance breakdown by operation name
```
reyang: @lzchen I just noticed the changelog is not reflecting your metrics work, please have this fixed, thanks.

victoraugustolls:
I ended up doing something similar locally, but I will explain the difference.

The `Operation Name`, as by Microsoft recommendation (see [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-model-request-telemetry) and [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-model-context)), should be something like `GET /application/path`.

For this to be accomplished, I edited the `Flask Middleware` to set the `http.path` attribute into the span, and use it in the operation name. This maintains the [specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes) and removes the root url from the operation name, and also the query parameters.

What do you think of this approach? If necessary, I can open a PR with the cited modifications.

reyang:
> The `Operation Name`, as by Microsoft recommendation (see [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-model-request-telemetry) and [here](https://docs.microsoft.com/en-us/azure/azure-monitor/app/data-model-context)), should be something like `GET /application/path`.

You're right. I also looked at this doc and was concerned about the multi-hosting scenario - if the flask service is listening on a public IP which has multiple domain names, we won't be able to distinguish them. Removing query parameter is a clear winner, but I'm not sure about the host. What do you think? I might need to check with the original author of that data model and get back here.

> For this to be accomplished, I edited the `Flask Middleware` to set the `http.path` attribute into the span, and use it in the operation name. This maintains the [specification](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes) and removes the root url from the operation name, and also the query parameters.
> 
> What do you think of this approach? If necessary, I can open a PR with the cited modifications.

This sounds like a good idea to me! One caveat is that we will need to update Django and Pyramid extensions as well.

reyang:
BTW, we plan to do another major release next week. If we miss the release train, we can do a minor release once this get resolved.

victoraugustolls:
> You're right. I also looked at this doc and was concerned about the multi-hosting scenario - if the flask service is listening on a public IP which has multiple domain names, we won't be able to distinguish them. Removing query parameter is a clear winner, but I'm not sure about the host. What do you think? I might need to check with the original author of that data model and get back here.

I think you are right, considering a multi-host scenario it might be easier to look by host. But also, the is the `CloudRole / Instance` attribute, that might be used to this? I could determine which requests were coming from `localhost` and which were coming from the cloud. But I guess it would be better to double check with the author.

> This sounds like a good idea to me! One caveat is that we will need to update Django and Pyramid extensions as well.

Those are simple changes! It can be made until next week I think. I will post below the changes I made in the `FlaskMiddleware`.

```
def _before_request(self):
        """A function to be run before each request.

        See: http://flask.pocoo.org/docs/0.12/api/#flask.Flask.before_request
        """
        # Do not trace if the url is blacklisted
        if utils.disable_tracing_url(flask.request.url, self.blacklist_paths):
            return

        try:
            span_context = self.propagator.from_headers(flask.request.headers)

            tracer = tracer_module.Tracer(
                span_context=span_context,
                sampler=self.sampler,
                exporter=self.exporter,
                propagator=self.propagator,
            )

            span = tracer.start_span()
            span.span_kind = span_module.SpanKind.SERVER
            # Set the span name as the name of the current module name
            span.name = "[{}]{}".format(
                flask.request.method, flask.request.url
            )
            tracer.add_attribute_to_current_span(
                "http.host", flask.request.host
            )
            tracer.add_attribute_to_current_span(
                HTTP_METHOD, flask.request.method
            )
            tracer.add_attribute_to_current_span(
                "http.path", flask.request.path
            )
            tracer.add_attribute_to_current_span(
                "http.route", flask.request.path
            )
            tracer.add_attribute_to_current_span(
                HTTP_URL, str(flask.request.url)
            )
            execution_context.set_opencensus_attr(
                "blacklist_hostnames", self.blacklist_hostnames
            )
        except Exception:  # pragma: NO COVER
            log.error("Failed to trace request", exc_info=True)
```

victoraugustolls:
Maybe we could have a parameter to `AzureExporter` saying `multi_host` to let the user choose?

reyang:
> Maybe we could have a parameter to `AzureExporter` saying `multi_host` to let the user choose?

Let's put multi_host aside for now, this can be treated as a feature addition. The more I think about it, the more I agree with your approach, we are basically deciding if we should fix a bug or add a feature, we can have this bug fixed first, then have another discussion around the multi-host support (also the current `url` already contains the host info, it is just not supported in the UI breakdown).

reyang:
> Those are simple changes! It can be made until next week I think. I will post below the changes I made in the `FlaskMiddleware`.

Let me know if you want me to work on Django/Pyramid part, I'd be glad to do it :)
And thanks for the suggestion and offering help!

contrib/opencensus-ext-azure/CHANGELOG.md
lzchen: Are you referring to the PR number?

victoraugustolls:
Okay! What do you think it’s the best? I open 2 PR’s, one for the flask middleware and one for the azure exporter, or just 1 for both?

reyang:
> Okay! What do you think it’s the best? I open 2 PR’s, one for the flask middleware and one for the azure exporter, or just 1 for both?

Up to you, either way works for me, thanks!

reyang:
Closing this PR as @victoraugustolls is working on a better solution.

PR #738, opened 2019-07-31
Added set_status to span, updated pymongo integration

victoraugustolls:
- Added `set_status` to spans, according to [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#status)

- Added `set_status_to_current_span` to `Tracer`

- Swapped pymongo integration attributes to specs described [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md)

Added tests do maintain 100% coverage

### DIFF
--- a/contrib/opencensus-ext-pymongo/CHANGELOG.md
+++ b/contrib/opencensus-ext-pymongo/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+- Changed attributes names to match [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md)
+([#738](https://github.com/census-instrumentation/opencensus-python/pull/738))
 
 ## 0.1.3
 Released 2019-05-31
--- a/contrib/opencensus-ext-pymongo/CHANGELOG.md
+++ b/contrib/opencensus-ext-pymongo/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-- Changed attributes names to match [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md)
+- Changed attributes names to make it compatible with [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md), maintaining OpenCensus specs fidelity
 ([#738](https://github.com/census-instrumentation/opencensus-python/pull/738))
 
 ## 0.1.3
--- a/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
+++ b/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
@@ -83,5 +83,6 @@ class MongoCommandListener(monitoring.CommandListener):
         self._stop(code_pb2.UNKNOWN, 'MongoDB error', event.failure)
 
     def _stop(self, code, message='', details=None):
-        self.tracer.set_status_to_current_span(code, message, details)
+        span = self.tracer.current_span()
+        span.set_status(code, message, details)
         self.tracer.end_span()
--- a/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
+++ b/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
@@ -16,25 +16,26 @@ import logging
 
 from pymongo import monitoring
 
+from google.rpc import code_pb2
+
 from opencensus.trace import execution_context
 from opencensus.trace import span as span_module
 
 
 log = logging.getLogger(__name__)
 
-MODULE_NAME = 'pymongo'
+MODULE_NAME = "pymongo"
 
-COMMAND_ATTRIBUTES = ['filter', 'sort', 'skip', 'limit', 'pipeline']
+COMMAND_ATTRIBUTES = ["filter", "sort", "skip", "limit", "pipeline"]
 
 
 def trace_integration(tracer=None):
     """Integrate with pymongo to trace it using event listener."""
-    log.info('Integrated module: {}'.format(MODULE_NAME))
+    log.info("Integrated module: {}".format(MODULE_NAME))
     monitoring.register(MongoCommandListener(tracer=tracer))
 
 
 class MongoCommandListener(monitoring.CommandListener):
-
     def __init__(self, tracer=None):
         self._tracer = tracer
 
@@ -44,11 +45,23 @@ class MongoCommandListener(monitoring.CommandListener):
 
     def started(self, event):
         span = self.tracer.start_span(
-            name='{}.{}.{}.{}'.format(MODULE_NAME,
-                                      event.database_name,
-                                      event.command.get(event.command_name),
-                                      event.command_name))
+            name="{}.{}.{}.{}".format(
+                MODULE_NAME,
+                event.database_name,
+                event.command.get(event.command_name),
+                event.command_name,
+            )
+        )
         span.span_kind = span_module.SpanKind.CLIENT
+
+        self.tracer.add_attribute_to_current_span('component', 'mongodb')
+        self.tracer.add_attribute_to_current_span('db.type', 'mongodb')
+        self.tracer.add_attribute_to_current_span(
+            'db.instance', event.database_name
+        )
+        self.tracer.add_attribute_to_current_span(
+            'db.statement', event.command.get(event.command_name)
+        )
 
         for attr in COMMAND_ATTRIBUTES:
             _attr = event.command.get(attr)
@@ -56,18 +69,19 @@ class MongoCommandListener(monitoring.CommandListener):
                 self.tracer.add_attribute_to_current_span(attr, str(_attr))
 
         self.tracer.add_attribute_to_current_span(
-            'request_id', event.request_id)
+            'request_id', event.request_id
+        )
 
         self.tracer.add_attribute_to_current_span(
-            'connection_id', str(event.connection_id))
+            'connection_id', str(event.connection_id)
+        )
 
     def succeeded(self, event):
-        self._stop('succeeded')
+        self._stop(code_pb2.OK)
 
     def failed(self, event):
-        self._stop('failed')
+        self._stop(code_pb2.UNKNOWN, 'MongoDB error', event.failure)
 
-    def _stop(self, status):
-        self.tracer.add_attribute_to_current_span('status', status)
-
+    def _stop(self, code, message='', details=None):
+        self.tracer.set_status_to_current_span(code, message, details)
         self.tracer.end_span()
--- a/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
+++ b/contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
@@ -24,14 +24,14 @@ from opencensus.trace import span as span_module
 
 log = logging.getLogger(__name__)
 
-MODULE_NAME = "pymongo"
+MODULE_NAME = 'pymongo'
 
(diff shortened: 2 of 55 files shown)


contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
```diff
@@ -16,25 +16,26 @@
...
 from opencensus.trace import span as span_module
 
 
 log = logging.getLogger(__name__)
 
-MODULE_NAME = 'pymongo'
+MODULE_NAME = "pymongo"
 
-COMMAND_ATTRIBUTES = ['filter', 'sort', 'skip', 'limit', 'pipeline']
+COMMAND_ATTRIBUTES = ["filter", "sort", "skip", "limit", "pipeline"]
```
reyang: Curious, is there a reason to change single quote to double quote?
If it is for consistency, I can still see lots of single quotes at line57, etc.

victoraugustolls: Missed that, sorry.

contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
```diff
@@ -44,30 +45,43 @@ def tracer(self):
...
+        )
         span.span_kind = span_module.SpanKind.CLIENT
 
+        self.tracer.add_attribute_to_current_span('component', 'mongodb')
+        self.tracer.add_attribute_to_current_span('db.type', 'mongodb')
+        self.tracer.add_attribute_to_current_span(
+            'db.instance', event.database_name
+        )
+        self.tracer.add_attribute_to_current_span(
+            'db.statement', event.command.get(event.command_name)
```
reyang: This is very useful :)

contrib/opencensus-ext-pymongo/opencensus/ext/pymongo/trace.py
victoraugustolls: PS: shouldn't the linter get this?

contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
```diff
@@ -90,12 +98,16 @@ def test_failed(self):
             'opencensus.trace.execution_context.get_opencensus_tracer',
             return_value=mock_tracer)
 
-        expected_attrs = {'status': 'failed'}
+        expected_attrs = {
```
reyang: `expected_status`?

victoraugustolls: Done

contrib/opencensus-ext-pymongo/tests/test_pymongo_trace.py
```diff
@@ -124,8 +136,14 @@ def start_span(self, name=None):
...
         span.name = name
         span.attributes = {}
+        span.status = {}
         self.current_span = span
         return span
 
     def add_attribute_to_current_span(self, key, value):
         self.current_span.attributes[key] = value
+
+    def set_status_to_current_span(self, code, message, details):
```
reyang: We might want to use `get_current_span` to get the span and use `Span.set_status`, instead of introducing a new API for Tracer. The fact that we had `add_attribute_to_current_span` has been corrected by [OpenTelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#getcurrentspan).

victoraugustolls: So `add_attribute_to_current_span` should be removed in the future? Will change `set_status` now.

victoraugustolls: Shouldn't `current_span` be renamed to `get_current_span` following the spec? Or the name is not mandatory?

reyang: We will use the correct name in OpenTelemetry https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-api/src/opentelemetry/trace/__init__.py#L208.

For OpenCensus, we cannot break the existing API unless we move to the next major version (which is OpenTelemetry).

victoraugustolls: > For OpenCensus, we cannot break the existing API unless we move to the next major version (which is OpenTelemetry).

Indeed

victoraugustolls: Fixed as requested! Please take a look!

opencensus/trace/span.py
```diff
@@ -346,6 +346,30 @@ def add_link(self, link):
             raise TypeError("Type Error: received {}, but requires Link.".
                             format(type(link).__name__))
 
+    def set_status(self, code, message='', details=None):
```
reyang: Please take a look at https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/status.py, we might need to consolidate them.

victoraugustolls: Do you think we should change the method to receive a `Status`? I followed the guidelines of method signature.

Or do you think we should eliminate the `Status` class?

reyang: Personally I want to omit `Status` class since it feels more Pythonic to me. We've also tried something similar in #602 to avoid introducing too many plain classes.

@c24t what do you think?

reyang: Looking at [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#set-status), it seems the `set_status` API should only take one argument, e.g. `Span.set_status(status: Status)`.
```
The Span interface MUST provide:
An API to set the Status where the new status is the only argument. This SHOULD be called SetStatus.
```

victoraugustolls: I do think the `Status` encapsulates some logic and I'm fine with using it.

It might be interesting to add some *default* values to the `Status` class, line `Ok`, `Unknown` and etc...

What do you think?

victoraugustolls: > Looking at the spec, it seems the set_status API should only take one argument, e.g. Span.set_status(status: Status).

How could I miss that... changing now.

victoraugustolls:
Should we update span to initiate with `OK` status? Reference [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#set-status)
> If used, this will override the default Span status, which is OK.

opencensus/trace/span.py
victoraugustolls: I resolved this, please take a look.

reyang: > It might be interesting to add some default values to the Status class, line Ok, Unknown and etc...
> What do you think?

I agree with you.

victoraugustolls:
> LGTM. Feel free to continue work on the default value for `Status` class, and let me know when you're ready to merge it.

Would it be okay to rewrite the `Status` class as this and fix the tests:

```

**Comment by victoraugustolls** [comment]
@reyang updated `Status` as described above, if you want do discuss it. I think the `Status` is a lot closer to spec, but maybe maintaining `code` and `message` as before, might be better for now. Let me know what you think.

**Comment by reyang** [comment]
> Would it be okay to rewrite the Status class as this and fix the tests:
> ...
> This is a lot closer to the specification.
> If it isn't ok because it is a breaking change making code and message private, we can leave them public until the migration, and then we just rename them in the future with the _ before.

It looks good to me! I don't know if the SDK consumers would use the status class directly. What they would do is to set status, and this is the place we might break them.

Given we're still in beta, I guess it is probably okay to have breaking changes for a small set of customers? @c24t for more input.

**Comment by c24t** [comment]
> Added `set_status` to spans, according to [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#status)

The change may be ok, but in general OpenCensus does not follow the OpenTracing spec. This isn't described in the [OC spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md).

File: contrib/opencensus-ext-pymongo/CHANGELOG.md [hunk 6]
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+- Changed attributes names to match [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md)
```
c24t: We shouldn't link to the OT spec here.

opencensus/trace/status.py
```diff
@@ -39,17 +39,31 @@ class Status(object):
                     See: https://cloud.google.com/trace/docs/reference/v2/
                          rest/v2/Status#FIELDS.details
     """
-    def __init__(self, code, message, details=None):
-        self.code = code
-        self.message = message
+    def __init__(self, code, message=None, details=None):
+        self._code = code
+        self._message = message
```
c24t: The changes to add `set_status` look good, but this looks like a breaking change for the sake of matching the names in a different spec.

victoraugustolls:
@c24t thanks for the review! Just some clarifications:

> The change may be ok, but in general OpenCensus does not follow the OpenTelemetry spec. This isn't described in the OC spec.

My idea was to don't do anything that breaks the OpenCensus Specs and change some internals to make an easier transition in the future (OpenTelemetry).

> The code and tests look great, but renaming Status.code and Status.message is a breaking change for no obvious benefit. LGTM otherwise.

Is it okay to maintain the `@property's` and the original `code` and `message` names? This does not breaks the OC specs and also make future transitions easier / OT compatible.

contrib/opencensus-ext-pymongo/CHANGELOG.md
reyang: I have slightly different opinion on this.
I agree that OpenCensus Python should follow OpenCensus spec. Following OpenTelemetry spec is not the goal here, but it is nice to have if:
1. It doesn't go against OpenCensus spec.
2. It doesn't introduce major breaking changes (given we're moving to OpenTelemetry soon) without a good rationale.
3. It helps us to form better understanding on our path to OpenTelemetry, or simply make the job easier when we try to steal some code and morph them into OpenTelemetry.

With these, I think having a link to OT spec is similar like saying "we are supporting W3C CorrelationContext version 2" - not something required in OC spec, but a nice to have feature.

victoraugustolls:
@reyang  @c24t  Is there anything pending here? Would be interesting to merge before releasing `0.7.0`!

reyang:
@victoraugustolls I'm fine with the change, and I've signed off.
@c24t would like to get your feedback before I merge this.

BTW, @victoraugustolls I've sent some messages to you on gitter, would you take a look? Thanks!

reyang:
@victoraugustolls we already have [0.7.0 release done](https://pypi.org/project/opencensus/0.7.0/).

We can consider a minor release (e.g. 0.7.1) if there is no breaking change.

c24t:
> Is it okay to maintain the `@property's` and the original `code` and `message` names? This does not breaks the OC specs and also make future transitions easier / OT compatible.

That sound great to me. Even though it's a minor change and doesn't technically run afoul of OC spec I think it's better to keep the existing names.

contrib/opencensus-ext-pymongo/CHANGELOG.md
c24t: I agree with all that, I'm just concerned about putting a link to the OT spec in something as public as the release notes. There's already a lot of confusion around OC/OT, linking the OT spec here might invite more.

victoraugustolls:
> That sound great to me. Even though it's a minor change and doesn't technically run afoul of OC spec I think it's better to keep the existing names.

Done, also I updated `pymongo` changelog, if you could take a look. Already updated with master. If I need to do anything else, please let me know!

c24t:
Looks great, thanks @victoraugustolls!

PR #739, opened 2019-07-31
Prepare for 0.7.0 release

lzchen:
Updates all versions of all packages to v0.7.0.

Updates necessary CHANGELOGs.

### DIFF
--- a/context/opencensus-context/CHANGELOG.md
+++ b/context/opencensus-context/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## 0.7.0
+Released 2019-07-31
+
 ## 0.1.1
 Released 2019-05-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## Unreleased
 
+## 0.7.0
+Released 2019-07-31
+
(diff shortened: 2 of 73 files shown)


context/opencensus-context/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## 0.7.0
+Released 2019-07-31
+
```
lzchen: Was wondering if we should put "-version update" under the notes or leave it as blank?

reyang: I don't have strong opinion, personally I would probably put "bumped version" here.

lzchen: Okay, seeing an empty space is awkward. I will make the change.

c24t:
Does it make sense to release updates to a bunch of packages without any code changes?

lzchen:
@c24t 
You bring up a good point. However, the main concern with having multiple different versions per package is that it's very difficult to maintain as contributors and complicated to debug as users. If the versions for all the packages are changed sooner than later, the better time we will have maintaining the code.

c24t:
LGTM if it doesn't include version updates for packages that haven't changed and don't rely on changes in dependencies.

lzchen:
Fixed in [#741]

PR #740, opened 2019-08-01
Bump version and Merge Changelog after 0.7.0 release

lzchen:


lzchen:
The change has been made so that only the packages that were changed have their versions bumped.

PR #743, opened 2019-08-02
Skip existing packages on twine upload

c24t:
Since we don't change all package version numbers on each release some uploads to PyPI will fail. Skip these with `--skip-existing`.

See https://circleci.com/gh/census-instrumentation/opencensus-python/2556?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link for an example of a failing build.

### DIFF
--- a/scripts/twine_upload.sh
+++ b/scripts/twine_upload.sh
@@ -41,5 +41,7 @@ done
 
 # Upload the distributions.
 for p in dist/* ; do
-  twine upload $p
+  twine upload --verbose $p
 done
+
+exit 0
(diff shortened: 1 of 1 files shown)


scripts/twine_upload.sh
```diff
@@ -41,5 +41,7 @@ done
 
 # Upload the distributions.
 for p in dist/* ; do
-  twine upload $p
+  twine upload --verbose $p
```
reyang: I think we should use [`--skip-existing`](https://github.com/pypa/twine/issues/115).

c24t: Oh that's much better, thanks.

PR #746, opened 2019-08-05
Requests library spec fidelity

victoraugustolls:
Updated `requests` integration for better fidelity, adding attributes:
- http host
- http method
- http path

Updated `span status` created by `requests` integration.
Updated `span name` to follow specs.

Added `http status code` to `grpc status code` mapping as an utility.

### DIFF
--- a/contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
+++ b/contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
@@ -23,6 +23,7 @@ except ImportError:
 from opencensus.trace import attributes_helper
 from opencensus.trace import execution_context
 from opencensus.trace import span as span_module
+from opencensus.trace import status as status_module
 from opencensus.trace import utils
 
 log = logging.getLogger(__name__)
@@ -33,8 +34,12 @@ REQUESTS_WRAP_METHODS = ['get', 'post', 'put', 'delete', 'head', 'options']
 SESSION_WRAP_METHODS = 'request'
 SESSION_CLASS_NAME = 'Session'
 
-HTTP_URL = attributes_helper.COMMON_ATTRIBUTES['HTTP_URL']
+HTTP_HOST = attributes_helper.COMMON_ATTRIBUTES['HTTP_HOST']
+HTTP_METHOD = attributes_helper.COMMON_ATTRIBUTES['HTTP_METHOD']
+HTTP_PATH = attributes_helper.COMMON_ATTRIBUTES['HTTP_PATH']
+HTTP_ROUTE = attributes_helper.COMMON_ATTRIBUTES['HTTP_ROUTE']
 HTTP_STATUS_CODE = attributes_helper.COMMON_ATTRIBUTES['HTTP_STATUS_CODE']
+HTTP_URL = attributes_helper.COMMON_ATTRIBUTES['HTTP_URL']
 
 
 def trace_integration(tracer=None):
@@ -74,22 +79,53 @@ def wrap_requests(requests_func):
         if utils.disable_tracing_hostname(dest_url, blacklist_hostnames):
             return requests_func(url, *args, **kwargs)
 
+        path = parsed_url.path if parsed_url.path else '/'
+
         _tracer = execution_context.get_opencensus_tracer()
         _span = _tracer.start_span()
-        _span.name = '[requests]{}'.format(requests_func.__name__)
+        _span.name = '{}'.format(path)
         _span.span_kind = span_module.SpanKind.CLIENT
+
+        # Add the requests host to attributes
+        _tracer.add_attribute_to_current_span(
+            HTTP_HOST, dest_url)
+
+        # Add the requests method to attributes
+        _tracer.add_attribute_to_current_span(
+            HTTP_METHOD, requests_func.__name__.upper())
+
+        # Add the requests path to attributes
+        _tracer.add_attribute_to_current_span(
+            HTTP_PATH, path)
 
         # Add the requests url to attributes
         _tracer.add_attribute_to_current_span(HTTP_URL, url)
 
-        result = requests_func(url, *args, **kwargs)
-
-        # Add the status code to attributes
-        _tracer.add_attribute_to_current_span(
-            HTTP_STATUS_CODE, str(result.status_code))
-
-        _tracer.end_span()
-        return result
+        try:
+            result = requests_func(url, *args, **kwargs)
+        except requests.Timeout:
+            _span.set_status(
+                status_module.Status(1, message='request timed out')
+            )
+        except requests.URLRequired:
+            _span.set_status(
+                status_module.Status(3, message='invalid URL')
+            )
+        except Exception as e:
+            _span.set_status(
+                status_module.Status(2, message=str(e))
+            )
+        else:
+            # Add the status code to attributes
+            _tracer.add_attribute_to_current_span(
+                HTTP_STATUS_CODE, str(result.status_code)
+            )
+            _span.set_status(
+                utils.status_from_http_code(result.status_code)
+            )
+            return result
+        finally:
+            _tracer.end_span()
 
     return call
 
@@ -113,10 +149,12 @@ def wrap_session_request(wrapped, instance, args, kwargs):
     if utils.disable_tracing_hostname(dest_url, blacklist_hostnames):
         return wrapped(*args, **kwargs)
 
+    path = parsed_url.path if parsed_url.path else '/'
+
     _tracer = execution_context.get_opencensus_tracer()
     _span = _tracer.start_span()
 
-    _span.name = '[requests]{}'.format(method)
+    _span.name = '{}'.format(path)
     _span.span_kind = span_module.SpanKind.CLIENT
 
     try:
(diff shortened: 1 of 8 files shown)


contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
```diff
@@ -74,22 +79,53 @@ def call(url, *args, **kwargs):
...
-        _tracer.end_span()
-        return result
+        # Add the requests url to attributes
+        _tracer.add_attribute_to_current_span(HTTP_URL, url)
+
+        try:
+            result = requests_func(url, *args, **kwargs)
+        except requests.Timeout:
+            _span.set_status(
+                status_module.Status(1, message='request timed out')
```
reyang: I guess `DEADLINE_EXCEEDED` might be a better fit. Do we know where to get clarification for such mapping?

BTW, it'll be helpful to use a named const than a literal int for readability, I wonder if the grpc lib provides these constant variables already.

reyang: It actually seems like a bug in the spec https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#client-errors-for-client-http-calls.

This is the gRPC error code https://grpc.github.io/grpc/core/md_doc_statuscodes.html.
Also in the server section `DEADLINE_EXCEEDED` does use 4 https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes.

contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
```diff
@@ -74,22 +79,53 @@ def call(url, *args, **kwargs):
...
+                status_module.Status(3, message='invalid URL')
+            )
+        except Exception as e:
+            _span.set_status(
+                status_module.Status(2, message=str(e))
+            )
+        else:
+            # Add the status code to attributes
+            _tracer.add_attribute_to_current_span(
+                HTTP_STATUS_CODE, str(result.status_code)
```
reyang: According to the spec, `status_code` is an int64. https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes

This again seems to be a spec bug in the [test-cases](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#test-cases) section, where `status_code` is put as a string.

opencensus/trace/utils.py
```diff
@@ -93,3 +95,40 @@ def disable_tracing_hostname(url, blacklist_hostnames=None):
...
+
+    :rtype: int
+    :returns: A instance of :class: `~opencensus.trace.status.Status`.
+    """
+    grpc_code = None
+    if http_code <= 199:
+        grpc_code = code_pb2.UNKNOWN
+    elif http_code >= 200 and http_code <= 399:
+        grpc_code = code_pb2.OK
+    elif http_code == 400:
```
reyang: Nit, potential way to simplify the flow.

```python
# validate that http_code is int

if http_code <= 199:
    return Status(code_pb2.UNKNOWN)

if http_code <= 399:
    return Status(code.pb2.OK)

grpc_code = {
    400: code_pb2.INVALID_ARGUMENT,
    401: code_pb2.UNAUTHENTICATED,
    # other code, sorted alphabetically
}.get(http_code, code_pb2.UNKNOWN)

return Status(grpc_code)

File: contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py (hunk 1 above)
Comment (victoraugustolls) [reply contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py]: I agree with using named constants! My doubt was if we should maps internal `requests` exceptions to OpenCensus exceptions, which are then mapped to GRPC codes, or importe the GRPC codes into the contribs.

File: contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py (hunk 2 above)
Comment (victoraugustolls) [reply contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py]: I think it is better to leave as int64, following specs. But open for discussion!

File: opencensus/trace/utils.py (hunk 3 above)
Comment (victoraugustolls) [reply opencensus/trace/utils.py]: Totally my fault, yours is way better, will change it.

Comment (victoraugustolls) [reply opencensus/trace/utils.py]: Done!

File: contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py (hunk 1 above)
Comment (victoraugustolls) [reply contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py]: Please take a look at the solution given!

File: opencensus/trace/exceptions_status.py [hunk 4]
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.rpc import code_pb2
+from opencensus.trace.status import Status
+
+DNS_RESOLUTION_FAILED = Status(code_pb2.UNKNOWN)
```
reyang: Where is `DNS_RESOLUTION_FAILED` used? Setting it to gRPC `UNKNOWN` could be misleading.

victoraugustolls: Is isn’t used but is listed in the [specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md)!

reyang: I see, it looks like this section in the specs is giving example rather than a comprehensive list of what should be handled/implemented.
I don't have strong opinion here, feel free to decide whether to keep it or not.

victoraugustolls: Yeah, they are just examples and I don’t agree with the unknown value also, but I think with OpenTelemetry we must have a stronger opinion about this! Will remove it and push.

opencensus/trace/exceptions_status.py
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.rpc import code_pb2
+from opencensus.trace.status import Status
+
+CANCELLED = Status(code_pb2.CANCELLED)
+INVALID_URL = Status(code_pb2.INVALID_ARGUMENT, message='invalid URL')
+TIMEOUT = Status(code_pb2.DEADLINE_EXCEEDED, message='request timed out')
```
c24t: @victoraugustolls there's an ongoing discussion about using grpc error codes in OpenTelemetry, if you've got an opinion here you may want to tune into the SIG meeting or follow the OT specs repo.

contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
victoraugustolls: > According to the spec, `status_code` is an int64. https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes

Right now `AzureExporter` for tracer is broken because `Flask`, `Django` and `Pyramid` integrations set `status_code` as string, and azures expect int, we can typecast on azure or change the other integrations, which way is preferred?

reyang: https://github.com/census-instrumentation/opencensus-python/blame/master/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py#L86

My understanding is that we'll still get the right `status_code`, but the `success` flag will be wrong.

victoraugustolls: > My understanding is that we'll still get the right `status_code`, but the `success` flag will be wrong.

Yes, and with this we have both options described above. I can take any of them, I prefer to put the `status_code` as int, it makes more sense to me.

reyang: I agree with you that we should use int. Converting int to string and then parse it to int sounds like crazy :)

victoraugustolls: I will open a new PR to fix this into `Flask`, `Django` and `Pyramid` integrations.

victoraugustolls: Did it [here](https://github.com/census-instrumentation/opencensus-python/pull/755).

contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
```diff
@@ -127,14 +159,37 @@ def wrap_session_request(wrapped, instance, args, kwargs):
...
+
+    # Add the requests url to attributes
+    _tracer.add_attribute_to_current_span(HTTP_URL, url)
 
-    _tracer.end_span()
-    return result
+    try:
+        result = wrapped(*args, **kwargs)
+    except requests.Timeout:
+        _span.set_status(exceptions_status.TIMEOUT)
```
isra17: @reyang  Isn't this `except` clause and the ones following going to eat the raised exception and returns `None` in case of any `requests` error? Unless I'm missing something this change is going to break many things.

victoraugustolls: Yes, but wont `requests` return `None` in case it raises an `Exception`? If so, the returned value is the same

isra17: Raising an exception should not "return" at all.

```python
try:
  requests.get("http://not-a-host")
  unreacheable() #actually reached with this PR
except Exception:
  reached() # actually unreachable with this PR
```

victoraugustolls: What can be changed, and is a good idea actually, is to threat the `Exception` and raise it again, and let the user do whatever it wants. That's what you meant?

contrib/opencensus-ext-requests/tests/test_requests_trace.py
```diff
@@ -204,6 +217,144 @@ def test_wrap_requests_exporter_thread(self):
...
+            'opencensus.ext.requests.trace.execution_context.'
+            'is_exporter',
+            return_value=False)
+
+        wrapped = trace.wrap_requests(mock_func)
+
+        url = 'http://localhost:8080/test'
+
+        with patch, patch_thread:
+            wrapped(url)
```
isra17: To be clearer, given `mock_func` raise an exception, this test case should fail gith here and never go further. The test should actually assert the `requests.TooManyRedirects` exception is raised from `wrapped(url)`.

contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py
isra17: This should be fixed by adding a `raise` statement in every `except` block.

isra17: The main goal here is to keep this hook transparent and keep 1:1 interface as `requests` (Exceptions are part of that interface).

victoraugustolls: > What can be changed, and is a good idea actually, is to threat the Exception and raise it again, and let the user do whatever it wants. That's what you meant?

So you agreed with this, right? Just to be clear. If so, I do agree with you and I will open a PR to change this.

isra17: Yes :+1: !

reyang: @isra17 You're right, we've missed this. It is a bug.

PR #749, opened 2019-08-05
Bump SD exporter to 0.7.1 for release

c24t:
#747 bumped `opencensus-ext-exporter` to `0.5.0`, but to be consistent with the core library releases this should be an `0.7.x` release. This PR bumps the core library and the exporter to `0.7.1` so we can release them together.

@reyang I think this is the most consistent way to version the packages even though it means changing the version number of the core library without any changes.

### DIFF
--- a/contrib/opencensus-ext-django/CHANGELOG.md
+++ b/contrib/opencensus-ext-django/CHANGELOG.md
@@ -5,8 +5,9 @@
 ## 0.7.0
 Released 2019-07-31
 
-- Updated span attributes to include some missing attributes listed [here](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes)
-([#735](https://github.com/census-instrumentation/opencensus-python/pull/735))
+- Updated span attributes to include some missing attributes listed
+  [here](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes)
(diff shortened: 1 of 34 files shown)


reyang:
@c24t I prefer to keep openceusus `0.7.0` and use `0.7.1` for StackDriver exporter. While I don't have strong objection on updating opencensus to `0.7.1` even if there is no code changes, feel free to decide.

c24t:
We can pick up #738 in this release too since it's a non-breaking change, it just missed the `0.7.0` release.

c24t:
We got confirmation from @victoraugustolls in #738.

contrib/opencensus-ext-django/CHANGELOG.md
```diff
@@ -5,8 +5,9 @@
 ## 0.7.0
 Released 2019-07-31
 
-- Updated span attributes to include some missing attributes listed [here](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes)
-([#735](https://github.com/census-instrumentation/opencensus-python/pull/735))
+- Updated span attributes to include some missing attributes listed
```
reyang: Need to update version files for these extensions: `opencensus-ext-django`, `opencensus-ext-flask`, `opencensus-ext-jaeger`, `opencensus-ext-ocagent`, `opencensus-ext-pymongo`.

PR #755, opened 2019-08-06
Hotfix/django flask pyramid status code

victoraugustolls:
Updating `http.status_code` to be an int, as discussed with @reyang. This affects the following contributions:
- Django
- Flask
- Httplib
- Pyramid

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Updated `django`, `flask`, `httplib` and `pyramid` modules.
 
 ## 0.7.1
 Released 2019-08-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Updated `django`, `flask`, `httplib` and `pyramid` modules.
+  ([#755](https://github.com/census-instrumentation/opencensus-python/pull/755))
 
 ## 0.7.1
 Released 2019-08-05
--- a/contrib/opencensus-ext-django/CHANGELOG.md
+++ b/contrib/opencensus-ext-django/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Updated `http.status_code` attribute to be an int.
+  ([#755](https://github.com/census-instrumentation/opencensus-python/pull/755))
 
 ## 0.7.0
 Released 2019-07-31
--- a/contrib/opencensus-ext-django/CHANGELOG.md
+++ b/contrib/opencensus-ext-django/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Updated `http.status_code` attribute to be an int.
 
 ## 0.7.0
 Released 2019-07-31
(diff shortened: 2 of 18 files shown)


reyang:
I haven't taken a deep look into the system test failures. It seems some of the system tests are looking for span labels instead of attributes, and label value is always string.

victoraugustolls:
I can update the tests to get the attributes instead, if possible!

reyang:
I haven't done the archaeology, would be curious to understand why we had labels instead of attributes in the first place.

victoraugustolls:
Who might know this? And is this a block for merging? If so, I can either swap for attributes or try to understand why we are using labels!

reyang:
I don't think it's blocking unless @c24t says otherwise.

victoraugustolls:
Waiting for his input so, and if is a block, I would please ask if @c24t can give an explanation of why we are using labels and if there is any workaround!

victoraugustolls:
@c24t any comments here?

victoraugustolls:
Pinging @c24t, if there are no responses, should we merge @reyang ?

reyang:
@victoraugustolls sorry for my delayed response. We can move forward by changing the test cases to use attributes instead of labels. Once the CI passed, we're good to merge.

victoraugustolls:
Will update later today!

victoraugustolls:
> @victoraugustolls sorry for my delayed response. We can move forward by changing the test cases to use attributes instead of labels. Once the CI passed, we're good to merge.

As we are using google cloud spans that does not contain `attributes`, only `labels`, this is not possible. The strangest thing to me is that in the stackdriver exporter tests, it was already expecting an int as status code, but it was receiving an string before my changes. I could add an special case for the stackdriver exporter where status code would be converted from int  to string, but don't know if this is desired. What do you think @c24t @reyang ?

reyang:
> The strangest thing to me is that in the stackdriver exporter tests, it was already expecting an int as status code, but it was receiving an string before my changes. I could add an special case for the stackdriver exporter where status code would be converted from int  to string, but don't know if this is desired. What do you think @c24t @reyang ?

@c24t @songy23 Looks like int should be the correct answer:
* https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-stackdriver/opencensus/ext/stackdriver/trace_exporter/__init__.py#L253
* https://cloud.google.com/trace/docs/reference/v2/rest/v2/Status

victoraugustolls:
@songy23 and what about the failing system tests?

victoraugustolls:
A course of action is needed here, system tests fails but the changes were approved. @c24t @songy23 @reyang

reyang:
It seems that Stackdriver labels are always strings. @victoraugustolls would the system test pass if you convert status from int to str in the Stackdriver exporter?

victoraugustolls:
Google client only accepts labels as strings due to the proto format, and during the conversion inside the client, it loses the status code. If I transform the status code to string, it would work, but wouldn’t be “right”. But I think this could be tracked in another issue so we can move this forward!

victoraugustolls:
@c24t @songy23 @reyang added tests to see if status code was being converted correctly, but still not enough for the system tests. Can someone take a look?

victoraugustolls:
Merged @reyang solution, thanks for the help!
@songy23  @c24t  we are receiving `429` from Google Client here, any tips?

PR #757, opened 2019-08-07
Django middleware: allow autogenerated span name override

mhindery:
In the Django Middleware, a span is created with an autogenerated name. The name setting is currently part of the `process_view()` function which also contains other logic for the middleware functionality, so it's not easy / safe to override this method for the purpose of customizing the span naming. With the Django new-style middleware in the future, there won't be a `process_view()` method alltogether (see https://github.com/census-instrumentation/opencensus-python/blob/63e2c97682ba9aea2dd0668ff3ac338581e683da/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py#L123), so people having customized the process_view method for the purpose of span naming will have an issue.

This PR adds a separate span naming method to the middleware as a hook to allow end-user customization. It's sole purpose is to generate a span name, which gives people an easy way to customize the naming convention by subclassing the middleware and overriding this single method without impacting the other functioning of the middleware (compatible with both this style and new-style middleware).

### DIFF
--- a/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
+++ b/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
@@ -188,7 +188,10 @@ class OpencensusMiddleware(MiddlewareMixin):
         except Exception:  # pragma: NO COVER
             log.error('Failed to trace request', exc_info=True)
 
-    def process_view(self, request, view_func, *args, **kwargs):
+    def get_span_name(self, request):
+        return utils.get_func_name(request.resolver_match.func)
+
+    def process_view(self, request, *args, **kwargs):
         """Process view is executed before the view function, here we get the
         function name add set it as the span name.
         """
@@ -202,7 +205,7 @@ class OpencensusMiddleware(MiddlewareMixin):
             # function name of the request.
             tracer = _get_current_tracer()
             span = tracer.current_span()
-            span.name = utils.get_func_name(view_func)
+            span.name = self.get_span_name(request)
         except Exception:  # pragma: NO COVER
             log.error('Failed to trace request', exc_info=True)
 
--- a/contrib/opencensus-ext-django/tests/test_django_middleware.py
+++ b/contrib/opencensus-ext-django/tests/test_django_middleware.py
@@ -32,6 +32,17 @@ class TestOpencensusMiddleware(unittest.TestCase):
     def setUp(self):
         from django.conf import settings as django_settings
         from django.test.utils import setup_test_environment
+        from opencensus.ext.django import middleware
+        from django.http import HttpResponse
+        from django.views import View
+
+        self.middleware_kls = middleware.OpencensusMiddleware
+
+        class MockViewOk(View):
+            def get(self, *args, **kwargs):
+                return HttpResponse(status=200)
+
+        self.view_func_ok = MockViewOk.as_view()
 
         if not django_settings.configured:
             django_settings.configure()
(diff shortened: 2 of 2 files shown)


reyang:
@mhindery Would you help to explain the scenario here? Why would the user want to change the span name?
In general, we're trying to follow the [spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#spans).

mhindery:
Our usage is that we're using the package for tracing with a set of service in Python & Go, and sending the traces to StackDriver.

- When your traces span services, it's not clear which service is responsible for a trace (the path is not useful for that). We prepend our service name and environment/deployment name to the span name so we can quickly spot which service is leading to a specific span. Having to click through on each span to look for possible attributes or labels on it is not feasible with a number of different service along the road. E.g. this is how a decent-sized trace looks like in the trace list: 
image and 
image

    - Some integrations with http libraries prefix their outgoing request spans with `Sent.` and in that light we add the counterpart `Recv.<service_name>` to indicate when a service starts with a request, so it shows the boundaries between services. Without it, using the path alone, it's very hard to spot which service is doing something in these large traces.
    - Not all spans contains a correct attribute to easily identify the service / deployment which is generating the trace.
- Having a name prefix allows for searching traces from a certain service. I can't speak for other platforms, but StackDriver does not allow to filter on labels or attributes, only by name.
- The opencensus Go package also does provide a hook for easily customizing the span name (https://godoc.org/go.opencensus.io/plugin/ochttp#Handler, the `FormatSpanName ` function) which in our case looks like this:

```go
	tracingWrapper := func(handler http.Handler) http.Handler {
		incomingSpanNamer := func(req *http.Request) string {
			return fmt.Sprintf("GoAPI.%s: %s", deployment, req.URL.Path)
		}

		return &ochttp.Handler{
			Propagation:    &propagation.HTTPFormat{},
			Handler:        handler,
			FormatSpanName: incomingSpanNamer,
		}
	}
```

I fully agree to have the default behavior not changed following the spec, but having a hook to allow customization when desired for somebody's use-case seems good.

reyang:
Thanks for the background. Now I understand the scenario.
I'm not convinced that we want to do this in OpenCensus unless we **also plan to support this in OpenTelemetry**. My worry is that we end up creating more public APIs for specific scenario instead providing generic primitives.

I think what you're looking for is something that can be done at exporter level (e.g. If Stackdriver wants to format the span names in a different way, the exporter `opencensus-ext-stackdriver` can do it). Alternatively, we can explore if this can be done using [span processor](https://github.com/open-telemetry/opentelemetry-specification/pull/205) - having span processor take the span data, transform to a new name, forward to exporter.

reyang:
@c24t @songy23 You know Stackdriver better, please help to share some insights. Thanks.

mhindery:
If you start a span yourself with `start_span()`, you can choose any name for it as well, according to your desired policy. I can't speak for other languages except for python & go, so I don't know whether they all have public methods for it, but in OpenCensus-Go I can choose whatever name I want both for an incoming server request and application-generated spans. Why not allow it here? Right now this is the only place in the chain of services in a request where the name cannot easily be chosen. 

Having it 'hardcoded' to a view is very restrictive, especially since with the dynamic nature of python & django views; you can have several structures and hierarchies where the initial View name isn't necessarily what you want as a representation. A concrete example from one of the codebases I'm working with is

```python
# urls.py file
    url(r'^api/v3/endpoint_cats/?', ApiView.as_view(input_generator=CatInputGenerator, <other_args_here>), name="apiv3_cats"),
    url(r'^api/v3/endpoint_dogs/?', ApiView.as_view(input_generator=DogInputGenerator, <other_args_here>), name="apiv3_dogs"),
    url(r'^api/v3/endpoint_birds/?', ApiView.as_view(input_generator=BirdsInputGenerator, <other_args_here>), name="apiv3_birds"),
    # a lot of other url entries ...
```

In this case `ApiView` is a view containing common logic for all api requests (think rate limiting, security, caching ...), but with a component as an argument which defines the use of the endpoint. All the endpoints here would have the same span name being `package_name.ApiView` while they should be distinct. In such a case, you should be able to use a different naming convention than the default one, as that just doesn't fit all setups. Regardless of 'you can do this example from above differently like this or that'; most of the time there are certain reasons for certain structures and setups, and this middleware is meant for all django projects so it should be able to handle these possibilities.

mhindery:
Closing this in favor of development on the opentelemetry project: https://github.com/open-telemetry/opentelemetry-python/pull/571

PR #758, opened 2019-08-08
Standard Metrics - Incoming Requests Per Second 

lzchen:
Part of [#695]. Standard metric for incoming requests per second. Incoming requests are under "http_requests".
Track incoming requests to an HTTPServer on the application. Currently, incoming requests are shown "performanceCounters" table in Kusto.

image


lzchen:
If you look at the attached image, the category for this metric is "ASP.NET application" which is actually not accurate. Maybe it would be more appropriate to put this as a custom metric instead.

reyang:
BTW, remember to update the changelog.

PR #759, opened 2019-08-10
Fixes value for `http.route` in Flask middleware

olance:
The [OpenCensus specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes) suggest that the `http.route` attribute should contain the URL template/route that was matched for the current request.  
Here, the request's _path_ is used, making it a duplicate of `http.path`.

Having the `http.route` correctly reported is quite useful to easily filter for all traces of a same endpoint.  
On Stackdriver Trace for instance, one can only filter by prefix or exact match, making it impossible to filter for endpoints like `/resources/<resource_id>/whatever` unless you've got a common denominator such as the route itself.

This PR fixes the reported attribute value for `http.route` in the Flask middleware by moving the addition of the attribute to the `after_request` handler, where `flask.request.url_rule` is available.

Tests are fixed/improved accordingly

### DIFF
--- a/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
+++ b/contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
@@ -156,9 +156,6 @@ class FlaskMiddleware(object):
                 HTTP_PATH, flask.request.path
             )
             tracer.add_attribute_to_current_span(
-                HTTP_ROUTE, flask.request.path
-            )
-            tracer.add_attribute_to_current_span(
                 HTTP_URL, str(flask.request.url)
             )
             execution_context.set_opencensus_attr(
@@ -179,6 +176,9 @@ class FlaskMiddleware(object):
 
         try:
             tracer = execution_context.get_opencensus_tracer()
+            tracer.add_attribute_to_current_span(
+                HTTP_ROUTE, flask.request.url_rule.rule
+            )
             tracer.add_attribute_to_current_span(
                 HTTP_STATUS_CODE,
                 str(response.status_code)
--- a/contrib/opencensus-ext-flask/tests/test_flask_middleware.py
+++ b/contrib/opencensus-ext-flask/tests/test_flask_middleware.py
@@ -51,6 +51,10 @@ class TestFlaskMiddleware(unittest.TestCase):
         def index():
             return 'test flask trace'  # pragma: NO COVER
 
+        @app.route('/wiki/<entry>')
+        def wiki(entry):
+            return 'test flask trace'  # pragma: NO COVER
+
         @app.route('/_ah/health')
         def health_check():
             return 'test health check'  # pragma: NO COVER
@@ -155,7 +159,6 @@ class TestFlaskMiddleware(unittest.TestCase):
                 'http.host': u'localhost',
                 'http.method': u'GET',
                 'http.path': u'/wiki/Rabbit',
-                'http.route': u'/wiki/Rabbit',
                 'http.url': u'http://localhost/wiki/Rabbit',
             }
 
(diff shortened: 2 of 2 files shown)


olance:
@googlebot I fixed it.

contrib/opencensus-ext-flask/tests/test_flask_middleware.py
```diff
@@ -278,13 +279,37 @@ def test__after_request_sampled(self):
         flask_trace_id = '00-{}-{}-00'.format(trace_id, span_id)
 
         app = self.create_app()
-        flask_middleware.FlaskMiddleware(app=app)
+        flask_middleware.FlaskMiddleware(
+            app=app,
+            sampler=samplers.AlwaysOnSampler()
```
olance: Note: Adding the `AlwaysOnSampler` made sure the request was always sampled, which did not seem to be the case before

reyang:
@victoraugustolls please help to review this change.

contrib/opencensus-ext-flask/tests/test_flask_middleware.py
victoraugustolls: @reyang don't know if we should do this, this was changed by @c24t [here](https://github.com/census-instrumentation/opencensus-python/pull/639). Waiting for his opinion on this.

victoraugustolls:
Other than my comment on the sampler changes, everything looks great.

contrib/opencensus-ext-flask/tests/test_flask_middleware.py
victoraugustolls: See comments [here](https://github.com/census-instrumentation/opencensus-python/pull/639#issuecomment-488202973)

olance: @victoraugustolls I’m aligned with @c24t comments regarding sampling on a production app, however I’m not sure how that’s a problem in tests?

The thing is, until I did this the tests wouldn’t pass because the default sampler has such a low sampling probability that the test span would actually never be created. I do think in the case of unit tests, we’d rather have deterministic execution than chance-driven tests :slightly_smiling_face:

victoraugustolls: Even being tests, I would change for 100% Probability Sampler because:

> AlwaysOnSampler is redundant though, and AFAIK just exists for completeness.

So who knows, in the future it may change, although I don't think it will!

victoraugustolls: > @victoraugustolls I’m aligned with @c24t comments regarding sampling on a production app, however I’m not sure how that’s a problem in tests?
> 
> The thing is, until I did this the tests wouldn’t pass because the default sampler has such a low sampling probability that the test span would actually never be created. I do think in the case of unit tests, we’d rather have deterministic execution than chance-driven tests slightly_smiling_face

Wow, what a luck :D, never happened to me!

reyang: +1 on having more deterministic execution in test case.

contrib/opencensus-ext-flask/opencensus/ext/flask/flask_middleware.py
```diff
@@ -179,6 +176,9 @@ def _after_request(self, response):
 
         try:
             tracer = execution_context.get_opencensus_tracer()
+            tracer.add_attribute_to_current_span(
+                HTTP_ROUTE, flask.request.url_rule.rule
```
jtbeach: Note that this will throw if url_rule is None, which it will for any 404 route. I have raised https://github.com/census-instrumentation/opencensus-python/issues/777 for this

PR #767, opened 2019-08-19
Implement Connection Strings for Azure Exporters

lzchen:
Implements connection strings as defined by the spec.

We allow the user to provide an instrumentation_key value and connection_string value explicitly, as well as have the environment variable counterparts to these.

Connection strings currently provide another way to derive the instrumentation key as well as define an ingestion endpoint. 
The priority for instrumentation key:
1. from explicit connection string in code
2. from explicit ikey in code
3. from environment variable connection string
4. from environment variable ikey

The priority for endpoint:
1. from explicit connection string in code
2. from environment variable connection string
3. Default Breeze endpoint

TODO:
Link to documentation for connection strings when created

### DIFF
--- a/contrib/opencensus-ext-azure/README.rst
+++ b/contrib/opencensus-ext-azure/README.rst
@@ -24,9 +24,8 @@ The **Azure Monitor Log Handler** allows you to export Python logs to `Azure Mon
 This example shows how to send a warning level log to Azure Monitor.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
-* Place your instrumentation key in a connection string and into ``APPLICATIONINSIGHTS_CONNECTION_STRING`` environment variable.
-* You can also put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
-* Alternatively, you can specify either the connection string or instrumentation key directly in your code, which will take priority over a set environment variable.
+* Place your instrumentation key in a `connection string` and directly into your code.
+* Alternatively, you can specify your `connection string` in an environment variable ``APPLICATIONINSIGHTS_CONNECTION_STRING``.
 
 .. code:: python
 
@@ -35,15 +34,16 @@ This example shows how to send a warning level log to Azure Monitor.
     from opencensus.ext.azure.log_exporter import AzureLogHandler
 
     logger = logging.getLogger(__name__)
-    logger.addHandler(AzureLogHandler())
+    logger.addHandler(AzureLogHandler(connection_string='InstrumentationKey=<your-instrumentation_key-here>'))
     logger.warning('Hello, World!')
+
+* Alternatively, you can specify your `connection string` in an environment variable ``APPLICATIONINSIGHTS_CONNECTION_STRING``.
 
 You can enrich the logs with trace IDs and span IDs by using the `logging integration <../opencensus-ext-logging>`_.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
-* Place your instrumentation key in a connection string and into ``APPLICATIONINSIGHTS_CONNECTION_STRING`` environment variable.
-* You can also put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
-* Alternatively, you can specify either the connection string or instrumentation key directly in your code, which will take priority over a set environment variable.
+* Place your instrumentation key in a `connection string` and directly into your code.
+* Alternatively, you can specify your `connection string` in an environment variable ``APPLICATIONINSIGHTS_CONNECTION_STRING``.
 
 .. code:: python
 
@@ -59,16 +59,19 @@ You can enrich the logs with trace IDs and span IDs by using the `logging integr
 
     logger = logging.getLogger(__name__)
 
-    handler = AzureLogHandler()
+    handler = AzureLogHandler(connection_string='InstrumentationKey=<your-instrumentation_key-here>')
     handler.setFormatter(logging.Formatter('%(traceId)s %(spanId)s %(message)s'))
     logger.addHandler(handler)
 
-    tracer = Tracer(exporter=AzureExporter(), sampler=ProbabilitySampler(1.0))
+    tracer = Tracer(
+        exporter=AzureExporter(connection_string='InstrumentationKey=<your-instrumentation_key-here>'),
+        sampler=ProbabilitySampler(1.0)
+    )
 
     logger.warning('Before the span')
     with tracer.span(name='test'):
         logger.warning('In the span')
-    logger.warning('After the span')
+    logger.warning('After the span')s
 
 Metrics
 ~~~~~~~
@@ -76,9 +79,8 @@ Metrics
 The **Azure Monitor Metrics Exporter** allows you to export metrics to `Azure Monitor`_.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
-* Place your instrumentation key in a connection string and into ``APPLICATIONINSIGHTS_CONNECTION_STRING`` environment variable.
-* You can also put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
-* Alternatively, you can specify either the connection string or instrumentation key directly in your code, which will take priority over a set environment variable.
+* Place your instrumentation key in a `connection string` and directly into your code.
+* Alternatively, you can specify your `connection string` in an environment variable ``APPLICATIONINSIGHTS_CONNECTION_STRING``.
 
 .. code:: python
 
@@ -107,7 +109,7 @@ The **Azure Monitor Metrics Exporter** allows you to export metrics to `Azure Mo
     def main():
         # Enable metrics
         # Set the interval in seconds in which you want to send metrics
-        exporter = metrics_exporter.new_metrics_exporter()
+        exporter = metrics_exporter.new_metrics_exporter(connection_string='InstrumentationKey=<your-instrumentation-key-here>')
         view_manager.register_exporter(exporter)
 
         view_manager.register_view(CARROTS_VIEW)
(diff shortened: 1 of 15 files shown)


contrib/opencensus-ext-azure/README.rst
```diff
@@ -25,7 +25,9 @@ This example shows how to send a warning level log to Azure Monitor.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
 * Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
-* You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
+* You can also utilize a connection string with an instrumentation key. Place the connetion string in ``APPLICATIONINSIGHTS_CONNECTION_STRING`` environment variable, which will take priority over the instrumentation key environment variable.
```
markwolff: connetion --> connection

contrib/opencensus-ext-azure/README.rst
```diff
@@ -76,7 +80,9 @@ The **Azure Monitor Metrics Exporter** allows you to export metrics to `Azure Mo
...
```
markwolff: see above

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,76 @@
...
+        return {}
+    try:
+        pairs = connection_string.split(';')
+        result = dict(s.split('=') for s in pairs)
+    except Exception:
+        raise ValueError('Invalid connection string: ' + connection_string)
+    # Validate authorization
+    auth = result.get(AUTHORIZATION)
+    if auth is None:
+        raise ValueError('Missing \'Authorization\' in connection string:'
```
markwolff: This can be a valid connection string as per
https://github.com/microsoft/ApplicationInsights-dotnet/pull/1187/files#diff-7421a28dc03eceb294044ed77594ec83R12-R18

contrib/opencensus-ext-azure/README.rst
reyang: If we have connection string, we should put it as the recommended approach (1st choice) and use it across all examples.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,76 @@
...
+
+
+def parse_connection_string(connection_string):
+    if connection_string is None:
+        return {}
+    try:
+        pairs = connection_string.split(';')
+        result = dict(s.split('=') for s in pairs)
+    except Exception:
+        raise ValueError('Invalid connection string: ' + connection_string)
```
TimothyMothra: Be aware that a user's connection string may contain sensitive information. You may not want to emit its full contents in an exception.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
TimothyMothra: The initial version of the Connection String spec mentions that Authorization is optional because today we only support "ikey" based auth.

lzchen: Sounds good.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,73 @@
...
+    env_ikey = os.getenv(ENV_INSTRUMENTATION_KEY)
+
+    options.instrumentation_key = code_cs.get(INSTRUMENTATION_KEY) \
+        or code_ikey \
+        or env_cs.get(INSTRUMENTATION_KEY) \
+        or env_ikey
+    endpoint = code_cs.get(INGESTION_ENDPOINT) \
+        or env_cs.get(INGESTION_ENDPOINT) \
+        or DEFAULT_BREEZE_ENDPOINT
+    options.endpoint = endpoint + '/v2/track'
```
reyang: If this endpoint encoded?
Do we do canonicalization?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,73 @@
...
+LOCATION = 'Location'
+
+
+def process_options(options):
+    code_cs = parse_connection_string(options.connection_string)
+    code_ikey = options.instrumentation_key
+    env_cs = parse_connection_string(os.getenv(ENV_CONNECTION_STRING))
+    env_ikey = os.getenv(ENV_INSTRUMENTATION_KEY)
+
+    options.instrumentation_key = code_cs.get(INSTRUMENTATION_KEY) \
```
reyang: The logic is confusing, it'll be helpful to document the decision flow here.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,73 @@
 
 from opencensus.ext.azure.common.protocol import BaseObject
 
+AUTHORIZATION = 'Authorization'
+DEFAULT_BREEZE_ENDPOINT = 'https://dc.services.visualstudio.com'
```
reyang: We should avoid codename here.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,73 @@
...
+ENDPOINT_SUFFIX = 'EndpointSuffix'
+ENV_CONNECTION_STRING = 'APPLICATIONINSIGHTS_CONNECTION_STRING'
+ENV_INSTRUMENTATION_KEY = 'APPINSIGHTS_INSTRUMENTATIONKEY'
```
reyang: If this is going to be used once, why do we want to define it here?
Putting the literal string in the code improves readability as one wouldn't have to jump to different places to read the code.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: Can you explain more about how the endpoint would be encoded?

reyang: For example, do you expect the endpoint to be something like https://zh.wikipedia.org/wiki/%E6%A3%89%E5%B0%BE%E5%85%94%E5%B1%9E ?

lzchen:
Waiting on [#789]

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: I do not believe that an encoded endpoint would be intentionally passed in. According to Mothra, none of the current endpoint configurations handle strings in any special way, so I believe it's safe to leave the url string as is.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: To clarify, is this because `DEFAULT_BREEZE_ENDPOINT` is a module level variable?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -17,14 +17,79 @@
...
+        raise ValueError('Invalid connection string')
+    # Validate authorization
+    auth = result.get('Authorization')
+    if auth is not None and auth.lower() != 'ikey':
+        raise ValueError('Invalid authorization mechanism')
+    # Construct the ingestion endpoint if not passed in explicitly
+    if result.get(INGESTION_ENDPOINT) is None:
+        endpoint_suffix = ''
+        location_prefix = ''
+        suffix = result.get('EndpointSuffix')
```
markwolff: I think the keys are supposed to be case-insensitive. Probably not a blocker since I don't see this actually being tested in other SDKs, but it could lead to some confusing errors for users that manually modify their conn strs :thinking:

contrib/opencensus-ext-azure/README.rst
```diff
@@ -40,9 +41,9 @@ This example shows how to send a warning level log to Azure Monitor.
 You can enrich the logs with trace IDs and span IDs by using the `logging integration <../opencensus-ext-logging>`_.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
-* Install the `logging integration package <../opencensus-ext-logging>`_ using ``pip install opencensus-ext-logging``.
-* Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
-* You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
+* Place your instrumentation key in a connection string and into ``APPLICATIONINSIGHTS_CONNECTION_STRING`` environment variable.
```
reyang: Need a link to the definition of connection string.

contrib/opencensus-ext-azure/README.rst
```diff
@@ -40,9 +41,9 @@ This example shows how to send a warning level log to Azure Monitor.
...
+* You can also put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
```
reyang: Do we want to keep this in the doc? I guess we want to remove it from the doc, and leave the logic in the codebase until we decided to bump major version.

contrib/opencensus-ext-azure/README.rst
lzchen: Good point. I included the connection string in the example code snippet, which is the bare minimum that a user needs to get started (so it is simple). The more formal definition will be included in the TSG for CSS engineers. This will help them troubleshoot any irregularities found with a user's connection string.

contrib/opencensus-ext-azure/README.rst
lzchen: As discussed, we should probably remove the method of `instrumentation_key` from the docs, as we do not want to keep supporting legacy methods. As well, having multiple ways to instrument is confusing to the user.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: Good point. It'll be safer to be case-insensitive.

PR #771, opened 2019-08-22
Fixed requests contrib to raise original exceptions

victoraugustolls:
Updates requests contrib to raise original exceptions, and updated tests.

Thanks @isra17 for noticing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added `http code` to `grpc code` status code mapping on `utils`
   ([#746](https://github.com/census-instrumentation/opencensus-python/pull/746))
 - Updated `requests` module
+  ([#771](https://github.com/census-instrumentation/opencensus-python/pull/771))
 
 ## 0.7.2
 Released 2019-08-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Changelog
 
 ## Unreleased
-- Updated `django`, `flask`, `httplib`, `requests` and `pyramid` modules.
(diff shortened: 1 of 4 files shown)


victoraugustolls:
With this and #755 I'm hopping we can release a new version?
@reyang @c24t

CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 # Changelog
 
 ## Unreleased
-- Updated `django`, `flask`, `httplib`, `requests` and `pyramid` modules.
+- Updated `django`, `flask`, `httplib`, `requests` and `pyramid` modules
   ([#755](https://github.com/census-instrumentation/opencensus-python/pull/755))
 - Added `http code` to `grpc code` status code mapping on `utils`
   ([#746](https://github.com/census-instrumentation/opencensus-python/pull/746))
+- Updated `requests` module
+  ([#771](https://github.com/census-instrumentation/opencensus-python/pull/771))
```
lzchen: Do we need to update the CHANGELOG for the core sdk?

victoraugustolls: Is just for tracking for the next release, so we don’t forget to update some module. I remember someone suggested this recently, but I can’t remember in which PR. It does not need to go into the release changelog (in my opinion), but this way we know what needs to be updated !

lzchen: I see. As long as we are consistent, I feel like it should be fine. So far it seems like we only update the changelog if there are changes in the actual package but your argument for it being a "release reminder" is valid :).

reyang:
@songy23 Sure, we'll take care of it.

PR #789, opened 2019-09-18
Validate instrumentation key in Azure Exporters

lzchen:
Azure exporters will now throw an exception upon passing in an invalid instrumentation key.

The definition of a VALID instrumentation key is as follows:
1. Not none
2. Not empty
3. Valid UUID according to [this](https://tools.ietf.org/html/rfc4122)
      - Every character is a hex character [0-9a-f]
      - 32 characters are separated into 5 sections via 4 dashes
      - First section has 8 characters
      - Second section has 4 characters
      - Third section has 4 characters, the first one in [1-5]
      - Fourth section has 4 characters, the first one in [89ab]
      - Fifth section has 12 characters

### DIFF
--- a/contrib/opencensus-ext-azure/README.rst
+++ b/contrib/opencensus-ext-azure/README.rst
@@ -24,6 +24,7 @@ The **Azure Monitor Log Handler** allows you to export Python logs to `Azure Mon
 This example shows how to send a warning level log to Azure Monitor.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
+* The instrumentation key must follow the standard UUID format as defined `here <https://tools.ietf.org/html/rfc4122>` _.
 * Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
 * You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
 
@@ -40,6 +41,7 @@ This example shows how to send a warning level log to Azure Monitor.
 You can enrich the logs with trace IDs and span IDs by using the `logging integration <../opencensus-ext-logging>`_.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
+* The instrumentation key must follow the standard UUID format as defined `here <https://tools.ietf.org/html/rfc4122>` _.
 * Install the `logging integration package <../opencensus-ext-logging>`_ using ``pip install opencensus-ext-logging``.
 * Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
 * You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
@@ -75,6 +77,7 @@ Metrics
 The **Azure Monitor Metrics Exporter** allows you to export metrics to `Azure Monitor`_.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
+* The instrumentation key must follow the standard UUID format as defined `here <https://tools.ietf.org/html/rfc4122>` _.
 * Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
 * You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
 
@@ -167,6 +170,7 @@ The **Azure Monitor Trace Exporter** allows you to export `OpenCensus`_ traces t
 This example shows how to send a span "hello" to Azure Monitor.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
+* The instrumentation key must follow the standard UUID format as defined `here <https://tools.ietf.org/html/rfc4122>` _.
 * Put the instrumentation key in ``APPINSIGHTS_INSTRUMENTATIONKEY`` environment variable.
 * You can also specify the instrumentation key explicitly in the code, which will take priority over a set environment variable.
 
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -39,3 +39,12 @@ class Options(BaseObject):
         storage_retention_period=7*24*60*60,
         timeout=10.0,  # networking timeout in seconds
     )
+
+
+def validate_instrumentation_key(instrumentation_key):
+    if not instrumentation_key:
+        raise ValueError("Instrumentation key cannot be none or empty.")
+    if len(instrumentation_key) > 26:
+        raise ValueError("Instrumentation key exceeds character limit.")
+    
+
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import re
 import sys
 
 from opencensus.ext.azure.common.protocol import BaseObject
@@ -41,10 +42,21 @@ class Options(BaseObject):
     )
 
 
-def validate_instrumentation_key(instrumentation_key):
+def validate_key(instrumentation_key):
     if not instrumentation_key:
         raise ValueError("Instrumentation key cannot be none or empty.")
-    if len(instrumentation_key) > 26:
-        raise ValueError("Instrumentation key exceeds character limit.")
+    # Validate UUID format
+    # Specs taken from https://tools.ietf.org/html/rfc4122
+    pattern = re.compile('[0-9a-f]{8}-' \
+                         '[0-9a-f]{4}-' \
+                         '[1-5][0-9a-f]{3}-' \
+                         '[89ab][0-9a-f]{3}-' \
+                         '[0-9a-f]{12}')
+    # re.fullmatch not available for python2
+    # We use re.match, we matches from the beginning, and then do a length
+    # check to ignore everything afterward.
+    match = pattern.match(instrumentation_key)
+    if len(instrumentation_key) != 36 or not match:
+        raise ValueError("Invalid instrumentation key.")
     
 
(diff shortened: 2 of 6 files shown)


contrib/opencensus-ext-azure/README.rst
```diff
@@ -24,6 +24,7 @@ The **Azure Monitor Log Handler** allows you to export Python logs to `Azure Mon
 This example shows how to send a warning level log to Azure Monitor.
 
 * Create an Azure Monitor resource and get the instrumentation key, more information can be found `here <https://docs.microsoft.com/azure/azure-monitor/app/create-new-resource>`_.
+* The instrumentation key must follow the standard UUID format as defined `here <https://tools.ietf.org/html/rfc4122>`_.
```
reyang: We don't need this.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -39,3 +40,21 @@ class Options(BaseObject):
...
         timeout=10.0,  # networking timeout in seconds
     )
+
+
+def validate_key(instrumentation_key):
+    if not instrumentation_key:
+        raise ValueError("Instrumentation key cannot be none or empty.")
+    # Validate UUID format
+    # Specs taken from https://tools.ietf.org/html/rfc4122
+    pattern = re.compile('[0-9a-f]{8}-'
```
reyang: This precompiled regular expression should be created once, not inside the function.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -39,3 +40,21 @@ class Options(BaseObject):
...
+                         '[0-9a-f]{4}-'
+                         '[1-5][0-9a-f]{3}-'
+                         '[89ab][0-9a-f]{3}-'
+                         '[0-9a-f]{12}')
+    # re.fullmatch not available for python2
+    # We use re.match, we matches from the beginning, and then do a length
+    # check to ignore everything afterward.
+    match = pattern.match(instrumentation_key)
```
reyang: This is weird, we should use something `^foobar$`.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -20,7 +20,7 @@
 from opencensus.common.schedule import Queue
 from opencensus.common.schedule import QueueExitEvent
 from opencensus.common.schedule import QueueEvent
-from opencensus.ext.azure.common import Options
+from opencensus.ext.azure import common
```
reyang: This needs to be cleaned up.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -39,3 +40,21 @@ class Options(BaseObject):
         storage_retention_period=7*24*60*60,
         timeout=10.0,  # networking timeout in seconds
     )
+
+
+def validate_key(instrumentation_key):
```
reyang: Why not put this inside `utils`?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: Sorry, can you explain a bit more about what you mean?

contrib/opencensus-ext-azure/README.rst
```diff
@@ -37,11 +37,11 @@ This example shows how to send a warning level log to Azure Monitor.
     logger.addHandler(AzureLogHandler(connection_string='InstrumentationKey=<your-instrumentation_key-here>'))
     logger.warning('Hello, World!')
 
-* Alternatively, you can specify your `connection string` in an environment variable ``APPLICATIONINSIGHTS_CONNECTION_STRING``.
```
hectorhdzg: Is not possible to use the environment variable anymore?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,23 @@ def timestamp_to_iso_str(timestamp):
...
+                     '[1-5][0-9a-f]{3}-'
+                     '[89ab][0-9a-f]{3}-'
+                     '[0-9a-f]{12}')
+
+
+def validate_key(instrumentation_key):
+    if not instrumentation_key:
+        raise ValueError("Instrumentation key cannot be none or empty.")
+    # re.fullmatch not available for python2
+    # We use re.match, we matches from the beginning, and then do a length
```
markwolff:
```suggestion
    # Use re.match instead: match from the beginning, and then do a length
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,23 @@ def timestamp_to_iso_str(timestamp):
...
+# Validate UUID format
+# Specs taken from https://tools.ietf.org/html/rfc4122
+pattern = re.compile('[0-9a-f]{8}-'
+                     '[0-9a-f]{4}-'
+                     '[1-5][0-9a-f]{3}-'
+                     '[89ab][0-9a-f]{3}-'
+                     '[0-9a-f]{12}')
+
+
+def validate_key(instrumentation_key):
```
markwolff: I assume "prefixed" ikeys, e.g. `AIF-1234abcd-5678-4efa-8abc-1234567890ab` would be rejected by this. Is this strict checking something we want to expose a way to turn off?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
hectorhdzg: ^ Start of string, or start of line in multi-line pattern
$ End of string, or end of line in multi-line pattern

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,23 @@ def timestamp_to_iso_str(timestamp):
...
+    if not instrumentation_key:
+        raise ValueError("Instrumentation key cannot be none or empty.")
+    # re.fullmatch not available for python2
+    # We use re.match, we matches from the beginning, and then do a length
+    # check to ignore everything afterward.
+    match = pattern.match(instrumentation_key)
+    if len(instrumentation_key) != 36 or not match:
+        raise ValueError("Invalid instrumentation key.")
```
markwolff: Could these be separated into 2 different conditionals so that some error information could be provided, invalid length vs invalid regex?

contrib/opencensus-ext-azure/README.rst
lzchen: It is, this was just a duplicated line in the docs.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: I've add the `^` and `$` to the regex so length does not need to be checked.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
lzchen: Changed the regex to match full string so comment is not needed.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
lzchen: We've decided that checking for GUID is sufficient enough for validation. As for "prefixed" ikeys, the specs that were given for prefixed ikeys do not seem to apply for AI customers, so we can safely assume that all ikeys should be GUIDs.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
lzchen: I've incorporated the length check into the regex. Normally, the customer would just re-check and copy/paste the entire ikey again if they run into this error.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -115,8 +115,7 @@ class AzureLogHandler(TransportMixin, BaseLogHandler):
 
     def __init__(self, **options):
         self.options = Options(**options)
-        if not self.options.instrumentation_key:
```
xiao-lix: Is it necessary to check null for ikey twice since it'll be checked in validate_key first?

lzchen: It is not necessary. The `if not self.options.instrumentation_key:` is a deleted line :)

lzchen:
@reyang 
Can you take another look? I've made the changes that were suggested.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,20 @@ def timestamp_to_iso_str(timestamp):
 
 def url_to_dependency_name(url):
     return urlparse(url).netloc
+
+
+# Validate UUID format
```
reyang: Why do we want UUID format link here?
We should have a link to the definition of "what's a valid instrumentation key for Azure Monitor".

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,20 @@ def timestamp_to_iso_str(timestamp):
...
+# Specs taken from https://tools.ietf.org/html/rfc4122
+pattern = re.compile('^[0-9a-f]{8}-'
```
reyang: `pattern` is a very bad name. We should avoid it.

reyang: Especially for global objects that are visible to external packages.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -65,3 +66,20 @@ def timestamp_to_iso_str(timestamp):
...
+                     '[0-9a-f]{4}-'
+                     '[1-5][0-9a-f]{3}-'
+                     '[89ab][0-9a-f]{3}-'
+                     '[0-9a-f]{12}$')
+
+
+def validate_key(instrumentation_key):
```
reyang:
```suggestion
def validate_instrumentation_key(instrumentation_key):
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
lzchen: I don't believe there is an official definition of "what's a valid instrumentation key for Azure Monitor". The links that Mothra had referred to (with the prefix and suffix) do not seem to apply to the ikeys that are sent to Azure Monitor, so we decided that the validation would be to check for UUID. With that being said, I will make note of this in the comments.

Issue #279, opened 2018-08-25
trace/Span: missing setter for Status or current method is cumbersome


odeke-em:
Currently a span has an accessible field `status`
https://github.com/census-instrumentation/opencensus-python/blob/2bc3745ef61719db90f091a62897c802b5762331/opencensus/trace/span.py#L143

However, to set in code, a user has to import `status` as well and create a `status.Status` object.
```python
from opencensus.trace import status

span.status = Status(1, "Exception occured")
```

I believe can be simplified to
```python
span.set_status(1, "Exception occured")
```
as all a user needs to set here is the code and the message

MetalDent:
Hi, is anyone working on this issue?

lzchen:
@MetalDent 
Currently no. You are more than welcome to submit a PR for it!

PR #798, opened 2019-10-05
#279 Add new method to set span status.

maeserichar:
Just an easy change to avoid needing to import status in order to set the status of a span.

### DIFF
--- a/opencensus/trace/span.py
+++ b/opencensus/trace/span.py
@@ -362,6 +362,20 @@ class Span(base_span.BaseSpan):
             raise TypeError("Type Error: received {}, but requires Status.".
                             format(type(status).__name__))
 
+    def set_status_fields(self, code, message=None, details=None):
+        """Set span status.
+ 
+        :type code: int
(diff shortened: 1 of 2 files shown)


victoraugustolls:
This was partially discussed [here](https://github.com/census-instrumentation/opencensus-python/pull/738#discussion_r309378259)!

@reyang What do you think?

lzchen:
With this change, is there a need for the `Status` class?
What do you think @victoraugustolls  @reyang ?

victoraugustolls:
This seems to me more like a helper method, and I wouldn't remove `Status` class, as not only it is a breaking change but is referenced in `OT` spec as mandatory, and it would make the job a lot difficult in the migration in case it was removed.

In my opinion, I would rather have utils method and not add it inside the `Span` class, but it is a personal view, would wait @reyang for comments.

reyang:
I agree with @victoraugustolls.

maeserichar:
Great! So something in the line of `setSpanStatus(span, code, message, details)` in `trace/utils.py' would be more aligned with that?

PR #804, opened 2019-10-15
MANIFEST.in: Add LICENSE, docs, contrib and tests

jayvdb:


reyang:
Do we need source distribution at all or we should just remove this file?

jayvdb:
As a packager, I find it very helpful for repos with multiple packages to have good sdists.  The core requirement is to be able to get a versioned tarball that includes tests.  At the moment, the only option here is to fetch the tarball for the repo release on GitHub.

The google cloud python repo is a great example to follow.

I have started packaging these.
As you can see at https://build.opensuse.org/package/show/home:jayvdb:py-new/python-opencensus-context , not having a package sdist with tests means the package needs to fetch the very big tarball, and logic of the .spec is a bit odd.

https://build.opensuse.org/package/show/home:jayvdb:py-new/python-opencensus is also more complicated than it needs to be, but that is also because the extensions are needed for the main test suite.

PR #811, opened 2019-11-06
Fix Django DB wrapper initialization

### LINKED ISSUE #810: Django DB Instrumentation problem
I'm facing problems to tracing database queries. In the `middleware.py` file, `execute_wrappers` are updating to include `_trace_db_call` function [here](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py#L168).

But when Django access  `execute_wrappers`  [here](https://github.com/django/django/blob/af5ec222ccd24e81f9fec6c34836a4e503e7ccf7/django/db/backends/utils.py#L75), it seems to be empty. Looking closely, I've found different object instances when the `execute_wrappers` are updated at first (`<django.db.backends.sqlite3.base.DatabaseWrapper object at 0x7fa410750240>`), and when it is accessed after (<django.db.backends.sqlite3.base.DatabaseWrapper object at 0x7fa4103a24a8>). Seems that Django instantiate multiple DatabaseWrapper objects, causing the problem.

If I change the initialization of `execute_wrappers` to `process_request` it works, but only log one query:

```
    def process_request(self, request):
        """Called on each request, before Django decides which view to execute.

        :type request: :class:`~django.http.request.HttpRequest`
        :param request: Django http request.
        """
        if django.VERSION >= (2,) and not _trace_db_call in connection.execute_wrappers:
            connection.execute_wrappers.append(_trace_db_call)
```

My environment:

```
Package                  Version  
------------------------ ---------
cachetools               3.1.1    
certifi                  2019.9.11
chardet                  3.0.4    
contextvars              2.4      
Django                   2.0      
google-api-core          1.14.3   
google-auth              1.7.0    
googleapis-common-protos 1.6.0    
idna                     2.8      
immutables               0.11     
mysql                    0.0.2    
mysql-connector          2.2.9    
mysqlclient              1.4.4    
opencensus               0.7.5    
opencensus-context       0.1.1    
opencensus-ext-dbapi     0.1.2    
opencensus-ext-django    0.7.2    
opencensus-ext-mysql     0.1.2    
opencensus-ext-zipkin    0.2.2    
pip                      19.3.1   
pkg-resources            0.0.0    
protobuf                 3.10.0   
pyasn1                   0.4.7    
pyasn1-modules           0.2.7    
PyMySQL                  0.9.3    
pytz                     2019.3   
requests                 2.22.0   
rsa                      4.0      
setuptools               39.0.1   
six                      1.12.0   
sqlparse                 0.3.0    
urllib3                  1.25.6

```

I'm trying to use it in the example included on plugin (https://github.com/census-instrumentation/opencensus-python/tree/master/contrib/opencensus-ext-django).

I just change settings to use SQLITE:

```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': os.path.join(BASE_DIR, '../db.sqlite3'),
    }
}
```

I'm doing something wrong?

tpyo:
Fixes #810

### DIFF
--- a/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
+++ b/contrib/opencensus-ext-django/opencensus/ext/django/middleware.py
@@ -164,8 +164,11 @@ class OpencensusMiddleware(MiddlewareMixin):
 
         self.blacklist_hostnames = settings.get(BLACKLIST_HOSTNAMES, None)
 
+    def __call__(self, request):
         if django.VERSION >= (2,):  # pragma: NO COVER
-            connection.execute_wrappers.append(_trace_db_call)
+            with connection.execute_wrapper(_trace_db_call):
+                return super(OpencensusMiddleware, self).__call__(request)
+        return super(OpencensusMiddleware, self).__call__(request)
 
     def process_request(self, request):
         """Called on each request, before Django decides which view to execute.


tpyo:
@jordanorc Would you mind testing this to see if it resolves your issue?

jordanorc:
@tpyo, Yes!!! It's working perfectly now! I also tested with Daphne and everything seems to be ok. Thank you very much for your help!!! 

I'm now facing problems with requests integration. Do you know if someone can help?

tpyo:
@jordanorc Thanks for testing.  Maybe open an issue if you're having trouble with the requests integration?

jordanorc:
Thanks @tpyo! I already do that here: https://github.com/census-instrumentation/opencensus-python/issues/812. I also found a solution to solve the "problem".But I don't know if it is a problem in fact.

AyWa:
I have some issue with django and db trace. So I hope this issue resolve it. 
Any work that need to be done on it ?

tpyo:
@AyWa Tests are failing but this does address the issue.  Haven't had a chance to figure out how to get it to work with the context manager Django uses.  Help is welcome!

PR #814, opened 2019-11-11
Add support for exporting Custom Events to Azure App Insights.

ivaramme:
This PR enables sending Custom Events to Application Insights (Azure) from an application by creating a new logging exporter with the expected data structure used by App Insights for this feature. 

Custom Events in App Insights are a special type of structured, searchable data points reflecting an event or action. An event must contain an event name and optionally a set of properties (in the form of dictionaries) that are attached to that entry.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
@@ -1,0 +1,24 @@
+from opencensus.ext.azure.common import utils
+from opencensus.ext.azure.common.protocol import Data, Envelope, Event
+from opencensus.ext.azure.log_exporter import AzureLogHandler
+
+
+class AzureEventHandler(AzureLogHandler):
+    def __init__(self, **options):
+        super(AzureEventHandler, self).__init__(**options)
+
+    def log_record_to_envelope(self, record):
+        envelope = Envelope(
+            iKey=self.options.instrumentation_key,
+            tags=dict(utils.azure_monitor_context),
+            time=utils.timestamp_to_iso_str(record.created),
+        )
+
+        envelope.name = "Microsoft.ApplicationInsights.Event"
+        data = Event(
+            name=record.msg,
+            properties=record.args,
+            measurements=None,
+        )
+        envelope.data = Data(baseData=data, baseType='EventData')
+        return envelope
--- a/contrib/opencensus-ext-azure/tests/test_azure_event_exporter.py
+++ b/contrib/opencensus-ext-azure/tests/test_azure_event_exporter.py
@@ -1,0 +1,69 @@
+import logging
+import os
+import shutil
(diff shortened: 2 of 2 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+        super(AzureEventHandler, self).__init__(**options)
+
+    def log_record_to_envelope(self, record):
+        envelope = Envelope(
+            iKey=self.options.instrumentation_key,
+            tags=dict(utils.azure_monitor_context),
+            time=utils.timestamp_to_iso_str(record.created),
+        )
+
+        envelope.name = "Microsoft.ApplicationInsights.Event"
```
lzchen: I thought we decided to use Logs as the actual telemetry type? This would be `Message` data.

contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
```diff
@@ -0,0 +1,24 @@
+from opencensus.ext.azure.common import utils
+from opencensus.ext.azure.common.protocol import Data, Envelope, Event
+from opencensus.ext.azure.log_exporter import AzureLogHandler
+
+
+class AzureEventHandler(AzureLogHandler):
```
lzchen: I don't think we need another handler type if we are going to be using `Message` telemetry type. We can leverage the already existing `AzureLogHandler` and place the new "properties" logic in there.

contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, **options):
+        super(AzureEventHandler, self).__init__(**options)
+
+    def log_record_to_envelope(self, record):
```
lzchen: I think adding an example of how this works in `examples.logs` will help. As well, updating the example in the README and adding an excerpt of this feature will be good.

contrib/opencensus-ext-azure/opencensus/ext/azure/events_exporter/__init__.py
ivaramme: But wouldn't that log the data as "searchable text" in the logs table (https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/MessageData.bond#L6)? 

The event definition is different (and this link might be outdated) and more resonant to what we are trying to accomplish here (custom events): https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/EventData.bond#L5

Maybe I'm missing something, please let me know.

lzchen:
Closed due to [#822 ]

Issue #816, opened 2019-11-13
FileExistsError: [Errno 17] File exists: '/root/.opencensus/.azure/gunicorn' when I try to run Docker


RenanAlonkin:
Hello there,

I'm currently having the following problem: I have an API, and I want to use Opencensus to export my exceptions to Azure's Application Insights, but, when I try to run it inside of a Docker Container this problem appears:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
    worker.init_process()
...

File "/app/opensensus_mid.py", line 39, in <module>
    connection_string=connection_string))
  File "/usr/local/lib/python3.7/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 127, in __init__
    retention_period=self.options.storage_retention_period,
  File "/usr/local/lib/python3.7/site-packages/opencensus/ext/azure/common/storage.py", line 90, in __init__
    self._maintenance_routine(silent=False)
  File "/usr/local/lib/python3.7/site-packages/opencensus/ext/azure/common/storage.py", line 112, in _maintenance_routine
    os.makedirs(self.path)
  File "/usr/local/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/root/.opencensus/.azure/gunicorn'
```

**Describe your environment.** 

- Python `3.8.0 (default, Nov  6 2019, 16:00:02)`
- Docker: 
```
Client:
 Version:      17.03.1-ce-rc1
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   3476dbf
 Built:        Wed Mar 15 20:33:22 2017
 OS/Arch:      windows/amd64

Server:
 Version:      17.03.1-ce-rc1
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   3476dbf
 Built:        Wed Mar 15 20:28:18 2017
 OS/Arch:      linux/amd64
```

- Opencensus `opencensus==0.7.5
opencensus-ext-azure==1.0.0`

- FastAPI: `fastapi==0.42.0
 uvicorn==0.10.8
gunicorn==20.0.0`

**Steps to reproduce.**
I've created an "Opencensus middleware" for FastAPI, the code can be found at [this link](https://pastebin.com/MyakznYk). All I do is call it inside FastAPI middleware. 

```
@app.middleware("http")
[...]

PR #817, opened 2019-11-14
Allowing multiple worker applications (issue #816).

RenanAlonkin:
Add `exist_ok=True` to `opencensus-ext-azure.ext.azure.common.storage._maintenance_routine` to avoid raising `FileExistsError` if the target directory already exists, which is the default behavior (exist_ok=False). This behavior (`exist_ok=False`) do not allow multiple worker applications (such as uvicorn) to perform `run`.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
@@ -114,7 +114,7 @@ class LocalFileStorage(object):
     def _maintenance_routine(self, silent=False):
         try:
             if not os.path.isdir(self.path):
-                Path(self.path).mkdir(exist_ok=True)
+                os.makedirs(self.path, exist_ok=True)
         except Exception:
             if not silent:
                 raise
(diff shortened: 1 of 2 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
```diff
@@ -109,7 +114,7 @@ def __exit__(self, type, value, traceback):
     def _maintenance_routine(self, silent=False):
         try:
             if not os.path.isdir(self.path):
-                os.makedirs(self.path)
+                os.makedirs(self.path, exist_ok=True)
```
lzchen: Can you put a comment explaining the reasoning for this? (raise condition when `os.makedirs` is called after the `if` check passes but the file was already created by another worker).

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
```diff
@@ -5,6 +5,11 @@
 
 from opencensus.common.schedule import PeriodicTask
 
+try:
+  from pathlib import Path
```
lzchen: Why do we need to import `Path`? I don't see it being used in the logic?

lzchen:
Address [#816]

lzchen:
Closing due to workaround.

PR #820, opened 2019-11-20
WIP: Add track_metric method to AzureLogHandler

iMichka:
The legacy azure logging library (https://github.com/microsoft/ApplicationInsights-Python) allowed to add metrics to azure insights in a few lines:

```
from applicationinsights import TelemetryClient
tc = TelemetryClient('<YOUR INSTRUMENTATION KEY GOES HERE>')
tc.track_metric('My Metric', 42)
tc.flush()
```

This change in opencensus would allow to do the equivalent, while benefiting from the threaded queue mechanism of opencensus:

```
handler = AzureLogHandler(connection_string="InstrumentationKey=<>")
handler.track_metric('My Metric', 42)
```

This would allow an easier migration for users coming from the legacy applicationinsights library.

Using the opencensus metrics can not be used for this situation as the metrics exporter will send metrics continuously over and over, which is not the expected behaviour (at least not in my case).

This change is a WIP as it just a quick hack to open the discussion about the strategy we may want to use to handle this use-case. I do not expect this to be merged as-is.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -24,6 +24,7 @@ from opencensus.ext.azure.common.protocol import (
     Envelope,
     ExceptionData,
     Message,
+    MetricData,
 )
 from opencensus.ext.azure.common.storage import LocalFileStorage
 from opencensus.ext.azure.common.transport import TransportMixin
@@ -147,7 +148,34 @@ class AzureLogHandler(TransportMixin, BaseLogHandler):
             if event:
                 event.set()
 
+    def track_metric(self, name, value):
+        self._queue.put((name, value), block=False)
+
     def log_record_to_envelope(self, record):
+        if isinstance(record, tuple):
+            envelope = Envelope(
+                iKey=self.options.instrumentation_key,
+                tags=dict(utils.azure_monitor_context),
+                time=utils.to_iso_str(),
+            )
+            envelope.name = 'Microsoft.ApplicationInsights.Metric'
+            name = record[0]
+            value = record[1]
+            data = MetricData(
+                metrics=[
+                    {'kind': None,
+                     'name': name,
+                     'ns': name,
+                     'max': value,
+                     'count': 1,
+                     'value': value,
+                     'min': value,
+                     'stdDev': value}],
+                properties={}
+            )
+            envelope.data = Data(baseData=data, baseType='MetricData')
+            return envelope
+        
         envelope = Envelope(
             iKey=self.options.instrumentation_key,
             tags=dict(utils.azure_monitor_context),


iMichka:
I am using a similar approach in production now:

````
from opencensus.ext.azure.common import utils
from opencensus.ext.azure.common.protocol import Data, Envelope, MetricData

class MyAzureLogHandler(AzureLogHandler):

    def track_metric(self, name, value):
        self._queue.put((name, value), block=False)

    def log_record_to_envelope(self, record):
        if isinstance(record, tuple):
            envelope = Envelope(
                iKey=self.options.instrumentation_key,
                tags=dict(utils.azure_monitor_context),
                time=utils.to_iso_str(),
            )
            envelope.name = 'Microsoft.ApplicationInsights.Metric'
            name = record[0]
            value = record[1]
            data = MetricData(
                metrics=[
                    {'kind': None,
                     'name': name,
                     'ns': name,
                     'max': value,
                     'count': 1,
                     'value': value,
                     'min': value,
                     'stdDev': 0}],
                properties={}
            )
            envelope.data = Data(baseData=data, baseType='MetricData')
            return envelope

        return super().log_record_to_envelope(record)
````

This allows me to override the AzureLogHandler class provided by opencensus on the fly, without needing to maintain a fork of opencensus.

This "hack" is quite dirty (the whole logger concept works with logging.Records, and I am inserting tuples in the queue (resulting in an isinstance call)). But this will do until the opentelemetry project has matured and we find a better solution there.

I did not expect this PR to get merged so I'll close this. But somebody may find my code useful as a quick hack to migrate from applicationinsights to opencensus.

**Comment by raphaelauv** [comment]
@iMichka Thank for your workaround.

**Comment by raphaelauv** [comment]
@iMichka 

I'm trying to publish the metric with custom domains https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-custom-overview#dimension-keys

I changed in you solution the 
```python
properties={}
```
by
```python
properties={"my_dimension":"a_value"}
```
 but in azure monitor when I select the metric , I can't filter on the dimension , the button FILTER stay gray and say : 
```
Filter is not supported for the selecte metric(s) ...
```
Screenshot from 2021-02-19 10-32-14

Any idee ? Thank you

iMichka:
I never used the filtering (I wasn't aware it existed), I mostly do the filtering in my code and send separate metrics. But I can understand that it might be useful. Not sure what the issue is.

I do not plan to provide more support for my workaround. I also do not want this to become the official way of doing things. It is more a hack I wrote to move forward while migrating from the old azure applicationinsights package to opencensus, to get rid of a deprecated dependency.

Last time I checked, opencensus was declared dead by Azure and https://github.com/open-telemetry should be used instead, but I did not have time to re-evaluate this and do yet another logging package migration (the third in 2 years). Looks like Microsoft and Azure are moving really fast, and leaving people with half-implemented solutions. I'll come back to logging issues as part of my work end of 2021 (depending on our internal roadmap and technical debt that needs to be handled), but I have no bandwidth allocated right now to work on logging issues, sorry.

raphaelauv:
Thank @iMichka for your quick answer :+1: 

You're right about "with half-implemented solutions" from azure, it's really frustrating :disappointed: 

I will try to get support from https://github.com/Azure/azure-sdk-for-python/issues/16447

lzchen:
@iMichka 
> opencensus was declared dead by Azure and https://github.com/open-telemetry 

Opencensus is not declared dead by Azure. It is our recommended Python solution to send data to appinsights.

@raphaelauv 
I have addressed your question [here](https://github.com/Azure/azure-sdk-for-python/issues/16447).

iMichka:
> Opencensus is not declared dead by Azure.

I might have misunderstood this blog post then:
https://cloudblogs.microsoft.com/opensource/2019/05/23/announcing-opentelemetry-cncf-merged-opencensus-opentracing/

Back then it looked like everything was merging into opentelemetry. This project (https://github.com/microsoft/opentelemetry-azure-monitor-python) looked like the new official replacement for OpenCensus. The project was then merged into the official azure SDK: https://github.com/Azure/azure-sdk-for-python

So I thought open-telemetry was the officially supported solution for logging and metrics, at least from a Azure point of view. But I probably misunderstood the subtleties, sorry for that.

lzchen:
Opentelemetry is currently in beta. It will be the official supported solution and recommended APM solution once it becomes GA. But until then, Opencensus is still Microsoft's recommended solution. Also, Opentelemetry does not support Python2.7, so for customers that cannot upgrade their Python versions, Opencensus is still their go-to solution.

PR #822, opened 2019-11-21
Add optional custom properties to logging messages.

ivaramme:
The idea behind this PR is to allow users to track additional information to logging actions by simply adding to the default properties, any additional metadata passed by the user in the form of a dictionary. 

The usage of this would look like this:

```
import logging

# TODO: set APPLICATIONINSIGHTS_CONNECTION_STRING as environment variable.
logger.warning('action', {'key-1': 'value-1', 'key-2': 'value2'})
```

Once this data is uploaded to App Insights, you can search in the `traces` table by keys included in the `customDimensions` column like this:

```
traces
| where customDimensions.key-1 == "value-1"
| limit 50
| order by timestamp desc 
```

This is a good option instead of using Custom Events which is not currently supported in Open Census.

### DIFF
--- a/contrib/opencensus-ext-azure/README.rst
+++ b/contrib/opencensus-ext-azure/README.rst
@@ -73,6 +73,20 @@ You can enrich the logs with trace IDs and span IDs by using the `logging integr
         logger.warning('In the span')
     logger.warning('After the span')
 
+You can also add custom properties to your log messages in the form of key-values.
+
+WARNING: For this feature to work, you need to pass a dictionary as the argument. If you pass arguments of any other type, the logger will ignore them. The solution is to convert these arguments into a dictionary.
+
+.. code:: python
+
+    import logging
+
+    from opencensus.ext.azure.log_exporter import AzureLogHandler
+
+    logger = logging.getLogger(__name__)
+    logger.addHandler(AzureLogHandler(connection_string='InstrumentationKey=<your-instrumentation_key-here>'))
+    logger.warning('action', {'key-1': 'value-1', 'key-2': 'value2'})
+
 Metrics
 ~~~~~~~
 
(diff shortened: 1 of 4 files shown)


contrib/opencensus-ext-azure/README.rst
```diff
@@ -73,6 +73,18 @@ You can enrich the logs with trace IDs and span IDs by using the `logging integr
         logger.warning('In the span')
     logger.warning('After the span')
 
+You can also add custom properties to your log messages in the form of key-values.
+
```
lzchen: You should add a warning noting that string formatting will be lost when using custom properties. There doesn't seem to be any way to avoid this.

contrib/opencensus-ext-azure/examples/logs/properties.py
```diff
@@ -0,0 +1,10 @@
+import logging
```
lzchen: Remember to add the OpenCensus copyright boilerplate.

lzchen:
Looks good! Just a couple of comments!

contrib/opencensus-ext-azure/README.rst
ivaramme: I tested this and if the logging instruction includes a dictionary and string formatting, like this `logger.warning("test_%s", { "SKU" : "SKU-1" })` the message will be logged like this: `test_{'SKU': 'SKU-1'}` and the dictionary would be logged as well inside the `customProperties` column.

lzchen: That's true but the common usage of the args parameter is using a string variable. If user's logging messages are already using string formatting, you are forcing them to change to passing in dictionaries.

ivaramme: Added the warning message.

Issue #826, opened 2019-11-22
Azure Monitor exporter  does not populate HTTP dependency properties correctly


lmolkova:
See the similar issue for OpenTelemtery https://github.com/microsoft/opentelemetry-exporters-python/issues/14

**Steps to reproduce**

* Enable requests integration.
* Make outgoing HTTP call

**What is the expected behavior?**
Dependency telemetry is populated properly with Data, Target and name:

* Data matches URL
* Target matches target url's authority (host:port) 
* name is "METHOD /path"

**What is the actual behavior?**

* Data not set
* Target not set
* name not correct

**Additional context**
https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py#L109..L123

Issue #825, opened 2019-11-22
Azure Monitor exporter always sets  dependency.type to HTTP for client spans


lmolkova:
https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py#L109

HTTP type should only be set in the presence of http.* attribute

PR #827, opened 2019-11-25
Fix HTTP attributes for deps, fix success and response/resultCode for Azure Exporter

lzchen:
Address [#825] and [#826], [#831] and [#832]

Adds "component" span attributes to `requests` and `httplib` integrations to specify "http" type.
Adds logic to popular success and responseCode based off span status or status_code in span attributes.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
@@ -27,6 +27,11 @@ from opencensus.ext.azure.common.storage import LocalFileStorage
 from opencensus.ext.azure.common.transport import TransportMixin
 from opencensus.trace.span import SpanKind
 
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
 logger = logging.getLogger(__name__)
 
 __all__ = ['AzureExporter']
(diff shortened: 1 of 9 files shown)


lzchen:
@lmolkova 
Please review if you have the time. It consists of the same changes as for OpenTelemetry.

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -87,32 +92,45 @@ def span_data_to_envelope(self, sd):
...
+                        # name is METHOD/path
+                        data.name = sd.attributes['http.method'] \
+                            + "/" + parse_url.path
                 if 'http.status_code' in sd.attributes:
-                    data.resultCode = str(sd.attributes['http.status_code'])
+                    status_code = sd.attributes["http.status_code"]
+                    data.resultCode = str(status_code)
+                    data.success = 200 <= status_code < 400
+                elif sd.status.code == 0:
+                    data.success = True
```
lmolkova: what if it's not SERVER or CLIENT span (Unspecified), we create Dependency and we should still set status

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -55,22 +60,22 @@ def span_data_to_envelope(self, sd):
...
+                id='{}'.format(sd.span_id),
                 duration=utils.timestamp_to_duration(
                     sd.start_time,
                     sd.end_time,
                 ),
-                responseCode='0',
-                success=False,
+                responseCode=str(sd.status.code),
+                success=False,  # Modify based off attributes or status
                 properties={},
```
lmolkova: this was not listed as a bug, but I noticed we don't populate properties for HTTP requests. Can we do it?

request.name = "METHOD /route" (if route is not populated, then path)
request.url = url

lzchen:
@lmolkova 
I have made the changes you've suggested!

Issue #834, opened 2019-12-20
Custom Properties/Dimensions Design Discussion


xinyi-joffre:
I'm happy to help contribute the code, but wanted to share design first and discuss to see if it is right approach.

**Is your feature request related to a problem?**

Support for custom properties/dimensions was recently added for logging messages using args.

```python
logger.info("my message", {"customProperty": "myValue"})
```

There were a few concerns with the current approach when applying it to existing code:

1) Current approach with *args breaks existing logging functionality for other registered logging handlers, which are using args for lazy string interpolation. We would need to go back and change existing log calls to remove existing args (interpolate strings first) and change arg type to dictionary, if we wanted to log custom properties.
2) It breaks LoggerAdapter and filter patterns for passing contextual information to all messages. For LoggerAdapter pattern, we cannot modify args at all. For filter pattern, we would currently need to overwrite the args there. 
3) We need to pass the custom dimensions into every call to the logger because of concern number 2.
3) It only supports messages and not metrics/exceptions, so approach is not consistent.

**Describe the solution you'd like.**

Instead of using the *args as a dict, I would suggest using the **kwargs approach.

For the logging library, **kwargs are added to the __dict__ of LoggerRecord. When we process the record, we could look for customDimensions to be present on the record as an attribute. If it is there, then add it to the properties to be sent to Application Insights.

This solution would leave in place existing behavior, since end users can still use kwargs for storing other important attributes as long as it does not conflict with customDimensions (which is very application insights specific term).

As part of this request, it would also be great to have consistency across messages, exceptions, events, etc, instead of only supporting custom properties for messages.

Below are examples of logging use cases with the suggested approach:

**Example 1: Simple Case for Adding Properties Once**

```python
properties = {‘customDimensions’: {‘testKey’: ‘testValue’}}
logger.warning(“Starting retry %s”, retries, properties)
```

**Example 2: Pass Custom Properties as Context with LoggerAdaptor pattern**

```python
# FYI, just quickly sketching out code for explaining, more checking is needed for nulls, etc.
class CustomAdapter(logging.LoggerAdapter):
    def process(self, msg, kwargs):
	kwargs['customDimensions'].update(self.extra[‘customDimensions’])
        return msg, kwargs

logger = logging.getLogger(__name__)
adapter = CustomAdapter(logger, {‘customDimensions’: {‘testKey’: ‘testValue’}})
adapter.warning(‘hello’)
[...]

PR #837, opened 2020-01-10
Add custom_dimensions using kwargs instead of args

xinyi-joffre:
This PR allows users to add customDimensions to their logs for both traces and exceptions. This approach was discussed in #834 before implementing.

As a part of this shift, it will break the changes introduced as part of #822. However, the new implementation allows for a smoother transition for those wanting to keep existing logging functionality with string interpolation and args. In addition, this PR adds support for exceptions which was not covered as part of the previous implementation.

Usage:

```python
import logging

# TODO: set APPLICATIONINSIGHTS_CONNECTION_STRING as environment variable.

properties = {'custom_dimensions': {'key_1': 'value_1', 'key_2': 'value_2'}}

# Use properties in logging statements
logger.warning('action', extra=properties)

# Use properties in exception logs
try:
    result = 1 / 0  # generate a ZeroDivisionError
except Exception:
    logger.exception('Captured an exception.', extra=properties)

```

Once this data is uploaded to App Insights, you can search in the traces and exceptions table by keys included in the customDimensions column like this:

```
traces
| where custom_dimensions.key_1 == "value_1"
| limit 50
| order by timestamp desc
```

```
exceptions
| where custom_dimensions.key_1 == "value_1"
| limit 50
| order by timestamp desc
```

Finally, this PR also contains contextual logging examples, for those who need to add context to a large number of logs without needing to add the `extra=` keyword to each log statement.


xinyi-joffre:
The CircleCI tests seem to be failing on `ModuleNotFoundError: No module named '_sqlite3'`. I don't think I changed anything that would break the existing modules. Could this be an existing issue with the build? @lzchen

contrib/opencensus-ext-azure/examples/logs/properties.py
```diff
@@ -21,4 +21,14 @@
 # and place it in the APPLICATIONINSIGHTS_CONNECTION_STRING
 # environment variable.
 logger.addHandler(AzureLogHandler())
-logger.warning('action', {'key-1': 'value-1', 'key-2': 'value2'})
+
+properties = {'customDimensions': {'key-1': 'value-1', 'key-2': 'value2'}}
```
lzchen: I would follow the [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names) standard of naming variables using lower case and underscores.

xinyi-joffre: Thanks! I will update!

xinyi-joffre:
@lzchen, do you want to me also keep the deprecated args implementation as well? I can add it, and print out a warning message about new functionality

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
```diff
@@ -0,0 +1,108 @@
...
+            return {}
+
+    def process(self, msg, kwargs):
+
+        if self.custom_dimensions:
+            if 'extra' not in kwargs:
+                kwargs['extra'] = {'custom_dimensions': self.custom_dimensions}
+
+            elif 'extra' in kwargs and 'custom_dimensions' in kwargs['extra']:
+                kwargs['extra'] = {
```
lzchen: Would this overwrite other entries in the `extra` dict?

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
```diff
@@ -0,0 +1,108 @@
...
+                    'custom_dimensions': {
+                        **self.custom_dimensions,
+                        **kwargs['extra']['custom_dimensions']
+                    }
+                }
+
```
lzchen: I think there's the missing case of if `extra` is in `kwargs` but `custom_dimensions` is not.

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
xinyi-joffre: Good catch! You are right!

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
```diff
@@ -0,0 +1,108 @@
...
+        if hasattr(record, 'custom_dimensions'):
+            record.custom_dimensions.update(self.custom_dimensions)
+        else:
+            setattr(record, 'custom_dimensions', self.custom_dimensions)
+
+        return True
+
+
+f = CustomDimensionsFilter(custom_dimensions={
```
lzchen: These are great examples!

lzchen:
> @lzchen, do you want to me also keep the deprecated args implementation as well? I can add it, and print out a warning message about new functionality

I think we want to introduce this as the proper method to be adding custom dimensions, so it's okay to make this a breaking change.

lzchen:
@xinyi-joffre  
Looks good and thanks for contributing! I will be taking a look at the build errors.

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
xinyi-joffre: Another questions is that this is currently a sample of how someone can add contextual logging easily to their code, however, they would still need to implement these classes and support them (even though they are just related to updating a properties field for log export. Is it worthwhile to consider supporting these two classes as part of the package?

lzchen: The `Azure` package supports solely on features that relate to sending telemetry to Azure Monitor. How the user wants to do this is up to them. We might make the `properties` feature more robust in the future (by adding the classes as you've mentioned) but I think for now it is fine to just include them in the example file.

lzchen:
@xinyi-joffre
Actually now that I think about it, introducing these classes in the examples might cause confusion to users (they might assume these exist as features). I think it will be better to remove this example file.

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
xinyi-joffre: Makes sense!

contrib/opencensus-ext-azure/examples/logs/contextual_logging.py
xinyi-joffre: Thank you 👍

xinyi-joffre:
I also realized that the opencensus also uses 'attributes' in tracing for contextual information. Do you like the custom_dimensions name (since it matches the Application Insights field name)? Or should I change it to something simpler like attributes, properties, etc.?

lzchen:
@xinyi-joffre 
Since this feature is specific to `Azure`, I would prefer to use `custom_dimensions`. `attributes` is taken from `span_attributes` which is a concept more related to tracing, so it is fine for logging to use a different name (metrics uses `tags` to represent custom dimensions). Also, `attributes` is a common enough word that users might be using it for other things.

lzchen:
@xinyi-joffre 
Can you rebase? I believe [#842] has fixed the build errors.

xinyi-joffre:
@lzchen this PR is ready to go, and passed the checks! Please let me know if you want me to fix anything else :). I can also help update the docs.microsoft.com entry for this once this is released.

lzchen:
@xinyi-joffre 
Thanks for contributing. I can handle updating the docs, so no worries :)

PR #842, opened 2020-01-14
Remove coverage temporarily to unblock CI

lzchen:
Unit tests are failing for CircleCI due to failed import of `sqlite3` which is a dependency for `coverage`.
All Python environments fail except for `3.4`.

@c24t did some investigation and apparently using a different CircleCI image did not make the error show up, but this does not solve the issue. Removing coverage temporarily to unblock other PRs.

image

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -52,7 +52,7 @@ def _install_dev_packages(session):
 def _install_test_dependencies(session):
     session.install('mock')
     session.install('pytest==4.6.4')
-    # session.install('pytest-cov')
+    # session.install('pytest-cov').
     session.install('retrying')
     session.install('unittest2')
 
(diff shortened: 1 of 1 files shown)


c24t:
For more context: coverage is passing locally, but failing on CircleCI because of a missing `sqlite3` installation. It's possible this is because the base image ([googleapis/nox](https://hub.docker.com/r/googleapis/nox)) is missing a required package. This image hasn't been updated in two years, so it's not clear why this would break now.

The base images that CircleCI provides for python projects ([circleci/python](https://hub.docker.com/r/circleci/python)) work as expected, but each image only includes a single version of python. Since we test multiple python versions on the same image now, we'd have to make significant changes to our nox and CircleCI config to use these images instead.

PR #848, opened 2020-01-21
Add Fixed-rate sampling logic for Azure Log Exporter

lzchen:
Implements a `SamplingFilter` class for `AzureLogHandler`.
`SamplingFilter` extends from the Python `Filter` logging [class](https://docs.python.org/3/library/logging.html#filter-objects). Upon construction of the AzureLogHandler, a filter is added to the handler which filters out log records based off of a random number generated. The fixed rate is passed in to the `AzureLogHandler` by the `logging_sampling_rate` parameter which should be in the range [0,1.0]. The default rate if none passed is 1.0.

`AzureLogHandler(logging_sampling_rate=0.5)`

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -127,8 +127,7 @@ class AzureLogHandler(TransportMixin, BaseLogHandler):
     def __init__(self, **options):
         self.options = Options(**options)
         utils.validate_instrumentation_key(self.options.instrumentation_key)
-        if self.options.logging_sampling_rate < 0 or \
-           self.options.logging_sampling_rate > 1.0:
+        if not 0 <= self.options.logging_sampling_rate <= 1:
             raise ValueError('Sampling must be in the range: [0,1]')
         self.export_interval = self.options.export_interval
         self.max_batch_size = self.options.max_batch_size
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import random
 import threading
 import time
 import traceback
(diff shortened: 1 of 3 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -116,6 +127,9 @@ class AzureLogHandler(TransportMixin, BaseLogHandler):
     def __init__(self, **options):
         self.options = Options(**options)
         utils.validate_instrumentation_key(self.options.instrumentation_key)
+        if self.options.logging_sampling_rate < 0 or \
```
hectorhdzg: nit:You can move this at the beginning of the method to avoid doing anything else if sampling rate is incorrect

lzchen: Validating the instrumentation key is also a validation and fast fail in itself.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -107,6 +108,16 @@ def stop(self, timeout=None):  # pragma: NO COVER
             return time.time() - start_time  # time taken to stop
 
 
+class SamplingFilter(logging.Filter):
```
xinyi-joffre: Is there any desire to support adaptive sampling for Application Insights as well? If so, it may be worth it to name this `FixedRateSamplingFilter` to be more specific.

lzchen: Currently in the roadmap there is no plan to add adaptive sampling. If this is ever added, it will be simple to iterate on this since the class is not exposed to customers.

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
```diff
@@ -78,6 +78,12 @@ def test_ctor(self):
         self.assertRaises(ValueError, lambda: log_exporter.AzureLogHandler())
         Options._default.instrumentation_key = instrumentation_key
 
+    def test_invalid_sampling_rate(self):
+        self.assertRaises(ValueError, lambda: log_exporter.AzureLogHandler(
```
c24t: FWIW you can do `with self.assertRaises(ValueError) as err` to avoid the lambda here.

contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
```diff
@@ -207,3 +213,37 @@ def test_log_with_invalid_custom_properties(self, requests_mock):
 
         self.assertFalse('not_a_dict' in post_body)
         self.assertFalse('key_1' in post_body)
+
+    @mock.patch('requests.post', return_value=mock.Mock())
```
c24t: Surprising to see `post` patched here, but the only easy alternative I see is `TransportMixin._transmit`, which isn't obviously a better target.

PR #851, opened 2020-01-27
Implement TelemetryProcessors for Azure exporters

lzchen:
Implement TelemetryProcessors for Azure Exporters.

TelemetryProcessors allow users to define a callback function to modify the envelopes before they are sent to the Azure Monitor backend. They can also be used to omit envelopes from being exported. Refer to the `README` on how users can do this and what the schema for the payloads look like.

See [this article](https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-filtering-sampling#javascript-web-applications) on how other SDKs are doing this in other languages for Azure Monitor.

PR #845, opened 2020-01-15
Add local storage and retry logic for Azure Metrics Exporter + flush telemetry on exit

lzchen:
Similar to how `AzureLogHandler` and `AzureExporter` have local storage and retry logic, metrics exporter now implements these as well.

Also exports all current metrics on application exit.

Resolves [#786]

PR #853, opened 2020-02-03
Release for v0.7.7

lzchen:
Minor release. Consists of mostly `azure` related changes.

Trace span attribute changes: [#827]
Log Exporter custom dimension API changes: [#837]
Local storage + retry for metrics exporter: [#845]
Fixed rate sampling for log exporter: [#848]
Telemetry processors for exporters: [#851]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 0.7.7
+Released 2020-01-31
+- Update `azure` module
+  ([#837](https://github.com/census-instrumentation/opencensus-python/pull/837))
+
+
+
 ## 0.7.6
 Released 2019-11-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@
 ## Unreleased
 
 ## 0.7.7
-Released 2020-01-31
+Released 2020-02-03
 - Update `azure` module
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+## 0.7.7
+Released 2020-02-03
```
c24t:
```suggestion
Released 2020-02-03

```

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+- Update `azure` module
```
c24t:
```suggestion
- Updated `azure` module
```

To be consistent with the others

lzchen:
> Do we need to update the minimum required `opencensus` version for `opencensus-ext-httplib` and `opencensus-ext-requests`?
> 
> Look good otherwise.

Correct if if I'm wrong, but don't we need to update the minimum required version of `opencensus` only if there were new changes in `opencensus` that `opencensus-ext-httplib` and `opencensus-ext-requests` would rely on?

c24t:
You're right, I see that these changes just add some new span attributes in the ext packages.

You might want to change the message in the changelog though. "Support span attribute changes in `opencensus>=0.7.7`" makes it sound like the ext libraries were changed to support attribute changes in the core library.

PR #867, opened 2020-03-07
Travis Integration

bu3alwa:
Continuing off #402, I tried to follow in opentelemetary `tox.ini`. Right now the only two failing tests in with tox are lint and docs and both fail due to warnings. 

With the added eachdist.py from opentelemetry we get a warning when `black . --diff --check` is run. There would be reformating needed for the warnings to go away and the tests to succeed. 

The second warning is when building the docs its missing `opencensus.ext`.

Since I am new to this project I am not sure how to proceed. For the first warning is it safe to reformat the source with black? And then for the second I am not sure why ext is missing from opencensus module. To run those specific test `tox -e lint` and `tox -e docs`

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,0 +1,7 @@
+pylint~=2.3
+flake8
+isort~=4.3.21
+sphinx~=2.1
+pytest==4.6.4
+pytest-cov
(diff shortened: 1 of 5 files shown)


bu3alwa:
@googlebot I signed it!

PR #868, opened 2020-03-10
Rename long named tests + refactor tox/nox

lzchen:
1. In Windows, `os.rename` fails sometimes when the file name created in a test blob creates a full path that is too long (260 characters). Changing names of certain tests that hit this use case.
See [this](https://stackoverflow.com/questions/51618629/python-large-file-name-rename-error-win-error-3)

2. Also, pinning the `mock` library to a specific version due to new updates causing infinite recursion errors.

3. CircleCI is failing due to error in `py27`. Update docker image to [0.18.2](https://hub.docker.com/layers/googleapis/nox/0.18.2/images/sha256-c85187d9a80cff7b7412592e52a3f15dce18c1f2298b57a3f3998d29409acc28?context=explore) to fix this.
image

4. Removed Python3.4 unit test env. Tests were failing due to `psycopg2` not installing properly in the virtualenv. We have seen this problem before [here](https://github.com/census-instrumentation/opencensus-python/pull/602#issuecomment-480023045) but the solution no longer works.

image


lzchen:
https://github.com/census-instrumentation/opencensus-python/issues/877

PR #876, opened 2020-03-31
Implement max size logic for local storage for Azure exporters

lzchen:
Previously, `storage_max_size` configuration in exporters did not do anything.
Implements this logic in `storage` to stop storing files in local storage once max size has been reached (may have some spillage if the envelope is actually larger than the remaining size).

Runs size check everytime a `put` is called in storage, instead of saving the size of the folder. Keeping state will require us to handle concurrency issues and also introduces complications when adding/deleting.

See [dot net](https://github.com/microsoft/ApplicationInsights-dotnet/blob/e2752df59499dd364469fff7487470a4117e32d7/BASE/src/ServerTelemetryChannel/Implementation/TransmissionStorage.cs#L18-L25) and [java](https://github.com/microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java#L140) implementations of this logic.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -97,7 +97,7 @@ class Options(BaseObject):
         minimum_retry_interval=60,  # minimum retry interval in seconds
         proxy=None,
         storage_maintenance_period=60,
-        storage_max_size=100*1024*1024,  # 100MiB
+        storage_max_size=50*1024*1024,  # 50MiB
         storage_path=os.path.join(
             os.path.expanduser('~'),
             '.opencensus',
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -97,7 +97,7 @@ class Options(BaseObject):
         minimum_retry_interval=60,  # minimum retry interval in seconds
         proxy=None,
         storage_maintenance_period=60,
-        storage_max_size=100*1024*1024,
+        storage_max_size=100*1024*1024,  # 100MiB
         storage_path=os.path.join(
             os.path.expanduser('~'),
             '.opencensus',
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
@@ -1,9 +1,12 @@
 import datetime
 import json
+import logging
 import os
 import random
 
 from opencensus.common.schedule import PeriodicTask
+
+logger = logging.getLogger(__name__)
 
 
 def _fmt(timestamp):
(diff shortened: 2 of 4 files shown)


lzchen:
CI blocked. Awaiting deprecation of CircleCI.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, *args, **kwargs):
         minimum_retry_interval=60,  # minimum retry interval in seconds
         proxy=None,
         storage_maintenance_period=60,
-        storage_max_size=100*1024*1024,
+        storage_max_size=100*1024*1024,  # 100MiB
```
hectorhdzg, line 100: It would be good to change it to 50 MB to align with .NET

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
```diff
@@ -170,3 +175,22 @@ def put(self, data, lease_period=0, silent=False):
...
+    def _check_storage_size(self):
+        size = 0
+        for dirpath, dirnames, filenames in os.walk(self.path):
+            for f in filenames:
+                fp = os.path.join(dirpath, f)
+                # skip if it is symbolic link
+                if not os.path.islink(fp):
+                    size += os.path.getsize(fp)
+                    if size >= self.max_size:
+                        logger.warning(
```
hectorhdzg, line 188: I'm thinking this log could be bad here, the scenario where the storage size limit is reached is mostly because the machine cannot reach Azure endpoint for some reason and all telemetry is being rejected, if they are also sending log telemetry this could cause a perf impact depending on how often the telemetry is being sent, I can see other SDKs actually log the error so not a big deal just something to consider.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
```diff
@@ -170,3 +175,22 @@ def put(self, data, lease_period=0, silent=False):
...
         return blob.put(data, lease_period=lease_period, silent=silent)
+
+    def _check_storage_size(self):
+        size = 0
+        for dirpath, dirnames, filenames in os.walk(self.path):
+            for f in filenames:
+                fp = os.path.join(dirpath, f)
+                # skip if it is symbolic link
+                if not os.path.islink(fp):
+                    size += os.path.getsize(fp)
```
hectorhdzg, line 186: It would be good to have try/except logic here in case something goes wrong reading the files

contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
lzchen: Loggers use `name` to determine uniqueness. The logger used here is not the logger used in the user's application, it is the logger corresponding to `storage.py`, so no telemetry will ever be sent.

lzchen:
Closing due to pending design discussions with local storage.

PR #885, opened 2020-04-24
Add exception tracing to django middleware

tonybaloney:
Adds additional functionality to the Django middleware to capture exceptions and add the error message, type, and traceback to the trace span attributes.

Will attempt to use both Python 2 and 3 tracebacks

I'm new to this project, so please let me know if there's anything that needs changing

PR #886, opened 2020-04-24
Enhance Azure trace exporter to emit exception envelopes for failed SERVER kinds

tonybaloney:
- Emits ExceptionData envelopes for SERVER SpanKind, if the SpanKind has an error message
- Includes the exception stack trace if present
- Sets the outerId of the the ExceptionData envelope to the Request envelope

I had to change the internal method to a generator to support this, as it only expected 1 envelope per span data beforehand.


tonybaloney:
It seems to emit the same envelope twice. will work on this

lzchen:
I believe we want to leave it up to the user to choose whether or not they want to send exception telemetry as well as configuring what would be in the payload. This is why we separate out exception data sending into [the loghandler](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py#L210).

tonybaloney:
@lzchen thanks
This could be optional, but my understanding of the the logger is that it would require the developer to send the exception to the logger and it loses all the context of the request which caused the exception, like the operation ID, request state, etc. 

This is related to PR #885 where I’m adding unhandled request exceptions to the request properties. I can’t see an Exception property in the RequestData model?

lzchen:
@tonybaloney 
Users can correlate the log messages that are sent with the spans that they are logged from with the [logging integration](https://github.com/census-instrumentation/opencensus-python/tree/master/contrib/opencensus-ext-logging).

tonybaloney:
Ready for review @lzchen

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -75,6 +76,21 @@ def span_data_to_envelope(self, sd):
                 sd.parent_span_id,
             )
         if sd.span_kind == SpanKind.SERVER:
+            if 'error.message' in sd.attributes:
```
lzchen, line 79: Can we use `attributes_helper.COMMON_ATTRIBUTES['ERROR_NAME']` instead of a hardcoded string?

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -62,7 +63,7 @@ def __init__(self, **options):
         super(AzureExporter, self).__init__(**options)
         atexit.register(self._stop, self.options.grace_period)
 
-    def span_data_to_envelope(self, sd):
+    def span_data_to_envelopes(self, sd):
```
lzchen, line 66: Because "technically" this is exposed in the api, I would not want to change the name of this function. Making it plural also doesn't add TOO much value.

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -75,6 +76,21 @@ def span_data_to_envelope(self, sd):
...
             )
         if sd.span_kind == SpanKind.SERVER:
+            if 'error.message' in sd.attributes:
+                envelope.name = 'Microsoft.ApplicationInsights.Exception'
+                data = ExceptionData(
+                    exceptions=[{
+                        'id': 1,
+                        'outerId': '{}'.format(sd.span_id),
+                        'typeName': sd.attributes.get('error.name', ''),
+                        'message': sd.attributes['error.message'],
```
lzchen, line 86: Same here

lzchen (changes requested):
Some comments

lzchen:
Please add a CHANGELOG entry as well.

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
lzchen, line 79: We probably want this feature for SpanKind.CLIENT too.

tonybaloney: I'd rather do that in a separate PR as it'll need testing in other places.

tonybaloney:
@lzchen requested changes have all been made

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -23,11 +23,22 @@
...
     Data,
     Envelope,
+    ExceptionData,
     RemoteDependency,
     Request,
 )
 from opencensus.ext.azure.common.storage import LocalFileStorage
 from opencensus.ext.azure.common.transport import TransportMixin
+from opencensus.ext.django.middleware import (
+    ERROR_MESSAGE,
```
lzchen, line 33: We don't want to be taking a dependency on `django` integration. Please use `attributes_helper.COMMON_ATTRIBUTES['ERROR_NAME']` directly.

tonybaloney: Done

lzchen (approved):

PR #887, opened 2020-04-30
Default FileExporter file_mode to a

gjoseph92:
The current default, `w+`, overwrites the contents of the file on the first write. Since `emit` opens and writes to the file every time it's called, this effectively rewrites the file on every `emit`. Perhaps this is intended behavior, but it took me a while to understand why there was only one span written to my file, even though I could tell multiple were being emitted. I didn't have my [Python file mode tables](https://docs.python.org/3/library/functions.html#open) memorized, so at first glance `w+` sounded like "write mode, + for append", and I overlooked it while debugging.

I think `a` (append) would be a more intuitive default.


gjoseph92:
@googlebot I signed it!

Issue #896, opened 2020-05-29
Initializing with the "proxy" option passed in doesn't appear to work when initializaing AzureLogHandler and AzureExporter


skinamdar:
https://github.com/census-instrumentation/opencensus-python/blob/99821a0b826445cecce0881a01790c8cc6958ceb/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py#L98

See partial sample invocation below where the proxy is passed in. 
```
from opencensus.ext.azure.log_exporter import AzureLogHandler
from opencensus.ext.azure.trace_exporter import AzureExporter
non_azure_https_proxy = {"https": "https://www.google.com"} # to test that the command fails
app_insights_connection_string = "insert_here"
azure_logger = AzureLogHandler(connection_string=app_insights_connection_string, proxy=azure_https_proxy)
azure_exporter = AzureExporter(connection_string=app_insights_connection_string, proxy=azure_https_proxy)
```

Running a similar command over curl 
```
curl -d '{"name": "MetricData", "time":"2020-05-29T00:18:28.9586379Z","iKey":"{insert_i_key_here}","data":{"baseType":"MetricData","baseData":{"metrics":[{"name":"Custom metric","value":2,"count":1}]}}}' https://dc.services.visualstudio.com/v2/track -x https://www.google.com -v
```
results in a curl address error (which we expect as we want to make sure the requests passed via opencensus are routed to a proxy which we've defined)

Can you please update the code to make sure it works with a custom proxy passed in?

skinamdar:
Also proxy appears in only one line `➜  opencensus-python git:(master) grep -r "proxy" *
contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py:        proxy=None,`

lzchen:
Yes `proxy` as a configuration does not do anything at the moment.

skinamdar:
@lzchen Can the proxy functionality be enabled by the library as a bug fix?

lzchen:
I'm tagging this as a feature request.

Please include a description of the behaviour that you want. Can you also reformat your original post so that the code sample looks more readable?

skinamdar:
Thank you. 

Behavior description:
[...]

PR #902, opened 2020-06-11
Add support to initialize azure exporters with proxies

skinamdar:
Context:
Proxy functionality when instantiating azure exporters is currently unsupported so I had created issue https://github.com/census-instrumentation/opencensus-python/issues/896. To expedite this feature request, I created this PR which supports posting requests with proxies that are passed in. 

Testing:
* Tested locally to ensure proxy options were set. 
* Tested using nox 3.6 - all tests pass. Note that there are existing linting issues for files unrelated to this PR. To decouple changes, I did not address them in the PR. However, I can fix those linting issues if the maintainers prefer that in this PR. 
flake_8_error

* Nox status
```
nox > Session docs was successful.
nox > Ran multiple sessions:
nox > * unit-2.7: failed
nox > * unit-3.5: success
nox > * unit-3.6: success
nox > * system-2.7: skipped
nox > * system-3.6: skipped
nox > * lint: failed
nox > * lint_setup_py: success
nox > * docs: success
```

Additional Notes:
* Running nox with python 2.7 fails on my computer when trying to install `contrib/opencensus-ext-grpc` - I'm still troubleshooting that but wanted to create this PR as it's been tested with python3 at least. If the maintainers could help run this PR on their local machine with nox to ensure python2 compatibility that would be super appreciated as well.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -56,6 +56,9 @@ def process_options(options):
                 TEMPDIR_PREFIX + TEMPDIR_SUFFIX
             )
 
+    if options.proxies is None:
+        options.proxies = '{}'
+
 
 def parse_connection_string(connection_string):
     if connection_string is None:
@@ -105,7 +108,7 @@ class Options(BaseObject):
         logging_sampling_rate=1.0,
         max_batch_size=100,
         minimum_retry_interval=60,  # minimum retry interval in seconds
-        proxy=None,
+        proxies=None,  # string maps url schemes to the url of the proxies
         storage_maintenance_period=60,
         storage_max_size=50*1024*1024,  # 50MiB
         storage_path=None,
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
@@ -52,6 +52,7 @@ class TransportMixin(object):
                     'Content-Type': 'application/json; charset=utf-8',
                 },
                 timeout=self.options.timeout,
+                proxies=json.loads(self.options.proxies),
             )
         except requests.Timeout:
             logger.warning(
--- a/contrib/opencensus-ext-azure/CHANGELOG.md
+++ b/contrib/opencensus-ext-azure/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Change default path of local storage
   ([#903](https://github.com/census-instrumentation/opencensus-python/pull/903))
+- Add support to initialize azure exporters with proxies
+  ([#902](https://github.com/census-instrumentation/opencensus-python/pull/902))
 
 ## 1.0.1
 Released 2019-11-26
--- a/contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
+++ b/contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
@@ -85,6 +85,17 @@ class TestAzureLogHandler(unittest.TestCase):
                 logging_sampling_rate=4.0,
             )
 
+    def test_init_handler_with_proxies(self):
+        handler = log_exporter.AzureLogHandler(
+            instrumentation_key='12345678-1234-5678-abcd-12345678abcd',
+            proxies='{"https":"https://test-proxy.com"}',
+        )
+
+        self.assertEqual(
+            handler.options.proxies,
+            '{"https":"https://test-proxy.com"}',
+        )
+
     @mock.patch('requests.post', return_value=mock.Mock())
     def test_exception(self, requests_mock):
         logger = logging.getLogger(self.id())
--- a/contrib/opencensus-ext-azure/tests/test_azure_trace_exporter.py
+++ b/contrib/opencensus-ext-azure/tests/test_azure_trace_exporter.py
@@ -45,6 +45,17 @@ class TestAzureExporter(unittest.TestCase):
         self.assertRaises(ValueError, lambda: trace_exporter.AzureExporter())
         Options._default.instrumentation_key = instrumentation_key
 
+    def test_init_exporter_with_proxies(self):
+        exporter = trace_exporter.AzureExporter(
+            instrumentation_key='12345678-1234-5678-abcd-12345678abcd',
+            proxies='{"https":"https://test-proxy.com"}',
+        )
+
+        self.assertEqual(
+            exporter.options.proxies,
+            '{"https":"https://test-proxy.com"}',
+        )
+
     @mock.patch('requests.post', return_value=mock.Mock())
     def test_emit_empty(self, request_mock):
         exporter = trace_exporter.AzureExporter(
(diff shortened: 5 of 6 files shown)


skinamdar:
@googlebot I fixed it.

skinamdar:
I see the linting issues are causing CI to fail - I will fix those items now

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, *args, **kwargs):
         logging_sampling_rate=1.0,
         max_batch_size=100,
         minimum_retry_interval=60,  # minimum retry interval in seconds
-        proxy=None,
+        proxy=None,  # string maps url schemes to the url of the proxies
```
lzchen, line 98:
```suggestion
        proxies=None,  # string maps url schemes to the url of the proxies
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -42,6 +42,10 @@ def _transmit(self, envelopes):
         This function should never throw exception.
         """
         try:
+            proxies = None
```
lzchen, line 45: Do you need this line?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -42,6 +42,10 @@ def _transmit(self, envelopes):
...
+            if self.options.proxy:
+                proxies = json.loads(self.options.proxy)
```
lzchen, line 47:
```suggestion
                proxies = json.loads(self.options.proxies)
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
skinamdar: I set proxies to None because running a `json.loads(None)`, if no proxies are passed in, results in an exception. Does that make sense?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
skinamdar: I like the idea about making sure the variable name matches the options name. I had used `self.options.proxy` to be backward compatible with the existing ```proxy``` Option. Are we comfortable updating either renaming the existing ```proxy``` field or introducing a new ```proxies``` Option with the default value of None? 

Since the currently ```proxy``` variable isn't actually used in Master, to reduce duplication, I'm leaning towards the rename so wanted to get your thoughts on that.

skinamdar:
> LGTM.
> 
> I've disabled the CircleCI checks since TravisCI is already testing everything we need. All you need to do is fix the build in TravisCI and then should be good to merge!

Great thanks. Do you know how I can re-start the build? I checked the log to identify if it's a CI error which perhaps restarting will fix it - or an actual failure. I ran nox locally with 3.7 and tests passed so wanted to see if there's a way to restart the build.
nox3 7

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -50,6 +54,7 @@ def _transmit(self, envelopes):
                     'Content-Type': 'application/json; charset=utf-8',
                 },
                 timeout=self.options.timeout,
+                proxies=proxies,
```
lzchen, line 57: What happens if proxies is an empty dict? Does it do anything?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: If `self.options.proxy` is None, are you able to change it to an empty dict string?

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: Agreed. Since `proxy` isn't implemented, the name change should be fine.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
skinamdar: Great suggestion, I tested that we're able to send a request with ```proxies=json.loads("{}")```. I will update the code to reflect that by setting the default proxies option to be ```"{}"```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
skinamdar: Yes the request still passes - thanks for the suggestion - updating now.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: Awesome. Make sure to initialize it in `process_options` instead of having a default `{}` value for proxies.

lzchen:
@skinamdar 
To run the build again, simply push another commit to your PR branch.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
skinamdar: I found that I need to initialize that value to `"{}"` in the Options map because even if we set it to a dictionary in process options , that options.proxies values ends up being converted to a string - I found this when running the unit tests. 

So it seems we'd have to keep that json.loads behavior in transport.py to make sure that argument is correctly loaded as a dictionary.

lzchen: You don't set it to a dictionary in process options. You set it to a string with an empty dictionary like you've said: `"{}"`.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, *args, **kwargs):
         logging_sampling_rate=1.0,
         max_batch_size=100,
         minimum_retry_interval=60,  # minimum retry interval in seconds
-        proxy=None,
+        proxies='{}',  # string maps url schemes to the url of the proxies
```
lzchen, line 98: I don't think I was clear in my previous explanation. You keep proxies=None in the constructor, but in `process_options` you check if proxies is None and you set it to `"{}"` in that function.

lzchen:
Also, can  you add an entry into CHANGEOG.md under opencensus-ext-azure?

skinamdar:
Currently blocked in travis ci for python2.7 and python3.4 builds due to https://github.com/census-instrumentation/opencensus-python/issues/905 

Still trying to investigate why the python3.7 build is failing since it's not clear from the logs

lzchen:
@skinamdar 
If you rebase with master, builds should pass

skinamdar:
Awesome, I see the build passed now. Thank you @lzchen for helping fix the builds - can you confirm I'm all set to merge now then?

skinamdar:
@lzchen Thank you for merging the changes to master. Do you know when the next release of opencensus-ext-azure will be in pip so we can pull in these changes when developing?

lzchen:
I'll probably do a release later next week.

skinamdar:
Sounds good, thank you.

PR #909, opened 2020-06-15
Fix async.py is_exporter context

lzchen:
Fixes `set_is_exporter` in async.py and moves it within the created thread.
Also resets the context in sync.py.


opencensus/common/transports/async_.py
```diff
@@ -141,9 +144,6 @@ def start(self):
             self._thread = threading.Thread(
                 target=self._thread_main, name=_WORKER_THREAD_NAME)
             self._thread.daemon = True
-            # Indicate that this thread is an exporter thread. Used for
-            # auto-collection.
-            execution_context.set_is_exporter(True)
```
hectorhdzg, line 146: So this will never be set as False right? are you moving this just to do this multiple times instead of a single time?

PR #911, opened 2020-06-17
Use chain.from_iterable in span.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.


aabmass:
Nice! Just a style/readability question, any thoughts on using two for loops instead?

cool-RR:
Thank you :) I would prefer this declarative approach rather than two for-loops, which is more imperative. That's a matter of taste though, I can see why some people would prefer the for-loops.

What I'd _really_ prefer is to cut down the existing for loop by doing `yield from chain.from_iterable(...)` but that's gonna have to wait until the project drops Python 2 support :)

aabmass:
If you want to get rid of the for loop, can you do something like this?
```py
return chain(
    chain.from_iterable(map(iter, self.children)),
    [self],
)
```

cool-RR:
I considered it, and it'll work, but it's debatable whether it's nicer than the existing version. I'd say 50-50.

aabmass:
Thank you @cool-RR!

Issue #927, opened 2020-07-11
Adding atexit handler to stackdriver exporter


aravinsiva:
This is needed such that SIGTERM's can be handled in OpenCensus in order to publish any remaining telemetry data to the GCP backend. Currently this exists in the azure exporter but not in the stackdriver exporter.

**Describe the solution you'd like.**
Add an atexit handler to the stackdriver expoter such that remaining telemetry is flushed upon SIGTERM being handled.

**Describe alternatives you've considered.**
Currently exists in Opentelemetry however since support for OpenCensus will continue for some time it would be advantageous to include this in OpenCensus. 

The following need to be done:

- Use Async transporter as the default transporter for Stackdriver trace exporter such that the atexit handler is used by default and main thread is not blocked (Update README).

- Create atexit handler for Stackdriver metrics exporter

aabmass:
@aravinsiva do you want me to assign this to you?

aravinsiva:
Yes that would be great!

aravinsiva:
Issue is currently being addressed in the above two PR's.

PR #929, opened 2020-07-16
Changing default transport for stackdriver exporter

aravinsiva:
Changing default transporter to async as sync leads to latency issues. In addition will allow GCP Cloud Run clients to leverage atexit abilities of async exporter. Also addresses the following issue:
https://github.com/census-instrumentation/opencensus-python/issues/927

(This issue is addressed in two PR's. One for setting the default trace processor and the other for adding atexit handlers in the metrics exporter).

### DIFF
--- a/contrib/opencensus-ext-stackdriver/README.rst
+++ b/contrib/opencensus-ext-stackdriver/README.rst
@@ -35,7 +35,7 @@ This example shows how to report the traces to Stackdriver Trace:
     pip install google-cloud-trace
     pipenv install google-cloud-trace
 
-By default, traces are exported asynchronously, to avoid latency during
+By default, traces are exported asynchronously, to reduce latency during
 your code's execution. If you would like to export data on the main thread
 use the synchronous transporter:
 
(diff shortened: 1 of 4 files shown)


aabmass:
This LGTM but I will defer to someone who knows why the other option is the current default.

contrib/opencensus-ext-stackdriver/README.rst
```diff
@@ -35,20 +35,19 @@ This example shows how to report the traces to Stackdriver Trace:
     pip install google-cloud-trace
     pipenv install google-cloud-trace
 
-By default, traces are exported synchronously, which introduces latency during
-your code's execution. To avoid blocking code execution, you can initialize
-your exporter to use a background thread.
+By default, traces are exported asynchronously, to avoid latency during
```
sjbrunst, line 38: nit: avoid -> reduce

I doubt even the async exporter can completely avoid latency.

lzchen:
> This LGTM but I will defer to someone who knows why the other option is the current default.

I think @c24t  would have more insight on this.

c24t:
> This LGTM but I will defer to someone who knows why the other option is the current default.

The decision to use the sync transport was made all the way back in https://github.com/census-instrumentation/opencensus-python/pull/63. It looks like that PR added support for async/background transport, but didn't change the default. In the meantime we've started using async transport as the default for other exporters, including OT exporters. I think we can do the same here.

PR #934, opened 2020-07-20
adding PeriodicMetricTask atexit handler

aravinsiva:
Adding atexit handlers to periodic tasks in order to flush all remaining data to backend on shutdown. Atexit handlers are similarly used in the async trace processor shown here:

https://github.com/census-instrumentation/opencensus-python/blob/1328fbfb14d3a26c3b585bbfa275d6036d490cd8/opencensus/common/transports/async_.py#L148

This PR also addresses this issue: 
https://github.com/census-instrumentation/opencensus-python/issues/927

### DIFF
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -48,7 +48,6 @@ class PeriodicTask(threading.Thread):
 
     def run(self):
         wait_time = self.interval
-        atexit.register(self.function, *self.args, **self.kwargs)
         while not self.finished.wait(wait_time):
             start_time = time.time()
             self.function(*self.args, **self.kwargs)
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -46,7 +46,6 @@ class PeriodicTask(threading.Thread):
         self.kwargs = kwargs or {}
         self.finished = threading.Event()
 
-
     def run(self):
         wait_time = self.interval
         atexit.register(self.function, *self.args, **self.kwargs)
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -14,7 +14,6 @@
 
 from six.moves import queue
 
-import atexit
 import threading
 import time
 
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -14,9 +14,9 @@
 
 from six.moves import queue
 
+import atexit
 import threading
 import time
-import atexit
 
 
 class PeriodicTask(threading.Thread):
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -16,6 +16,7 @@ from six.moves import queue
 
 import threading
 import time
+import atexit
 
 
 class PeriodicTask(threading.Thread):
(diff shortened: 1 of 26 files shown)


opencensus/common/schedule/__init__.py
```diff
@@ -47,6 +48,7 @@ def __init__(self, interval, function, args=None, kwargs=None, name=None):
 
     def run(self):
         wait_time = self.interval
+        atexit.register(self.function, *self.args, **self.kwargs)
```
lzchen, line 51: There are multiple classes that use the `PeriodicTask` class for workers, and not all of them want to flush on exit (for [example](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py#L93)). Maybe have this functionality only for the classes that require this kind of behaviour (e.g. PeriodicMetricTask).

Also, should the worker itself be responsible for an `atexit` implementation? In OpenTelemetry, we have a top level class that the user interacts with [MeterProvider](https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/metrics/__init__.py#L508) that is responsible for `atexit` behaviour, which in turn calls the "stop" functions of workers. This makes is so that the worker itself does not have to be forced to do a certain behaviour for all usages.

aravinsiva: Moved the atexit handler to the run method of PeriodicMetricTask.

aravinsiva: Moved the handler to PeriodicMetricTask

lzchen:
@aravinsiva 
I don't think my previous comment has been addressed.

```
Also, should the worker itself be responsible for an atexit implementation? In OpenTelemetry, we have a top level class that the user interacts with MeterProvider that is responsible for atexit behaviour, which in turn calls the "stop" functions of workers. This makes is so that the worker itself does not have to be forced to do a certain behaviour for all usages.
```

aravinsiva:
> @aravinsiva
> I don't think my previous comment has been addressed.
> 
> > Also, should the worker itself be responsible for an atexit implementation? In OpenTelemetry, we have a top level class that the user interacts with MeterProvider that is responsible for atexit behaviour, which in turn calls the "stop" functions of workers. This makes is so that the worker itself does not have to be forced to do a certain behaviour for all usages.

I moved the atexit handler to the PeriodicMetricTask such that only usages that would need the handler use it. With this implementation (in PeriodicMetricTask) Is your concern the use of the atexit handler will still be implemented in instances where it is not wanted?

lzchen:
@aravinsiva 
> I moved the atexit handler to the PeriodicMetricTask such that only usages that would need the handler use it. With this implementation (in PeriodicMetricTask) Is your concern the use of the atexit handler will still be implemented in instances where it is not wanted?

Your change addresses the first part of my comment, moving the logic to the `PeriodicMetricTask` is the correct thing to do. The second part of my comment refers to should the `atexit` logic be even part of the worker class? As mentioned before, the API classes that the user instantiates themselves (exporters in OpenCensus, MeterProvider in OpentElemetry), should be the ones that are responsible to hold that logic. Within that logic, it simply calls the worker.stop methods of the workers. See the [code](https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/metrics/__init__.py#L527) in OpenTelemetry.

lzchen:
I've actually implemented this myself in https://github.com/census-instrumentation/opencensus-python/pull/943 for AM exporter.
Feel free to take a look.

lzchen:
Closing this due to [#943]

aravinsiva:
> Closing this due to [#943]

Sorry leighton I have been busy with a more pressing project so I put this on the back burners. But I believe the same idea just needs to be implemented in the stackdriver exporter correct? And then we should be good to go?

PR #935, opened 2020-07-22
Attach rate metrics for VM

lzchen:
Continuation of [#930].

Detects whether current app is hosted in an Azure VM. We utilize [AzureMetaDataService](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service) to retrieve vm information. We need to periodically retrieve this (not only on load) because this information can change.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
@@ -16,15 +16,15 @@ import datetime
 import json
 import os
 import platform
-import requests
 from collections import OrderedDict
+
+import requests
 
 from opencensus.common.version import __version__ as opencensus_version
 from opencensus.ext.azure.common.version import __version__ as ext_version
 from opencensus.metrics.export.gauge import LongGauge
 from opencensus.metrics.label_key import LabelKey
 from opencensus.metrics.label_value import LabelValue
-
 
 _AIMS_URI = "http://169.254.169.254/metadata/instance/compute"
 _AIMS_API_VERSION = "api-version=2017-12-01"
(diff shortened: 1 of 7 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
```diff
@@ -48,17 +85,32 @@ def __init__(self):
...
-    def __call__(self):
-        """ Returns a derived gauge for the heartbeat metric.
+    def get_azure_compute_metadata(self):
+        try:
+            request_url = "{0}?{1}&{2}".format(
+                _AIMS_URI, _AIMS_API_VERSION, _AIMS_FORMAT)
+            response = requests.get(request_url, headers={"MetaData": "True"})
+        except requests.exceptions.ConnectionError:
+            # Not in VM
+            self.is_vm = False
```
hectorhdzg, line 105: Is connection error enough to determine is not a VM?, maybe adding generic except as well to be sure doesn't think is a VM

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
```diff
@@ -12,22 +12,59 @@
...
+            list(self.properties.values())
+        )
+
+    def get_metrics(self):
+        if self.is_vm:
+            # Only need to update if in vm (properties could change)
+            self.properties.clear()
+            self.update_properties()
+            self.heartbeat.clear()
+            # pylint: disable=protected-access
```
hectorhdzg, line 58: Is no other way to update the label keys value?, maybe creating the LongGauge again?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
lzchen: You are correct about the second except. I'll add it to catch anything else.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py
lzchen: I was thinking that maybe this would eventually take up too much memory, but Python gc is pretty efficient so we can do it this way.

Issue #790, opened 2019-09-20
gRpc client interceptor throws uncatchable exception when rpc is cancelled


z-oo:
**Describe your environment.** Describe any aspect of your environment relevant to the problem, including your Python version, [platform](https://docs.python.org/3/library/platform.html), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on master.

Python version: 2.7
Code built under bazel 0.28.1

**Steps to reproduce.**
1. Send an opencensus instrumented grpc
2. cancel the grpc
3. client side application crashes and a stacktrace like this:
```
Traceback (most recent call last):
    File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
        self.run()
      File "/usr/lib/python2.7/threading.py", line 754, in run
        self.__target(*self.__args, **self.__kwargs)
      File "/home/hlightman/worktrees/scene-finder-profile/bazel-bin/tool/scene_finder/query/server/query_server.runfiles/pip_deps/grpc/_channel.py", line 701, in channel_spin
        completed_call = event.tag(event)
      File "/home/hlightman/worktrees/scene-finder-profile/bazel-bin/tool/scene_finder/query/server/query_server.runfiles/pip_deps/grpc/_channel.py", line 174, in handle_event
        callback()
      File "/home/hlightman/worktrees/scene-finder-profile/bazel-bin/tool/scene_finder/query/server/query_server.runfiles/pip_deps/grpc/_channel.py", line 313, in <lambda>
        self._state.callbacks.append(lambda: fn(self))
      File "/home/hlightman/worktrees/scene-finder-profile/bazel-bin/tool/scene_finder/query/server/query_server.runfiles/pip_deps/opencensus/ext/grpc/client_interceptor.py", line 131, in callback
        proto_message=future_response.result(),
          File "/home/hlightman/worktrees/scene-finder-profile/bazel-bin/tool/scene_finder/query/server/query_server.runfiles/pip_deps/grpc/_channel.py", line 277, in result
        raise grpc.FutureCancelledError()
    FutureCancelledError
```
Please notice that here the exception is thrown from a thread that is not owned by the application and is impossible to catch.

**What is the expected behavior?**
No exception and rpc cancellation is correctly recorded by opencensus.

**What is the actual behavior?**
An uncaught exception is thrown from grpc thread. Please see above.

**Additional context.**
[...]

PR #948, opened 2020-08-26
Check if response has exception before trying to access result

AmicoJoules:
Small fix on the callback set by the grpc client interceptor. Trying to access the result raises the Future exception which can be problematic for asynchronous grpc calls.

This PR addresses issue https://github.com/census-instrumentation/opencensus-python/issues/790


AmicoJoules:
@googlebot I fixed it.

PR #949, opened 2020-09-11
Configurable queue capacity for azure exporters

lzchen:
The trace and log exporters utilize a queue to be added onto for whenever a span/log is created to be exported.
If queue is full, telemetry is dropped and a warning is logged.

This PR adds the ability to configure the max queue size, default is 8192.


hectorhdzg (approved):
LGTM

PR #961, opened 2020-10-14
Version.py not found when installing opencensus-context from source

### LINKED ISSUE #962: Installing opencensus-context from source fails because of Version.py
Python 3.7

**Steps to reproduce.**
Describe exactly how to reproduce the error. Include a code sample if applicable.
pip install opencensus-context --no-binary

What did you expect to see?
Successful install

What did you see instead?
"version.py can not be found"

**Additional context.**
I unzipped the .tar.gz from pypi, and can see that version.py is not included in the zip. This sub package used a different process then the main package, which is why there are issues.

dciborow:
Fixes #962


dciborow:
@googlebot I signed it!

dciborow:
Opened a bug as well.
Proposing this fix, as it is the way the main repo does the version number.

aabmass:
Hey @dciborow thanks for the PR. I don't think this will work though, the `version.py` file isn't included in the tarball. Can you try updating `setup.py` to include that file ?

PR #964, opened 2020-10-24
Fix comment, AsyncTransport is the default

TheJokersThief:
Just a small comment change, noticed when I was looking through the code that the comment calls out the synchronous transport as the default but the argument defaults to the async one :)


TheJokersThief:
@googlebot I signed it!

aabmass (approved):
Thanks!

PR #986, opened 2020-12-10
Loosen ikey validation to follow GUID

### LINKED ISSUE #985: Instrumentation key of azure china east does not match regex
I deployed an azure web app running gunicorn in python 3.7 in azure china east. The versions of open census and its extensions are
* opencensus 0.7.11
* opencensus_ext_azure 1.0.5
* opencensus_context 0.1.2
* opencensus_ext_flask 0.7.3
* opencensus_ext_logging 0.1.0  

**Steps to reproduce.**
Try to run opencensus_ext_azure with an InstrumentationKey of an application insights that is deployed to azure china east

The instrumentation key should be regarded as valid, so no exception should be raised.

The issue is that the application insights InstrumentationKey does not match the regex defined here:
https://github.com/census-instrumentation/opencensus-python/blob/82379aaa99d2771dd0112551cd96b25dd52193b8/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py#L63

The instrumentation key is composed of 5 groups separated by dashes. The first two and the last two groups match the provided regex. The problem is in the 3rd group that is supposed to match `'[1-5][0-9a-f]{3}`.
In the connection string given by the app insights in azure china east this group starts with a letter, not with a digit.

As a consequence, the web app does not spin up, because the `validate_instrumentation_key` function raises `ValueError("Invalid instrumentation key.")` 

The 3rd group in the InstrumentationKey of one of the european app insights looks like this:
`-dddd-`
while that from azure china east looks like this:
`-wwwd-`

`d` stands for digit and `w` for a letter / word-character

lzchen:
Fixes #985

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
@@ -58,12 +58,9 @@ def timestamp_to_iso_str(timestamp):
     return to_iso_str(datetime.datetime.utcfromtimestamp(timestamp))
 
 
-# Validate UUID format
-# Specs taken from https://tools.ietf.org/html/rfc4122
+# Validate GUID format
 uuid_regex_pattern = re.compile('^[0-9a-f]{8}-'
-                                '[0-9a-f]{4}-'
-                                '[1-5][0-9a-f]{3}-'
-                                '[89ab][0-9a-f]{3}-'
+                                '([0-9a-f]{4}-){3}'
                                 '[0-9a-f]{12}$')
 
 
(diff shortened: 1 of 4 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -58,12 +58,9 @@ def timestamp_to_iso_str(timestamp):
     return to_iso_str(datetime.datetime.utcfromtimestamp(timestamp))
 
 
-# Validate UUID format
-# Specs taken from https://tools.ietf.org/html/rfc4122
+# Validate GUID format
 uuid_regex_pattern = re.compile('^[0-9a-f]{8}-'
```
aabmass, line 62: Is this just a vanilla UUID now? If yes, could validate with [uuid.UUID](https://docs.python.org/3/library/uuid.html#uuid.UUID). Otherwise, the regex looks reasonable to me but im not too confident

aabmass (approved):

contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
lzchen: This is the same regex used in other internal Microsoft SDKs, so I would be consistent.

PR #1002, opened 2021-01-21
remove explicit mock dependency if available in unittest

bnavigator:
Python >= 3.3 provides mock through unittest.mock. No need to explicitly install the PyPI package (soo to be deprecated by newer pythons)

### DIFF
--- a/contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
+++ b/contrib/opencensus-ext-azure/tests/test_azure_log_exporter.py
@@ -17,7 +17,10 @@ import os
 import shutil
 import unittest
 
-import mock
+try:
+    from unittest import mock
+except ImportError:
+    import mock
 
 from opencensus.ext.azure import log_exporter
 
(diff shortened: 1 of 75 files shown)


bnavigator:
@googlebot I signed it!

aabmass:
> (soo to be deprecated by newer pythons)

The package will no longer install for new versions?

bnavigator:
That's what I heard and why we create these patches for openSUSE python packages. Can't find the reference right now. Ping @mcepl

mcepl:
> The package will no longer install for new versions?

Yes, we plan to eventually remove python-mock from openSUSE packages.

aabmass:
I think I'm missing some context here. If you're building an RPM for this, you shouldn't need the test code/dependencies anyway, right?

mcepl:
> I think I'm missing some context here. If you're building an RPM for this, you shouldn't need the test code/dependencies anyway, right?

All packages on most Linux distribution run as full as possible test suites to prevent regressions.

glaubitz:
This is still unfixed in version 0.11.4.

Issue #1005, opened 2021-01-27
Is it expected when using logger.error or logger.exception, the messages are treated as trace?


star286:
Environment:

python 3.8, opencensus-ext-azure 1.0.7

**Steps to reproduce.**
the code as below:

```
logger.error("it is an error")
logger.exception("it is an exception")
```

**What is the expected behavior?**
The messages which are logged by logger.error or logger.exception should be treated as exceptions in application insights.

**What is the actual behavior?**
These messages are treated as trace in application insights

Is this an expected behavior? and any workaround to fix this?

m1nkeh:
yep, this caught me out too!

lzchen:
`logger.error` is not always an exception. `error` is a specific Python [logging level](https://docs.python.org/3/library/logging.html#logging-levels) and users may use it as a mechanism to filter what kind of logs they want to see.

`logger.exception` ARE treated as exceptions, you simply need to log them within an actual exception for them to show in App insights.

lzchen:
There seems to be a bug where if user uses `logger.exception` with no actual exception, no log is sent. Will be making a fix for this.

empowerNate:
It seems like this is still not working - when I use logging.exception, it logs as a trace and not as an exception in app insights.

lzchen:
We have changed this so that users must use `logger.exception` with an actual exception, otherwise it is treated as a `trace` so that is indeed expected behavior.

empowerNate:
Hmm ok, I tried this and I don't think it worked, but can try again.

```
try:
    ...
except Exception as e:
    logging.exception(e)
```

is that not the right way to do it?

lzchen:
@empowerNate 

Yes that is exactly how you do it. If an exception is raised in your try block, the exception should be caught and then logged with `logger.exception`. 

Be sure to use the `logger` that you added `AzureLogHandler` to, not `logging` which is the module. See [this example](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/examples/logs/error.py) for specifics.

empowerNate:
Ah, that's it - I didn't add the handler.

PR #1006, opened 2021-01-27
Fix logger.exception error when no exception present in Azure log exporter

lzchen:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/1005


hectorhdzg (approved):
LGTM

PR #1014, opened 2021-02-25
Fix duplicate spans from requests integration

lzchen:
requests integration was patching all `request.get, etc..` methods as well as `Session.request`. Only the latter is needed.


aabmass (approved):
This LGTM. Do we have any integration test that previous usages like `requests.get()` still produce spans?

lzchen:
@aabmass 
I could add integration tests for this but it seems like they are not run in [tox](https://github.com/census-instrumentation/opencensus-python/blob/master/tox.ini#L52) regardless. I believe this was because the integration tests that were specific stackdriver kept breaking so we decided to disable them.

PR #1021, opened 2021-04-07
Support AAD authentication

lzchen:
Support for AAD auth depends on `azure-core` and `azure-identity` only for the exception handling.

Refer to this [doc](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/README.md#error-handling) for when each exception is thrown. One is a retry case, another is a (possible) drop case.

Users must pass in a [TokenCredential](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/credentials.py#L16) to exporter constructors to utilize this feature. Refer to [this](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/README.md#credential-classes) for a list of valid credential classes.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
@@ -48,6 +48,14 @@ def process_options(options):
         or 'https://dc.services.visualstudio.com'
     options.endpoint = endpoint + '/v2/track'
 
+    # Authorization
+    # `azure.core.credentials.TokenCredential` class must be valid
+    if options.credential and \
+        not hasattr(options.credential, 'get_token'):
+        raise ValueError(
+            'Must pass in valid TokenCredential.'
+        )
+
     # storage path
     if options.storage_path is None:
         TEMPDIR_SUFFIX = options.instrumentation_key or ""
@@ -101,6 +109,7 @@ class Options(BaseObject):
 
     _default = BaseObject(
         connection_string=None,
+        credential=None,
         enable_local_storage=True,
         enable_standard_metrics=True,
         endpoint='https://dc.services.visualstudio.com/v2/track',
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
@@ -17,8 +17,11 @@ import logging
 
 import requests
 
-logger = logging.getLogger(__name__)
+from azure.core.exceptions import ClientAuthenticationError
+from azure.identity._exceptions import CredentialUnavailableError
 
+logger = logging.getLogger(__name__)
+_MONITOR_OAUTH_SCOPE = "https://monitor.azure.com/.default"
 
 class TransportMixin(object):
     def _transmit_from_storage(self):
@@ -45,13 +48,17 @@ class TransportMixin(object):
         if not envelopes:
             return 0
         try:
+            headers = {
+                'Accept': 'application/json',
+                'Content-Type': 'application/json; charset=utf-8',
+            }
+            if self.options.credential:
+                token = self.options.credential.get_token(_MONITOR_OAUTH_SCOPE)
+                headers["Authorization"] = "Bearer {}".format(token.token)
             response = requests.post(
                 url=self.options.endpoint,
                 data=json.dumps(envelopes),
-                headers={
-                    'Accept': 'application/json',
-                    'Content-Type': 'application/json; charset=utf-8',
-                },
+                headers=headers,
                 timeout=self.options.timeout,
                 proxies=json.loads(self.options.proxies),
             )
@@ -59,11 +66,26 @@ class TransportMixin(object):
             logger.warning(
                 'Request time out. Ingestion may be backed up. Retrying.')
             return self.options.minimum_retry_interval
-        except Exception as ex:  # TODO: consider RequestException
+        except requests.RequestException as ex:
             logger.warning(
                 'Retrying due to transient client side error %s.', ex)
             # client side error (retryable)
             return self.options.minimum_retry_interval
+        except CredentialUnavailableError as ex:
+            logger.warning(
+                'Error with credential configuration %s', ex
+            )
+            return -1
+        except ClientAuthenticationError as ex:
+            logger.warning(
+                'Error getting token %s', ex
+            )
+            return self.options.minimum_retry_interval
+        except Exception as ex:
+            logger.warning(
+                'Error when sending request %s. Dropping telemetry.', ex)
+            # Extraneous error (non-retryable)
+            return -1
 
         text = 'N/A'
         data = None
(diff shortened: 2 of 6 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -45,25 +48,44 @@ def _transmit(self, envelopes):
...
+                'Error with credential configuration %s', ex
+            )
+            return -1
+        except ClientAuthenticationError as ex:
+            logger.warning(
+                'Error getting token %s', ex
+            )
+            return self.options.minimum_retry_interval
+        except Exception as ex:
+            logger.warning(
```
hectorhdzg, line 85: Looks like you used to return self.options.minimum_retry_interval for generic exceptions here, I suppose for networks error when endpoint is unavailable or similar and now is return non retriable

lzchen: As discussed, `RequestsException` handles everything regarding the `request` call.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -17,8 +17,11 @@
 
 import requests
 
-logger = logging.getLogger(__name__)
+from azure.core.exceptions import ClientAuthenticationError
+from azure.identity._exceptions import CredentialUnavailableError
 
+logger = logging.getLogger(__name__)
+_MONITOR_OAUTH_SCOPE = "https://monitor.azure.com/.default"
```
kryalama, line 24: This should be "https://monitor.azure.com//.default" ? "https://monitor.azure.com/.default" as per the example you have shared will generate a token with "aud": "https://monitor.azure.com", but the expected audience is "https://monitor.azure.com/" (as per Zaki's inputs).

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -48,6 +48,13 @@ def process_options(options):
         or 'https://dc.services.visualstudio.com'
     options.endpoint = endpoint + '/v2/track'
 
+    # Authorization
+    # `azure.core.credentials.TokenCredential` class must be valid
+    if options.credential and not hasattr(options.credential, 'get_token'):
```
ramthi, line 53: Checking to see if a method exists by name to check type seems odd.  You should be checking for type.  Consider using isinstance(..) https://docs.python.org/2.7/library/functions.html#isinstance

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -45,25 +48,44 @@ def _transmit(self, envelopes):
...
             return self.options.minimum_retry_interval
-        except Exception as ex:  # TODO: consider RequestException
+        except requests.RequestException as ex:
             logger.warning(
                 'Retrying due to transient client side error %s.', ex)
             # client side error (retryable)
             return self.options.minimum_retry_interval
+        except CredentialUnavailableError as ex:
+            logger.warning(
+                'Error with credential configuration %s', ex
```
ramthi, line 76: Not sure if the friendly message will always be in sync with the actual exception.  My suggestion is to simply say 'Credential error' and append the exception message.  What might be useful is to add 'dropping telemetry' coz that is something the ex message wont convey.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -45,25 +48,44 @@ def _transmit(self, envelopes):
...
+            )
+            return -1
+        except ClientAuthenticationError as ex:
+            logger.warning(
+                'Error getting token %s', ex
```
ramthi, line 81: Same here.  Just let the ex message convey the message.

ramthi (changes requested):

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
lzchen: Yes type checking is definitely more preferred. However, if you take a look at the actual implementations `TokenCredential` classes in Python Azure SDK, they actually implement them without an interface. There is no base class that is common amongst all types of `TokenCredential` s. This means that if ever a `TokenCredential` is implemented that does NOT implement `get_token` (since in Python, it is not enforced), our code will actually crash if we do not have the `hasattr` check. With `hasattr`, no matter how azure sdk implements the `TokenCredential`, as long as we have the function we need, we are okay.

Take a look  at [ManagedIdentityCredential](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/azure/identity/_credentials/managed_identity.py#L42) as an example. Notice how it only extends `object`.

Also, Python azure sdk only defined [TokenCredential](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/credentials.py#L16) class for type checking (linting). There is no enforcement of actual implementations at all.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: Refer to the [error handling](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/identity/Azure.Identity/README.md#error-handling) section of azure identity
s README. `CredentialUnavailableError` is only returned if the configuration is incorrect. This is the contract that azure identity has given for their apis, and they must respect it. I think the message is an appropriate blanket statement for all of the errors that can throw this exception.

The reason why I did not add "Dropping telemetry" is because we still have not decided whether or not we are going to do this (as a team). As noted above in the PR description, the exception handling is still a WIP.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: See above.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
ramthi: At best you are duplicating the message you are going to log.  If you are going to do anything addition to what the exception message already says then dont add extra stuff.  All it will take it is for Azure SDK to throw CredentialUnavailableError for another scenario and your message will be out of sync.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
ramthi: uh.  That is weird!  Oh well.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: Alright, will change the error message.

kryalama (approved):

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -120,6 +143,14 @@ def _transmit(self, envelopes):
             )
             # server side error (retryable)
             return self.options.minimum_retry_interval
+        # Authentication error
+        if response.status_code == 401:
```
hectorhdzg, line 147: Can you double check if 403 is possible as well?, I know QuickPulse return it in certain cases would be good to confirm with Breeze

hectorhdzg (approved):
LGTM

PR #1035, opened 2021-05-18
Django user data is optional

tibnor:
Collection of user identifiable data could give problems regarding GDPR.

I have therefore added an option so that the sending of django user id and user name is optional


cristina0botez:
Any updates on this one? It would be nice to have it for Django 2.2.

Issue #1043, opened 2021-06-28
Trace Integration of "logging" doesn't reuse existing tracer


se02035:
**Describe your environment.** 
- Python v 3.6/3.7

**Steps to reproduce.**
1. Create a custom tracer instance (with some correlation ids)
1. Invoke config_integration.trace_integrations(['logging']) 
1. use this logger to create a log entry 
1. the logger/log entry will have a different correlation id that then defined tracer -> **NOK**

**What is the expected behavior?**
The opencensus logging ext contains trace_integration(tracer=None). But the parameter "tracer" is not used. If it is set the execution context should be updated with that tracer. That ensures the internally used TraceLogger instance will use the correct correlation ids (defined by the tracer instance)

**What is the actual behavior?**
Specified tracer instance is ignored. therefore, some predefined correlation ids are used. Which obviously are wrong -> log entries can be correlated anymore. 

**Additional context.**
This is an actual problem when using the opencensus extension for Azure Functions. The Azure Function extension creates a custom tracer for e.g. tracking incoming requests & dependencies. The correlation information is required to correlated these calls. Since of the bug in the logging component e.g. error log entries can be correlated with the rest of the Azure Function log information.

lzchen:
I'm a bit confused as to the issue. When a `Tracer` is created, it is set as the current [tracer](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/tracer.py#L89) already so subsequent log records should have their records enriched with trace data.

se02035:
Sorry for the missing details (and creating confusion). I think at the end it comes down execution order of the code. I mean the trace integration (`config_integration.trace_integrations(['logging'])`) has to happen before the logger instance is created (which sometimes is not clear when a logger instance is created a the module level (which will be execute on import). 

Since the trace integration replaces the logging's logger class (TraceLogger) it makes more sense to be me now. I guess, that caused my confusion; a azure log entry is still created but with not trace information (since I didn't have the TraceLogger). 

I guess the only way to avoid that anyone else taps in to the same pitfall is updating the doc (or providing a very simple sample).

But since you asked this is a short sample (explaining the issue):

module globals.py
```python
def getLogger(
    name: str,
    instrumentation_conn_string: str = INSTRUMENTATION_KEY,
    propagate: bool = False,
) -> Logger:
    """Get a new logging instance with a handler to send logs to Application Insights

    Args:
        name([str]): [The name of the logger]
[...]

PR #1044, opened 2021-06-28
Se02035/fix/use existing tracer for logging

se02035:
# What we are trying to address

Fixes bug in the opencensus logging extension (bug: provided tracer instance will be ignored). If the provided tracer instance is ignored, then the logger won't use the tracer properties (e.g. correlation ids). As a result logs won't have the wrong correlation id. 

## Description of Changes

- Ensure that the execution context gets updated with the provided tracer instance (if there is one)
- Added an unit test to ensure that the execution context gets updated. 

## Additional Information
This PR is a fix for issue **https://github.com/census-instrumentation/opencensus-python/issues/1043**

### DIFF
--- a/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
+++ b/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
@@ -15,6 +15,7 @@
 import logging
 
 from opencensus.log import TraceLogger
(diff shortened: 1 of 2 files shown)


se02035:
@googlebot I signed it!

se02035:
Closed/abandoned PR because the proposed change doesn't resolve the original issue.

PR #1049, opened 2021-07-13
Add google cloud core dependency to opencensus-ext-google-cloud-clientlibs

aabmass:
This dependency was not explicit before because it should be there if
you are using a google cloud clientlib. However, if you install this
package without the `google-cloud-core` package, you will see import
errors.

`google-cloud-core` dropped support for 3.4 and 3.5, but kept support
for 2.7. The version specifiers reflect this.


lzchen (approved):
Should you add this to the changelog?

PR #1050, opened 2021-07-14
Drop python <3.6 in opencensus-ext-stackdriver

aabmass:
Drop support for Python versions < 3.6 (breaking change) for `opencensus-ext-stackdriver`. This is in preparation for upgrading to the `google-cloud-monitoring ~= 2.0`

- Update setup.py classifiers and python_requires
- Update tox.ini to only include this package in compatible python version builds


aabmass:
See #1056 instead

PR #1054, opened 2021-07-30
[Stats] Apply init + long logic to attach rate metrics

lzchen:
- Add long interval
- Export initial metrics on program start
- Do not flush stats metrics on close


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
```diff
@@ -125,7 +139,7 @@ def _get_attach_metric(self, metric):
         else:
             # Not in any rp or VM metadata failed
             properties.append(LabelValue(_RP_NAMES[3]))
-            properties.append(LabelValue(""))
+            properties.append(LabelValue(_RP_NAMES[3]))
```
hectorhdzg, line 142: This is kinda of weird, basically you are setting rp and rpid properties with "unknown" value, but this is tight to the order of the properties you appended in _get_attach_properties and the order of _RP_NAMES, are you following some standard or something?, this code looks easy to break in my perspective as it relies on the position of the string in the array and is quite hard to read as well.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
```diff
@@ -52,8 +53,17 @@ def _get_stats_short_export_interval():
         return _DEFAULT_STATS_SHORT_EXPORT_INTERVAL
 
 
+def _get_stats_long_export_interval():
+    ei_env = os.environ.get("APPLICATION_INSIGHTS_STATS_LONG_EXPORT_INTERVAL")
```
hectorhdzg, line 57: Having interval and CS as env variables could cause unexpected behaviors if configured incorrectly, do we really need to make these configurable?

lzchen: These were suggested by @heyams to be used for testing. We will not have this in our documentation for the customer to use.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
lzchen: Yes in my opinion it is quite easy to break as well, but we are tied to how OpenCensus designs metrics to be used. The label keys and properties are separate arrays that match up based off of index. In OpenTelemetry, this will be much more robust as labels will most likely be dictionaries but for now we have to do things like this.

hectorhdzg (approved):
LGTM

PR #1059, opened 2021-08-04
[Stats] Implement network stats - requests success count

lzchen:
Long and short stats are sent using the same exporter thread.
There is a counter that needs to hit the threshold before the long stats are exported.


contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
```diff
@@ -80,78 +85,140 @@ def _get_attach_properties():
...
-
-        return stats_metrics
-
-    def _get_attach_metric(self, metric):
+        metrics = []
+        try:
+            # Initial metrics use the long export interval
+            # Only export once long count hits threshold
+            self._long_threshold_count = self._long_threshold_count + 1
+            if self._long_threshold_count >= _STATS_LONG_INTERVAL_THRESHOLD:
```
heyams, line 151: how does this threshold scale when shorter and long intervals get changed later? can you use the explicit intervals? what about multi-threading env?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
```diff
@@ -33,9 +35,12 @@
 _DEFAULT_STATS_LONG_EXPORT_INTERVAL = 86400  # 24 hours
 
 _ATTACH_METRIC_NAME = "Attach"
+_REQ_SUC_COUNT_NAME = "Request Success Count"
```
heyams, line 38: what about the rest of the network count? failure, duration avg, throttle, exception and retry?

lzchen: I only added success count so the PR would be easier to read, as outlined in the description. Will be adding the rest in a followup PR.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
lzchen: What do you mean if the intervals get changed later? The intervals are defined once during start of the application and the threshold is calculated based off of those at that time.

> what about multi-threading env?

I'll add a locking mechanism to `_StatsbeatMetric`.

heyams: not really.. we might increate the attach internal.. or any other statsbeat metrics.  interval can be adjusted later based ont he data we collect.. it will offer some insights whether or not it's a good idea to update it.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
heyams: sure.

hectorhdzg (approved):
LGTM

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/statsbeat_metrics/statsbeat.py
lzchen: Right, for now there's only two different interval types. If more are added, more thresholds can be created. It's not worth to define separate exporter threads for each type of interval.

PR #1060, opened 2021-08-05
upgrade opencensus-ext-stackdriver to use google-cloud-monitoring v2 library

### LINKED ISSUE #1046: Upgrade opencensus-ext-stackdriver to monitoring_v3 2.* API
Before opening a feature request against this repo, consider whether the feature should/could be implemented in the [other OpenCensus client libraries](https://github.com/census-instrumentation). If so, please [open an issue on opencensus-specs](https://github.com/census-instrumentation/opencensus-specs/issues/new) first.

**Is your feature request related to a problem?**
If so, provide a concise description of the problem.

I can not upgrade `google.cloud.monitoring_v3` library to 2.0.0 or above in my environment because of breaking changes in its API from 1.* which `opencensus-ext-stackdriver` depends on.

**Describe the solution you'd like.**
What do you want to happen instead? What is the expected behavior?

I'd love it if `opencensus-ext-stackdriver` can be upgraded to using `google.cloud.monitoring_v3` 2.* APIs.

**Describe alternatives you've considered.**
Which alternative solutions or features have you considered?

Staying with `google.cloud.monitoring_v3` 1.1.0, but it's getting less and less attractive as it has [released](https://pypi.org/project/google-cloud-monitoring/#history) 4 new minor versions in the last 9 months since 2.0.0 came out.

**Additional context.**
Add any other context about the feature request here.

aabmass:
Fixes #1046

See migration guide [here](https://googleapis.dev/python/monitoring/2.4.2/UPGRADING.html), which mostly covered everything.

In addition, the `google-cloud-monitoring` v2 library uses [proto-plus](https://github.com/googleapis/proto-plus-python) instead of raw generated protoc objects. These have slightly different usage, e.g. no `.add()` on repeated fields and different timestamp behavior.

I manually tested the exporter example and things are working

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,6 @@ jobs:
         with:
           path: .tox
           # bump version prefix to fully reset caches
-          key: v2-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini') }}
+          key: v1-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini', '**/setup.py') }}
       - name: run tox
         run: tox -f ${{ matrix.python-version }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,6 @@ jobs:
         with:
           path: .tox
           # bump version prefix to fully reset caches
-          key: v1-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini') }}
+          key: v2-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini') }}
       - name: run tox
         run: tox -f ${{ matrix.python-version }}
--- a/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
+++ b/contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
@@ -852,10 +852,12 @@ class TestCreateTimeseries(unittest.TestCase):
         )
         self.assertEqual(value.distribution_value, expected_distb)
 
-        self.assertEqual(time_series.points[0].interval.start_time.seconds, 1545699663)
-        self.assertEqual(time_series.points[0].interval.start_time.nanos, 4053)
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -39,6 +39,6 @@ jobs:
         with:
           path: .tox
           # bump version prefix to fully reset caches
-          key: v1-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini') }}
+          key: v2-tox-${{ matrix.python-version }}-${{ hashFiles('tox.ini') }}
```
aabmass, line 42: The tox envs are incorrectly cached otherwise because the `hashFiles()` doesn't account for modified `setup.py` files.

jsuereth (approved):

.github/workflows/build.yml
punya: What breaks if you change to `hashFiles('tox.ini', '**/setup.py')` ?

punya (approved):

contrib/opencensus-ext-stackdriver/tests/test_stackdriver_stats.py
```diff
@@ -840,6 +852,13 @@ def test_create_timeseries(self, monitor_resource_mock):
         )
         self.assertEqual(value.distribution_value, expected_distb)
 
+        start_time_pb = time_series.points[0].interval.start_time.timestamp_pb()
+        end_time_pb = time_series.points[0].interval.end_time.timestamp_pb()
+        self.assertEqual(start_time_pb.seconds, 1545699663)
+        self.assertEqual(start_time_pb.nanos, 4053)
+        self.assertEqual(end_time_pb.seconds, 1545699723)
+        self.assertEqual(end_time_pb.nanos, 4053)
```
aabmass, line 860: Added this test in 68bda19c5e68429448fd2fbb9dadb7307ca87313 (before other changes in this PR) to test this works with new protoplus timestamps.

Issue #1052, opened 2021-07-19
Azure Exporter support providing cloud role via Options


se02035:
It is unclear how to set the cloud role value for a log record. According to https://github.com/census-instrumentation/opencensus-python/blob/5ad2a7cc62fc9663d186e15636a7b0306c9a8d7a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py#L27 the cloud is always `'ai.cloud.role': os.path.basename(sys.argv[0]) or 'Python Application'` 

A correct cloud role name is essential for Azure Application Insights Application Map. In order to use another role name (which is likely true for most of the cases) a custom callback has be used to update the role name. It would be better to have an parameter on the AzureExporter(options) (https://github.com/census-instrumentation/opencensus-python/blob/1ca601584cbbd3b6019bddb9c7120c069ff4f819/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py#L104) to set the cloud role (and instance) - similar to the instrumentation key/connection string,

lzchen:
You can modify this field via [telemetry processors](https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-filtering-sampling#opencensus-python-telemetry-processors)

se02035:
Yeah, I know but wouldn't it be nicer/cleaner to have a env var here? I mean I see a couple of benefits:
1. other AI SDKs also allow to set the cloud role name via an env var.
1. Similar approach is used for passing the AI instrumentation key
1. the telemetry processor has to be configured for every AzureExporter/AzureLogHandler instance. E..g you need to do something like this (when using Azure Functions): 
    - `OpenCensusExtension._exporter.add_telemetry_processor(..)` plus use the same for other AzureLogHandler (if available).

lzchen:
@se02035 

Those are excellent points and I agree with it. Feel free to open a PR and I will review it :)

se02035:
@lzchen , created a PR for allowing devs to set the cloud role name via an envvar. https://github.com/census-instrumentation/opencensus-python/pull/1061

PR #1061, opened 2021-08-11
Support setting cloud role name via envvar

se02035:
# What we are trying to address
The changes in this PR allows devs to configure the logged cloud role name (in AppInsights) using an environment variable (APPLICATIONINSIGHTS_ROLE_NAME) - if specified. (related issue: <https://github.com/census-instrumentation/opencensus-python/issues/1052>)

# Description of Changes
- Created a method to retrieve the azure monitor context (increased testability)

# Ran it using
- using existing TOX ini/config (for all python envs)

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
@@ -23,19 +23,35 @@ from opencensus.common.utils import timestamp_to_microseconds, to_iso_str
 from opencensus.common.version import __version__ as opencensus_version
 from opencensus.ext.azure.common.version import __version__ as ext_version
 
-azure_monitor_context = {
-    'ai.cloud.role': os.path.basename(sys.argv[0]) or 'Python Application',
-    'ai.cloud.roleInstance': platform.node(),
-    'ai.device.id': platform.node(),
-    'ai.device.locale': locale.getdefaultlocale()[0],
-    'ai.device.osVersion': platform.version(),
-    'ai.device.type': 'Other',
-    'ai.internal.sdkVersion': 'py{}:oc{}:ext{}'.format(
-        platform.python_version(),
-        opencensus_version,
-        ext_version,
-    ),
-}
+_azure_monitor_context = {}
+
+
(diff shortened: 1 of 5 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/common/utils.py
```diff
@@ -23,19 +23,35 @@
...
-        platform.python_version(),
-        opencensus_version,
-        ext_version,
-    ),
-}
+_azure_monitor_context = {}
+
+
+def get_azure_monitor_context(
+        role_name=(
```
lzchen, line 30: How come these are default kwargs? Shouldn't these just be assignments in the function?

se02035: @lzchen , that was part of the original code (default values). Since the azure_monitor_context is used in potentially many different places, I decided to keep the current default values. Makes sense?

lzchen: Yes keeping the default values makes sense. I meant since `get_azure_monitor_context()` is a function now, having kwargs implies that you can actually instantiate these values as different values (instead of just the constants), which actually shouldn't be allowed. If we are always calling it with no parameters, why even have the kwargs? Just assign them to default values in the actual function.

se02035: oh, yes. sure. the reason for providing default arguments, is because of testability. Having method arguments makes it easy to validate that the monitor context will contain the correct values (and we don't have to deal with mocking env variables, etc). What do you think.

But method args also provides some additional flexibility (if needed) .

lzchen: No we don't really want to have changed functionality for this. Users are not meant to change the values of the context.

molntamas:
Was this feature ever implemented in the end?

lzchen:
@se02035 
Are you still working on this or can somebody else pick this up?

lzchen:
Closing due to inactivity.

PR #1075, opened 2021-09-16
Remove google-api-core dependency from core opencensus package

aabmass:


aabmass:
Ah, so this won't work because of one actual use of protobuf library here: https://github.com/census-instrumentation/opencensus-python/blob/5ad2a7cc62fc9663d186e15636a7b0306c9a8d7a/opencensus/tags/propagation/binary_serializer.py#L21

PR #1077, opened 2021-09-22
Fix a few bugs for Stats

lzchen:
rpid -> rpId
Remove single "/" in rpId for vm
Change disableAll env to check for explicit 'True" string

Also fixed some constructor tests.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -64,7 +64,7 @@ class BaseLogHandler(logging.Handler):
         self._worker = Worker(self._queue, self)
         self._worker.start()
         # start statsbeat on exporter instantiation
-        if os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL") != "True":
+        if not os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"):
             statsbeat_metrics.collect_statsbeat_metrics(self.options)
 
     def _export(self, batch, event=None):  # pragma: NO COVER
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -64,7 +64,7 @@ class BaseLogHandler(logging.Handler):
         self._worker = Worker(self._queue, self)
         self._worker.start()
         # start statsbeat on exporter instantiation
-        if not os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"):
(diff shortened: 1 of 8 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, **options):
         self._worker = Worker(self._queue, self)
         self._worker.start()
         # start statsbeat on exporter instantiation
-        if not os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"):
+        if os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL") != "True":
```
hectorhdzg, line 67: APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=true will not work in this case, we usually just check for the env variable to exist and value is not that important, if you want to check for value ensure lower case and extra spaces doesn't mess with the check

lzchen: If it's normal practice to check if it exists then that is fine with me. Will leave it as a `None` check then.

hectorhdzg (approved):
LGTM

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
heyams: value matters though.. 

convert string to boolean and then check it.  Boolean.parseBoolean(value) (java way) handle case for you.

contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, **options):
         super(AzureExporter, self).__init__(**options)
         atexit.register(self._stop, self.options.grace_period)
         # start statsbeat on exporter instantiation
-        if not os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"):
+        if os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL") != "True":
```
heyams: check boolean value to support case sensitivity.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
lzchen: I'm fine either way, just trying to be consistent with node js sdk. @hectorhdzg  thoughts?

hectorhdzg: Like I mentioned node.js SDK does this kind of only checking for env to exist and not checking the value, we also mention this in docs when apply, is up to you how to handle it, casting works as well

lzchen: Will leave it as None.

PR #1084, opened 2021-10-12
change the way for logging integration

serjshevchenko:
Fix problem related to Note "Note that this only takes effect for loggers created after the integration."
Face with such issue when was trying to integrate opencensus-logging into uvicorn+FastAPI. 
uvicorn starts application in new thread/process and creates uvicorn logger object in new thread/process before opencensus-logging-integration execution. 
`TraceLogger` just overrides `makeRecord`, `logging.setLogRecordFactory` is a better approach because it has not drawback with time of logger object initialization.

### DIFF
--- a/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
+++ b/contrib/opencensus-ext-logging/opencensus/ext/logging/trace.py
@@ -14,17 +14,20 @@
 
 import logging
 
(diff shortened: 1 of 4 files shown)


serjshevchenko:
@aabmass @hectorhdzg @lzchen 
hello may I get review on this PR ? 
It is pending about 2 months.

PR #1100, opened 2022-02-11
Statsbeat changes

lzchen:
1. Don't send instrumentation/feature statsbeats on 0 values
2. Don't count throttles as retries as well
3. Shorten statsbeats host names


hectorhdzg (approved):
LGTM

PR #1106, opened 2022-03-02
Check disabled tracing URLs based on regex

jonasmiederer:
Currently the disabled tracing URLs are checked by testing whether [the path starts with one of the provided excludelist paths](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/utils.py#L45)

However if I want to exclude the root (/) from being traced, this is not possible. If an empty string is provided to `disable_tracing_url`, every path is disabled as every string starts with an empty string.

```python
'ui'.startswith('')
True
```

A check using regex would solve the problem and also make it more flexible.


tcihak-fqa:
Is this change going to be released?
I would like to exclude the root requests that an Azure App Service sends for "Always On".

PR #1112, opened 2022-03-15
Make sure handler.flush() doesn't deadlock.

gukoff:
Currently `flush()` deadlocks during process termination if there's any unsent messages in the queue.

This is because `atexit` first calls `handler.close()` and then `logging.shutdown()`, that in turn calls `handler.flush()` without arguments. I.e. `handler.close()` kills the worker, and then `handler.flush()` forever waits for the dead worker to send the messages from the queue.

Stacktrace dump by `py-spy` of the application in a deadlock:
```
Thread 8900 (idle): "MainThread"
    wait (threading.py:296)
    wait (threading.py:552)
    wait (opencensus\common\schedule\__init__.py:75)
    flush (opencensus\common\schedule\__init__.py:127)
    flush (opencensus\ext\azure\log_exporter\__init__.py:109)
    shutdown (logging\__init__.py:2036)
``` 

After this change, the deadlock is still possible if another thread concurrently closes the handler during the flush. However, this scenario is much less likely.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -110,14 +110,14 @@ class BaseLogHandler(logging.Handler):
             return
 
         # We must check the worker thread is alive, because otherwise flush
-        # is useless. Also, it would deadlock if not timeout is given, and the
+        # is useless. Also, it would deadlock if no timeout is given, and the
         # queue isn't empty.
         # This is a very possible scenario during process termination, when
         # atexit first calls handler.close() and then logging.shutdown(),
         # that in turn calls handler.flush() without arguments.
         if not self._worker.is_alive():
             logger.warning("Can't flush %s, worker thread is dead. "
-                           "Pending messages will be lost.", self)
+                           "Any pending messages will be lost.", self)
             return
 
         self._queue.flush(timeout=timeout)
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -106,6 +106,20 @@ class BaseLogHandler(logging.Handler):
         raise NotImplementedError  # pragma: NO COVER
 
     def flush(self, timeout=None):
+        if self._queue.is_empty():
+            return
+
+        # We must check the worker thread is alive, because otherwise flush
+        # is useless. Also, it would deadlock if not timeout is given, and the
+        # queue isn't empty.
+        # This is a very possible scenario during process termination, when
+        # atexit first calls handler.close() and then logging.shutdown(),
+        # that in turn calls handler.flush() without arguments.
+        if not self._worker.is_alive():
+            logger.warning("Can't flush %s, worker thread is dead. "
+                           "Pending messages will be lost.", self)
+            return
+
         self._queue.flush(timeout=timeout)
 
 
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -112,6 +112,9 @@ class Queue(object):
     def gets(self, count, timeout):
         return tuple(self._gets(count, timeout))
 
+    def is_empty(self):
+        return not self._queue.qsize()
+
     def flush(self, timeout=None):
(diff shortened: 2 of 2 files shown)


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -106,6 +106,20 @@ def log_record_to_envelope(self, record):
         raise NotImplementedError  # pragma: NO COVER
 
     def flush(self, timeout=None):
+        if self._queue.is_empty():
+            return
```
aabmass, line 110: It looks like Queue.flush() already checks for this condition

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -106,6 +106,20 @@ def log_record_to_envelope(self, record):
...
+
+        # We must check the worker thread is alive, because otherwise flush
+        # is useless. Also, it would deadlock if not timeout is given, and the
+        # queue isn't empty.
+        # This is a very possible scenario during process termination, when
+        # atexit first calls handler.close() and then logging.shutdown(),
+        # that in turn calls handler.flush() without arguments.
+        if not self._worker.is_alive():
+            logger.warning("Can't flush %s, worker thread is dead. ",
+                           "Pending messages will be lost.", self)
```
aabmass, line 120: probably meant to concatenate?

```suggestion
            logger.warning("Can't flush %s, worker thread is dead. "
                           "Pending messages will be lost.", self)
```

gukoff: Indeed, good catch

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
gukoff: It does. I added this check only to not `log.warning` about the pending messages in vain.

aabmass (approved):
LGTM, I'll let @lzchen merge this since it's azure related

gukoff:
@lzchen here:
https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L2201-L2203

lzchen:
@gukoff 
I see. This is a great find. It looks like the logic in the `BaseLogHandler` was done under the assumption that IT would be responsible for the shutdown logic. However it looks like the `logging` library itself has logic to do that, just without a timeout. Might be a bit unrelated to your change, but should we simply remove the `atexit` in the azurehandler and leave the responsibility to `logging`? `logging` is already calling `hander.close()` so actually in this case `close` would be called twice it seems like.

gukoff:
This would mean either: 
1. Lose `grace_period` on close or 
2. change the behavior of `close()` without arguments to wait only for `grace_period` instead of indefinitely. 

... because `shutdown()` calls `close()` without arguments, while the custom hook in the constructor calls it with `grace_period`.

I too thought about this option, but because it felt like a breaking change for `close(timeout=None)`, I chose a safer way.

Now when I think about it, we could make this change non-breaking with the [sentinel pattern](https://python-patterns.guide/python/sentinel-object/):
```python 
_sentinel = object()

...

def close(timeout=_sentinel):
  if timeout is _sentinel:  # no arguments passed -> close with the default grace_period
    timeout = self.options.grace_period
  ...
```
  

What do you think?

lzchen:
@gukoff 
I am fine with not making a change to the current `close()` for now. The guard against the empty queue should be sufficient for this use case.

lzchen (approved):

lzchen:
@gukoff 
Could you fix the build error so we can get this merged in? :)

gukoff:
> @gukoff Could you fix the build error so we can get this merged in? :)

If I'm reading [the CI log](https://github.com/census-instrumentation/opencensus-python/runs/5586793406?check_suite_focus=true#step:6:179) correctly, `py39-bandit` failed at the piece of code unrelated to this PR, which was fixed in [this commit](https://github.com/census-instrumentation/opencensus-python/commit/c609ef0fabe046a80c5adf0eb038de759fc6a6b5). 

Try rerunning CI checks? ;)

gukoff:
@lzchen would it be possible to release a new version with this fix in it?

lzchen:
@gukoff 
Will be releasing some time this week :)

PR #1113, opened 2022-03-17
Fix statsbeat bugs

lzchen:
1. Miscounted failures (must not include 439 and 402)
2. For 206, only count as retry if errored envelopes are retriable
3. Fix 429 to be retry instead of throttle
4. Count exception for 307 redirect if circular redirect
5. Count 402 and 439 as throttle


aabmass (approved):
LGTM

hectorhdzg (approved):
LGTM

PR #1114, opened 2022-03-20
Fix TypeError when local file storage is above max limit in opencensus-ext-azure

tephe:
This is a pretty ugly issue if we go over the limit of 50 MB local storage. (default limit)
The logging is broken and each time the max limit will be reached opencensus will go into some kind of infinite cycle because an exception will be thrown each time it tries to log the warning and so on. (we saw even 2 million TypeErrors in application insights in ~30 minutes)

Please take into consideration this issue because it's quite critical to azure-exporter usage.

Added test that reproduces it. Fix is to remove `format`
Full error stack
```
Unhandled exception from exporter.
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 138, in run
    dst._export(batch)
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 75, in _export
    envelopes = [self.log_record_to_envelope(x) for x in batch]
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 75, in <listcomp>
    envelopes = [self.log_record_to_envelope(x) for x in batch]
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 223, in log_record_to_envelope
    message=self.format(record),
  File "/usr/lib64/python3.8/logging/__init__.py", line 925, in format
    return fmt.format(record)
  File "/usr/lib64/python3.8/logging/__init__.py", line 664, in format
    record.message = record.getMessage()
  File "/usr/lib64/python3.8/logging/__init__.py", line 369, in getMessage
    msg = msg % self.args
TypeError: must be real number, not str 
```

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
@@ -199,7 +199,7 @@ class LocalFileStorage(object):
                             "reached. Currently at %fKB. Telemetry will be "
                             "lost. Please consider increasing the value of "
                             "'storage_max_size' in exporter config.",
-                            format(size/1024)
+                            size/1024
                         )
                         return False
         return True
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/storage.py
@@ -196,10 +196,10 @@ class LocalFileStorage(object):
                     if size >= self.max_size:
                         logger.warning(
                             "Persistent storage max capacity has been "
-                            "reached. Currently at %fKB. Telemetry will be "
+                            "reached. Currently at %sKB. Telemetry will be "
                             "lost. Please consider increasing the value of "
                             "'storage_max_size' in exporter config.",
-                            size/1024
+                            format(size/1024)
                         )
                         return False
         return True
--- a/contrib/opencensus-ext-azure/tests/test_storage.py
+++ b/contrib/opencensus-ext-azure/tests/test_storage.py
@@ -151,6 +151,16 @@ class TestLocalFileStorage(unittest.TestCase):
                     os_mock.return_value = True
                 self.assertTrue(stor._check_storage_size())
 
+    def test_check_storage_size_above_max_limit(self):
+        input = (1, 2, 3)
+        with LocalFileStorage(os.path.join(TEST_FOLDER, 'asd5'), 1) as stor:
+            with mock.patch('os.path.getsize') as os_mock:
+                os_mock.return_value = 52000000
+                stor.put(input)
+                with mock.patch('os.path.islink') as os_mock:
+                    os_mock.return_value = True
+                self.assertFalse(stor._check_storage_size())
+
     def test_maintenance_routine(self):
         with mock.patch('os.makedirs') as m:
             LocalFileStorage(os.path.join(TEST_FOLDER, 'baz'))


lzchen:
>   File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 223, in log_record_to_envelope
    message=self.format(record),

It looks like from the stacktrace that the error is coming from the log exporter itself when it tries to format the log record that comes in? How is this related to the error message in `storage.py`?

tephe:
> > File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/**init**.py", line 223, in log_record_to_envelope
> > message=self.format(record),
> 
> It looks like from the stacktrace that the error is coming from the log exporter itself when it tries to format the log record that comes in? Are you using a log formatter? 
Yes, I am using a log formatter. Could it be an issue here? The error above is reproduced very easily though.
> How is this related to the error message in `storage.py`?
We found this error right before a lot of TypeError, and I suspect it triggered the issue.

lzchen:
@tephe 
The error message looks like it is from the [handler.format()](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py#L225)  function called from the log exporter itself, not the `format()` in `storage.py`. Possibly the log formatter is not playing well with the `hander.format()`. Could you paste a snippet of how you are using the `AzureLogHandler` as well as your log formatter?

tephe:
> @tephe The error message looks like it is from the [handler.format()](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py#L225) function called from the log exporter itself, not the `format()` in `storage.py`. Possibly the log formatter is not playing well with the `hander.format()`. Could you paste a snippet of how you are using the `AzureLogHandler` as well as your log formatter?

Sure, I'm adding the AzureLogHandler to the root logger. 
```
    log_handler = AzureLogHandler(connection_string=f"InstrumentationKey={instrumentation_key}")
    log_handler.add_telemetry_processor(_on_send_telemetry)  # on send telemetry we just remove healthchecks 
    log_handler.setLevel(logging.INFO)
    logging.getLogger().addHandler(log_handler)
```

Also I checked we don't use any log formatter in this project, it's the default one. 

My hunch is that due to the exception above opencensus goes into an infinite loop. We are also seeing an infinite loop with `Queue is full. Dropping telemetry`
 
```
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 101, in emit
    self._queue.put(record, block=False)
  File "/usr/local/lib/python3.8/site-packages/opencensus/common/schedule/__init__.py", line 134, in put
    logger.warning('Queue is full. Dropping telemetry.')
  File "/usr/lib64/python3.8/logging/__init__.py", line 1454, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/lib64/python3.8/logging/__init__.py", line 1585, in _log
    self.handle(record)
  File "/usr/lib64/python3.8/logging/__init__.py", line 1595, in handle
    self.callHandlers(record)
  File "/usr/lib64/python3.8/logging/__init__.py", line 1657, in callHandlers
    hdlr.handle(record)
  File "/usr/lib64/python3.8/logging/__init__.py", line 950, in handle
    self.emit(record)
  File "/usr/local/lib/python3.8/site-packages/opencensus/ext/azure/log_exporter/__init__.py", line 101, in emit
    self._queue.put(record, block=False)
  File "/usr/local/lib/python3.8/site-packages/opencensus/common/schedule/__init__.py", line 134, in put
    logger.warning('Queue is full. Dropping telemetry.')
  File "/usr/lib64/python3.8/logging/__init__.py", line 1454, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/lib64/python3.8/logging/__init__.py", line 1583, in _log
    record = self.makeRecord(self.name, level, fn, lno, msg, args,
  File "/usr/lib64/python3.8/logging/__init__.py", line 1552, in makeRecord
    rv = _logRecordFactory(name, level, fn, lno, msg, args, exc_info, func,
  File "/usr/lib64/python3.8/logging/__init__.py", line 314, in __init__
    self.levelname = getLevelName(level)
  File "/usr/lib64/python3.8/logging/__init__.py", line 134, in getLevelName
    result = _levelToName.get(level)
RecursionError: maximum recursion depth exceeded while calling a Python object

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.8/site-packages/opencensus/common/schedule/__init__.py", line 132, in put
    self._queue.put(item, block, timeout)
  File "/usr/lib64/python3.8/queue.py", line 136, in put
    raise Full
queue.Full

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.8/site-packages/opencensus/common/schedule/__init__.py", line 132, in put
    self._queue.put(item, block, timeout)
  File "/usr/lib64/python3.8/queue.py", line 136, in put
    raise Full
queue.Full

During handling of the above exception, another exception occurred:
```

lzchen:
@tephe 
Thanks for the code snippet. If the queue is full, it seems that the AzureLogHandler is picking up logs from ITSELF (due to some error) and adding that to the queue, thus resulting in an infinite loop. This is because you have the handler attached to the ROOT handler. If you want a quick fix to this, try attaching the handler to a namespaced logger.

```
logger = logging.getLogger("current_logger").add_handler(AzureLogHandler())
logger.warning("Hello World!")
```
This should prevent the infinite loop. However, I am curious as to what is generating the original `format` error.

tephe:
>This should prevent the infinite loop. However, I am curious as to what is generating the original format error.

It's pretty hard to generate this exact same message for `TypeError: must be real number, not str`
This means that we are sending a string and the format is expecting a real number, which is the case for the warning that i fixed here `logger.warning(`Currently at %fKB.`, format(size/1024))`

So I believe there's a very high probability the original TypeError is from here.

Should I fix this issue in a different way? Perhaps use an f string for this warning.

tephe:
>However, did https://github.com/census-instrumentation/opencensus-python/pull/1114#issuecomment-1075472379 address the ORIGINAL issue of infinite looping?
I did not try that. We want to attach the log handler to the root logger, so we send all logs to appinsights.

I did disable local storage with `enable_local_storage=False` option plus doubled the queue size and I'm not seeing the TypeError issue anymore, nor the infinite loop when queue is full.

tephe:
> I believe we use format explicitly because f strings do not exist for Python 2.7. We can also try removing the %f and simply format it as a string?

I did that. changed from %f to %s. Verified that it also works on python 2.7. Btw the TypeError is also thrown on python 2.7 but the message is a bit different.
```
TypeError: float argument required, not str
Logged from file storage.py, line 202
```

lzchen (approved):

PR #1116, opened 2022-04-13
Fix statsbeat bugs

lzchen:
1. Do not log when sending statsbeat data
2. Set `is_exporter` to `True` when calling `azure_meta_data_service` so `requests` instrumentation will not track `requests` call as regular telemetry


hectorhdzg (approved):
LGTM

PR #1118, opened 2022-04-18
Add deprecation message for using instrumentation key

lzchen:


contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py
```diff
@@ -30,6 +33,13 @@ def process_options(options):
         os.getenv('APPLICATIONINSIGHTS_CONNECTION_STRING'))
     env_ikey = os.getenv('APPINSIGHTS_INSTRUMENTATIONKEY')
 
+    # Deprecation note about explicit instrumentation key usage
```
hectorhdzg, line 36: I believe we wanted to show this warning only when Connection String is not available and iKey is the only one present, this to avoid noise, can you confirm on current java logic?

lzchen: Confirmed this is how it should work. Making the change so warning only gets fired if connection string is not present.

hectorhdzg (approved):

PR #1127, opened 2022-06-07
Shutdown statsbeat after failure threshold is met

lzchen:
Following [specs](https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/ApplicationInsights/statsbeat.md#how-to-handle-statsbeat-failures)

Similar to Node js, retry only occurs on successes (200) so shutdown occurs only when 3 attempts are reached.

@hectorhdzg @heyams


contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -128,7 +135,12 @@ def _transmit(self, envelopes):
                             _requests_map['retry'] = _requests_map.get('retry', 0) + 1  # noqa: E501
                         else:
                             _requests_map['exception'] = _requests_map.get('exception', 0) + 1  # noqa: E501
-
+                if self._is_stats_exporter() and \
+                    not state.get_statsbeat_shutdown() and \
+                        not state.get_statsbeat_initial_success():
+                    # If ingestion threshold during statsbeat initialization is reached, return back code to shut it down
+                    if _statsbeat_failed_to_ingest():
+                        return -2
```
hectorhdzg, line 143: It would be nice to use some kind of constant instead of the value here, -2 is shutdown signal, but looking at the code here I have no clue what -1 means.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -143,6 +155,17 @@ def _transmit(self, envelopes):
                 data = json.loads(text)
             except Exception:
                 pass
+
+        if self._is_stats_exporter() and \
+            not state.get_statsbeat_shutdown() and \
```
hectorhdzg, line 160: I can see you check if shutdown was called in several places, is the exporter process for Statsbeat expected to keep running after shutdown?

heyams:
@lzchen you're tagging the wrong helen.

contrib/opencensus-ext-azure/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Statsbeat - shutdown stats when hit error/exception threshold
```
heyams, line 5:
```suggestion
- Shutdown Statsbeat when hitting error/exception threshold
```

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -271,3 +297,13 @@ def _transmit(self, envelopes):
                 with _requests_lock:
                     _requests_map['throttle'] = _requests_map.get('throttle', 0) + 1  # noqa: E501
         return -response.status_code
+
+
+def _reached_ingestion_status_code(status_code):
+    return status_code in _REACHED_INGESTION_STATUS_CODES
+
+
+def _statsbeat_failed_to_ingest():
```
heyams, line 306: maybe rename something like statsbeat_failure_reached_threshold?

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -71,6 +71,11 @@ def export_metrics(self, metrics):
         for batch in batched_envelopes:
             batch = self.apply_telemetry_processors(batch)
             result = self._transmit(batch)
+            # If statsbeat exporter and received signal to shutdown
+            if self._is_stats_exporter() and result == -2:
```
heyams, line 75: _statsbeat_failed_to_ingest above can the counter and here just check if the counter is >= 3. -2 seems so random.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: -1 is the exception signal for telemetry in general. -2 is the shutdown signal for only statsbeat exporter.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
lzchen: `_statsbeat_failed_to_ingest ` is a private function used only within `transport` to handle the count as well as determining whether it is reached. The returning of the result code back to the exporter is by design. I agree the codes are a bit random (-1, -2) but changing them can be part of a different PR.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
lzchen: This is for very specific race conditions in which multiple threads could be accessing the same piece of "check if we need to shutdown" logic. It also serves as a good sanity check to prevent from any statsbeat logic from executing if the statsbeat exporte ris already shutdown.

contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
hectorhdzg: Well using these kinds of numbers instead of enumerators or constants is usually a pretty bad practice in other languages, code is harder to understand and maintain by other developers, maybe this is the way to go in Python, just my two cents here.

heyams: is -2 introduce in this PR? if so, how much work does it take to refactor?

PR #1135, opened 2022-06-29
Add kwargs to derived gauge

lzchen:


aabmass (commented):
What does this PR do? I'm not very familiar with this code. It looks like it's unconditionally adding kwargs as labels for the DerivedGauge?

lzchen:
@aabmass 
DerivedGauges generates point value based off of a passed in [function](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/metrics/export/gauge.py#L219). The current design only accepts a function with no args. This PR adds the option to pass in `kwargs` to the DerivedGauge to execute the function with parameters.

aabmass (approved):
Got it, LGTM.

PR #1138, opened 2022-07-05
Implement exceptionType and statusCode for network statsbeat

lzchen:
From [specs](https://github.com/microsoft/Telemetry-Collection-Spec/pull/61)
Also fixes counting exceptions thrown by http library as `exceptions` rather than `retry`


hectorhdzg (approved):
LGTM

Issue #1139, opened 2022-07-06
Remove upper version limit for sqlalchemy in `opencensus-ext-sqlalchemy`


paul-ciobanu:
**Describe your environment.**
Remove upper version limit for sqlalchemy in `opencensus-ext-sqlalchemy`
```
        'SQLAlchemy >= 1.1.14, < 1.3.24',  # https://github.com/sqlalchemy/sqlalchemy/issues/6168 # noqa: E501
```
since [https://github.com/sqlalchemy/sqlalchemy/issues/6168](https://github.com/sqlalchemy/sqlalchemy/issues/6168) has been fixed
this is preventing me to use the latest `SQLAlchemy` version `1.4.39` in my project

PR #1142, opened 2022-07-12
remove sqlalchemy upper limit version

lwatskot:
https://github.com/census-instrumentation/opencensus-python/issues/1139

### DIFF
--- a/contrib/opencensus-ext-sqlalchemy/setup.py
+++ b/contrib/opencensus-ext-sqlalchemy/setup.py
@@ -42,7 +42,7 @@ setup(
     long_description=open('README.rst').read(),
     install_requires=[
         'opencensus >= 0.9.dev0, < 1.0.0',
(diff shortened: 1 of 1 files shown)


javier-martinez-codurance:
@hectorhdzg Hi, brief question do you know if the upper limit was there for a reason? I would like to add the extension but due to this constraint I can not do it properly. Thanks in advance!

lwatskot:
@hectorhdzg @lzchen can you please take a look at this?

aberres:
Getting this merged would be great!

panteamihai:
@hectorhdzg @lzchen @aabmass could you guys please take a look at this?

aabmass (approved):
Thanks for the PR and apologies for the slow response.

It looks like SQLAlchemy doesn't support python 3.4 anymore at all. https://github.com/sqlalchemy/sqlalchemy/blob/main/setup.cfg#L34. I think this change is fine as it is the user's responsiblity to pin the correct version of SQLAlchemy for their python interpreter version.

lzchen (approved):

PR #1143, opened 2022-07-29
Include version.py in distribution for certain packages

lzchen:
Applies to:

opencensus-context
opencensus-ext-flask
opencensus-ext-requests

specifically requested by a user.


contrib/opencensus-ext-requests/setup.py
```diff
@@ -12,13 +12,21 @@
...
 from setuptools import find_packages, setup
 
-from version import __version__
+BASE_DIR = os.path.dirname(__file__)
+VERSION_FILENAME = os.path.join(
+    BASE_DIR, "opencensus", "ext", "requests", "common", "version.py"
+)
+PACKAGE_INFO = {}
+with open(VERSION_FILENAME) as f:
+    exec(f.read(), PACKAGE_INFO)
```
aabmass, line 25: Why do the other packages in this PR import the `version` module directly but this one reads the file and `exec`s it? I think we do the latter in OTel, but not sure of the difference between the two.

aabmass (commented):
I think `context/opencensus-context/opencensus/common/version.py may conflict with this package https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/common/version/__init__.py?

Wouldn't they existin the same namespace package?

contrib/opencensus-ext-requests/setup.py
lzchen: Great question. `opencensus-ext-request` imports the [trace](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-requests/opencensus/ext/requests/__init__.py#L15) module in one of it's `__init__.py`, in which it imports [requests](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-requests/opencensus/ext/requests/trace.py#L17). So if we were to import the `version` module directly in `setup.py`, the imports will attempt to resolve first before `requests` can be installed which will result in `requests` not being able to be found. For consistency, we can make all of them read from the file. Not a strong opinion.

aabmass: That would be great if they can all jsut read the file, thanks for explaining

lzchen:
@aabmass 

> I think `context/opencensus-context/opencensus/common/version.py may conflict with this package https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/common/version/__init__.py?

I moved `context/opencensus-context/opencensus/common/version.py` to `context/opencensus-context/opencensus/common/runtime_context/version.py`. This way it would be more clear that the `version.py` belongs to the context package.

aabmass (approved):

PR #1150, opened 2022-08-09
Add storage existence checks

lzchen:


contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -79,17 +79,19 @@ def _export(self, batch, event=None):  # pragma: NO COVER
...
-                    self.storage.put(
-                        envelopes,
-                        self.options.minimum_retry_interval,
-                    )
-            if event:
-                if isinstance(event, QueueExitEvent):
-                    self._transmit_from_storage()  # send files before exit
-                return
-            if len(batch) < self.options.max_batch_size:
-                self._transmit_from_storage()
```
jeremydvoss, line 92: Could you explain the removal of these 2 if statements? Given that they deal with storage, but were outside the original self.storage if statement, I'm not sure what their exact purpose was.

contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/__init__.py
```diff
@@ -81,13 +81,14 @@ def export_metrics(self, metrics):
                 statsbeat.shutdown_statsbeat_metrics()
                 return
             # Only store files if local storage enabled
-            if self.storage and result is TransportStatusCode.RETRY:
-                self.storage.put(batch, self.options.minimum_retry_interval)
-
-        # If there is still room to transmit envelopes, transmit from storage
-        # if available
-        if len(envelopes) < self.options.max_batch_size:
-            self._transmit_from_storage()
```
jeremydvoss, line 90: Same here

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
lzchen: If you look at the git diff, the if statements are not being removed, they are just being placed inside `self.storage` check. We don't want to be storing into storage or trying to transmit from storage if `self.storage` doesn't exist.

jeremydvoss (approved):

Issue #928, opened 2020-07-14
AzureExporter not working with multiprocessing


dpgrodriguez:
**Describe your environment.** 
MacOS 10.14.6
Python 3.7.5
opencensus-ext-azure==1.0.4
opencensus-ext-requests==0.7.3

**Steps to reproduce.**
I have code that I want to monitor the dependency calls to Azure DevOps APIs. Our code is running multiprocessing using the `Process` class. When the exporter is ran outside of multiprocessing, it sends telemetry to App Insights. When ran inside a multiprocessing Process, it doesn't. I added a callback to print the spandata and it doesn't get called when using `Process`.

```python
from azure.devops.connection import Connection
from msrest.authentication import BasicAuthentication

from multiprocessing import Process, Pool, Queue
from base_insights import BaseInsights
from opencensus.common.runtime_context import RuntimeContext


class TestInsights:
    def __init__(self):
        self.tracer = BaseInsights.tracer

    def process(self):
       
        procs = []
        organization_url = 'https://dev.azure.com/org'
        credentials = BasicAuthentication('', '')
        
        p1 = Process(target=self.my_loop, args=[organization_url, credentials])
        p1.start()
        p1.join()

    def my_loop(self, organization_url, credentials, parent_span=None):
      
        with self.tracer.span(name='TestLoopProcessThreadingInside'):

            connection = Connection(base_url=organization_url, creds=credentials)
            core_client = connection.clients.get_core_client()

            org = core_client.get_project_collection("test")


TestInsights().process()
```

BaseInsights:
```python
import os

from opencensus.ext.azure.trace_exporter import AzureExporter
from opencensus.trace import file_exporter

from opencensus.trace import config_integration
from opencensus.trace.samplers import ProbabilitySampler, AlwaysOnSampler
from opencensus.trace.tracer import Tracer

config_integration.trace_integrations(['requests'])


def singleton(cls):
    return cls()


@singleton
class BaseInsights:
    def __init__(self):
        exporter = AzureExporter()
        exporter.add_telemetry_processor(self.callback_function)

        self.tracer = Tracer(exporter=exporter, sampler=AlwaysOnSampler())


    def callback_function(self, envelope):
        print(envelope)
```
**What is the expected behavior?**
Span data gets sent to Application Insights

**What is the actual behavior?**
Span data is not sent to Application Insights

lzchen:
@dpgrodriguez 
[...]

Issue #900, opened 2020-06-05
AzureLogHandler does not seem to work from Celery tasks


Bastien-Brd:
```
Python 3.5.2
Django 2.2.10
celery==4.2.2
opencensus==0.7.7
opencensus-context==0.1.1
opencensus-ext-azure==1.0.2
opencensus-ext-django==0.7.2
opencensus-ext-logging==0.1.0
```

After a loooong investigation and many experiments with various ways of configuring logging in Celery tasks and django, I am pretty sure that **for some reason, the AzureLogHandler does not send any log to Azure from a task running in a Celery worker**.

To make sure of that, in my django project, I setup Celery logging by discarding the default Celery logging behaviour (_the default is misleading because it overrides the root logger_) and instead I use the django logging configuration (from django's `settings.py`), and in addition I manually make sure the `celery` logger has both a local file log handler (making sure I capture all log messages locally for debugging this issue) and a AzureLogHandler. You can see how this is configuring in the code snippet below:

```python
import os
import logging

from celery import Celery
from celery.signals import setup_logging
from opencensus.ext.azure.log_exporter import AzureLogHandler

from django.conf import settings


os.environ.setdefault("DJANGO_SETTINGS_MODULE", "edgefolio.settings")

# Create the Celery app (called "api")
app = Celery('api')
app.config_from_object('django.conf:settings', namespace='CELERY')
app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)


def add_azure_log_handler_to_logger(logger, propagate=True):
    """
    Given a logger, add a AzureLogHandler to it
    :param logger:
    :param propagate:
    :return:
    """
    if settings.AZURE_APP_INSIGHTS_INSTR_KEY:
        formatter = logging.Formatter("[Celery/%(processName)s] %(message)s")
        # Azure Handler:
        azure_log_handler = AzureLogHandler(
            connection_string='InstrumentationKey=%s' % settings.AZURE_APP_INSIGHTS_INSTR_KEY)
        azure_log_handler.setFormatter(formatter)
        azure_log_handler.setLevel(logging.INFO)
        logger.addHandler(azure_log_handler)
        logger.setLevel(logging.INFO)
        logger.propagate = propagate
    return logger


@setup_logging.connect
def setup_loggers(*args, **kwargs):
    """
    Using the celery "setup_logging" signal to override and fully define the logging configuration for Celery
    :param args:
    :param kwargs:
    :return:
    """
[...]

PR #1158, opened 2022-08-23
Fixed AzureLogHandler with multiple processes.

JeremyVriens:
These changes fix both https://github.com/census-instrumentation/opencensus-python/issues/900 and https://github.com/census-instrumentation/opencensus-python/issues/928, but I'm not so sure about the compatibility with Python2.
I changed the normal queue.Queue to a multiprocessing.Queue and converted the LogRecord to an envelope before putting it on the queue as serializing a LogRecord didn't work when it contained a traceback.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -75,8 +75,7 @@ class BaseLogHandler(logging.Handler):
     def _export(self, batch, event=None):  # pragma: NO COVER
         try:
             if batch:
-                envelopes = [self.log_record_to_envelope(x) for x in batch]
-                envelopes = self.apply_telemetry_processors(envelopes)
+                envelopes = self.apply_telemetry_processors(batch)
                 result = self._transmit(envelopes)
                 # Only store files if local storage enabled
                 if self.storage:
@@ -110,7 +109,11 @@ class BaseLogHandler(logging.Handler):
         self.lock = None
 
     def emit(self, record):
-        self._queue.put(record, block=False)
+        # Convert the raw LogRecord to an envelope before putting it on the
+        # queue as a LogRecord object is not serializable, while an Envelope
+        # object is.
+        envelope = self.log_record_to_envelope(record)
+        self._queue.put(envelope, block=False)
 
     def log_record_to_envelope(self, record):
         raise NotImplementedError  # pragma: NO COVER
--- a/opencensus/common/schedule/__init__.py
+++ b/opencensus/common/schedule/__init__.py
@@ -15,6 +15,7 @@
 from six.moves import queue
 
 import logging
+import multiprocessing
 import threading
 import time
 
@@ -82,7 +83,7 @@ class QueueExitEvent(QueueEvent):
 class Queue(object):
     def __init__(self, capacity):
         self.EXIT_EVENT = QueueExitEvent('EXIT')
-        self._queue = queue.Queue(maxsize=capacity)
+        self._queue = multiprocessing.Queue(maxsize=capacity)
 
     def _gets(self, count, timeout):
         start_time = time.time()


lzchen:
@JeremyVriens 
Thanks for the contribution! Please sign the CLA when you have the time.

opencensus/common/schedule/__init__.py
```diff
@@ -82,7 +83,7 @@ class QueueExitEvent(QueueEvent):
 class Queue(object):
     def __init__(self, capacity):
         self.EXIT_EVENT = QueueExitEvent('EXIT')
-        self._queue = queue.Queue(maxsize=capacity)
+        self._queue = multiprocessing.Queue(maxsize=capacity)
```
lzchen, line 86: `queue.Queue` was created to work in concurrent environments spawned with the `threading` module, which is the behavior of the Azure exporters. With that being said, we probably shouldn't replace the default queue that is being used, since the original use case is supposed to be for the concurrent environments. I suggest adding the ability for the user to configure which type of queue they want by adding an option [here](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/common/__init__.py#L119). Then you can create a different type queue accordingly.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -110,7 +109,11 @@ def createLock(self):
         self.lock = None
 
     def emit(self, record):
-        self._queue.put(record, block=False)
+        # Convert the raw LogRecord to an envelope before putting it on the
+        # queue as a LogRecord object is not serializable, while an Envelope
+        # object is.
+        envelope = self.log_record_to_envelope(record)
+        self._queue.put(envelope, block=False)
```
lzchen, line 116: Why does object need to be serializable?

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -75,8 +75,7 @@ def __init__(self, **options):
     def _export(self, batch, event=None):  # pragma: NO COVER
```
lzchen, line 75: Please add an entry to the CHANGELOG

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -75,8 +75,7 @@ def __init__(self, **options):
...
         try:
             if batch:
-                envelopes = [self.log_record_to_envelope(x) for x in batch]
-                envelopes = self.apply_telemetry_processors(envelopes)
+                envelopes = self.apply_telemetry_processors(batch)
```
lzchen, line 78: This doesn't work because `apply_telemetry_processors` is expecting an `Envelope` data type.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
```diff
@@ -110,7 +109,11 @@ def createLock(self):
         self.lock = None
 
     def emit(self, record):
-        self._queue.put(record, block=False)
+        # Convert the raw LogRecord to an envelope before putting it on the
+        # queue as a LogRecord object is not serializable, while an Envelope
+        # object is.
+        envelope = self.log_record_to_envelope(record)
```
lzchen, line 115: I think there are performance implications of doing this. We don't want to be mapping every time `logger.X()` is called.

JeremyVriens:
> @JeremyVriens
> Thanks for the contribution! Please sign the CLA when you have the time.

I did, but I did a commit with the wrong git email config. Will try to fix this asap when I got time.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
JeremyVriens: It does work; the batch is a list of `Envelope` data type objects as I converted them already before putting them on the queue.

contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
JeremyVriens: There is indeed a performance loss here. However it is not possible to put a raw `LogRecord` on the multiprocessing.Queue as it pickles the object.

opencensus/common/schedule/__init__.py
JeremyVriens: It does work with threading, however in an environment where multiple processes are being used (see linked bug reports), the queue is not being shared between the different processes. Therefore a multiprocessing.Queue is required to get this to work.

Korred:
@JeremyVriens thanks for your POC!  I just came across issue #900 a few days ago, while trying to pass logs from Celery to Azure Application Insights and seeing that this was first reported 2 years ago I actually didn't expect this to be picked up and "fixed" in the near future. Getting this properly fixed would be really helpful for a lot of people with this kind of setup!

burkol:
Hi All,

I faced the same issue.
Is there a possibility that it will be accepted in the near future?

Regards,
Zoltan

vvxhid:
May I ask for an estimated merge date for this pull request? Thanks!

PR #1160, opened 2022-08-25
Refactor update requests map logic

lzchen:


contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
```diff
@@ -332,21 +338,18 @@ def _statsbeat_failure_reached_threshold():
...
-            _requests_map['duration'] = _requests_map.get('duration', 0) + value  # noqa: E501
-        elif type == "success":
-            _requests_map['success'] = _requests_map.get('success', 0) + 1  # noqa: E501
-        else:
-            # value will be a key (status_code/error message)
+        if value is None:  # success, count
+            _requests_map[type_name] = _requests_map.get(type_name, 0) + 1
+        elif type_name == "duration":  # value will be duration
+            _requests_map[type_name] = _requests_map.get(type_name, 0) + value  # noqa: E501
+        else:  # exception, failure, retry, throttle
```
jeremydvoss, line 348: Duration is the only typename that will count as a success? Or at least the only such typename where we will pass a value?

jeremydvoss (approved):

PR #1186, opened 2023-01-24
Enhance httpx tracing

macieyng:
In httpx `Advanced Usage` docs there is a [chapter `Request Instances`](https://www.python-httpx.org/advanced/#request-instances) where they show alternative way of sending requests, where you can use once initialized client to send various of different requests to the same host. 

Current implementation of tracing doesn't cover that use case, because calling structure is different while function `Client.send` is actually what is common and it returns `Response` object. 

In a nutshell:
```python 
class Client:
    ...
    def request(...) -> Response:
        ...
        request = self.build_request(...)
        return self.send(request)
```
so when I have something like this:
```python
req_1 = httpx.Request(...)
req_2 = httpx.Request(...)

with httpx.Client() as client:
    client.send(req_1)
    client.send(req_2)
```
It's not traced.

This is a use case that I have in my project and I would love to have that requests traced.

### DIFF
--- a/contrib/opencensus-ext-httpx/opencensus/ext/httpx/trace.py
+++ b/contrib/opencensus-ext-httpx/opencensus/ext/httpx/trace.py
@@ -41,7 +41,7 @@ def trace_integration(tracer=None):
         execution_context.set_opencensus_tracer(tracer)
 
     wrapt.wrap_function_wrapper(
-        MODULE_NAME, "Client.request", wrap_client_request
+        MODULE_NAME, "Client.send", wrap_client_request
     )
     # pylint: disable=protected-access
     integrations.add_integration(integrations._Integrations.HTTPX)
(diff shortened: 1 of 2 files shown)


miqm:
@macieyng bump the version of the extension

macieyng:
Thanks @lzchen for running the pipeline. I will fix it tomorrow.

macieyng:
@lzchen ready for running the pipeline 🚀
My bad for not running tox locally before pushing. It turns out that implementation has to be a little bit different from `requests`. 
@inirudebwoy can you review this too, please?

macieyng:
@lzchen can you run the pipeline? Thanks! 🙏

macieyng:
@aabmass @jeremydvoss @hectorhdzg @calleo 🙏

macieyng:
```
Failed to import opencensus.ext.zipkin.trace_exporter.
Possible hints:
* AttributeError: module 'opencensus' has no attribute 'ext'
* ModuleNotFoundError: No module named 'opencensus.ext'
Failed to import opencensus.ext.stackdriver.trace_exporter.
Possible hints:
* AttributeError: module 'opencensus' has no attribute 'ext'
* ModuleNotFoundError: No module named 'opencensus.ext'
```
That's not something that I touched and on my machine it fails even earlier, but's because I'm running it locally on mac. @lzchen can I get your support with sorting this out?

PR #1187, opened 2023-02-06
When batching, same object is mutated destroying exception information

calleo:
While testing Azure Exporter to see if exceptions (and traces) where propagated correctly to Application Insights I noticed that stack traces ended upp in a custom field called "stacktrace" but they where not tagged as exceptions in Application Insights.

Each time an exception occurs within a span, the function `span_data_to_envelope` will yield twice. The problem is that the same envelope object is re-used and modified. The current tests won't fail because each `yield` is tested in isolation, given the nature of `next`.

In my setup, `emit` is always called with a batch of span's, meaning all of them are processed before sent to Application Insights. As a result I end up with two envelopes of type `Microsoft.ApplicationInsights.Request`. I would expect one of type `Microsoft.ApplicationInsights.Exception` and one of `Microsoft.ApplicationInsights.Request`.


.python-version
```diff
@@ -0,0 +1 @@
+3.6
```
lzchen, line 1: What's this for?

lzchen:
Good catch, please fix the build errors and then we can merge this.

.python-version
calleo: My bad, will remove 👍

calleo:
> Good catch, please fix the build errors and then we can merge this.

Thanks @lzchen. Updated the way the object is copied, so hopefully works now. Had some unrelated tests fail locally, but I believe something is not properly setup in my environment.

lzchen:
@calleo 
Please rebase from latest master to get the builds running :)

lzchen (approved):
@calleo 
Would you kindly add a CHANGELOG entry?

calleo:
> @calleo Would you kindly add a CHANGELOG entry?

Thanks' for reviewing. I've added this fix to the changelog now.

PR #1196, opened 2023-03-13
Adds str fallback to envelope serialization

ddeschepper:
When serialization fails (e.g. when a non-JSON-serializable value is added to the properties) the entire batch that envelope belongs to does not get sent. Falling back to to str as a means of serialization when JSON serialization fails seems like a sane default that greatly reduces the risk of missed telemetry.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
@@ -113,7 +113,7 @@ class TransportMixin(object):
             endpoint += '/v2.1/track'
             response = requests.post(
                 url=endpoint,
-                data=json.dumps(envelopes),
+                data=json.dumps(envelopes, default=str),
                 headers=headers,
                 timeout=self.options.timeout,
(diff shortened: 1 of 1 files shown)


ddeschepper:
resolves #1197

lzchen:
@ddeschepper 

I'm curious if you ran into a case in which the envelope was not serializable?

ddeschepper:
> @ddeschepper
> 
> I'm curious if you ran into a case in which the envelope was not serializable?

I did: App Insights supports adding an additional "property bag", which is called customDimensions and which can be filled by adding a key "custom_dimensions" to the dictionary that can be provided to the "extra" argument, e.g:

``` python
logger.log(
    level,
    message,
    extra={
      'custom_dimensions'=my_additional_info
    }
)
```

If any of the values of the dict my_additional_info is not JSON-serializable, serializing the envelope fails. this happens for example with things as trivial as a datetime. The workaround of course, is to only add properties that are either JSON-serializable, or serialize them to string manually, but I think this minor addition makes life a lot easier, while also preventing lost telemetry.

lzchen (approved):

PR #1205, opened 2023-05-30
Azure wants ExceptionDetail.outerId to be an int

### LINKED ISSUE #1201: Azure extension: Data drop 400: 106: Field 'outerId' on type 'ExceptionDetail' is of incorrect type. Expected: int.
I'm using the Python Opencensus library with the Azure extension. The code is sending a string value, where an integer value is expected.

**Steps to reproduce.**
I don't have a reproduceable example yet, but I found where it goes wrong:
https://github.com/census-instrumentation/opencensus-python/blob/b85e4767edcd6524515d263853247335f9ee69f4/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py#L104

Not get a 400 bad request from Azure.

You get a 400 bad request from Azure.

**Additional context.**
-

mgedmin:
Fixes #1201.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
@@ -101,7 +101,7 @@ class AzureExporter(BaseExporter, TransportMixin, ProcessorMixin):
                 data = ExceptionData(
                     exceptions=[{
                         'id': 1,
-                        'outerId': '{}'.format(sd.span_id),
+                        'outerId': 0,
                         'typeName': sd.attributes.get(ERROR_NAME, ''),
                         'message': sd.attributes[ERROR_MESSAGE],
                         'hasFullStack': STACKTRACE in sd.attributes,


contrib/opencensus-ext-azure/opencensus/ext/azure/trace_exporter/__init__.py
```diff
@@ -101,7 +101,7 @@ def span_data_to_envelope(self, sd):
                 data = ExceptionData(
                     exceptions=[{
                         'id': 1,
-                        'outerId': '{}'.format(sd.span_id),
+                        'outerId': sd.span_id,
```
mgedmin, line 104: Actually this might be pointless, if sd.span_id is already a string.

mgedmin:
I think the code changes in this PR are too trivial and therefore not copyrightable, but IANAL and dealing with contributor license agreements is not something I have mental headspace right now.

lzchen:
@mgedmin 

Thanks for your contribution. I have replicated the change [here](https://github.com/census-instrumentation/opencensus-python/pull/1221) so no need for signing the CLA.

PR #1210, opened 2023-06-20
Updating extra properties value to customDimensions property

narenjcs:
This is for loguru module.
We would like to pass common extra parameters into customDimensions in azure insights. But existing code supports with logging module but not supporting for loguru module.


narenjcs:
Hi @aabmass @hectorhdzg @jeremydvoss @lzchen Could you review this PR.

Issue #1219, opened 2023-09-13
ModuleNotFoundError: No module named 'six'


mosneji:
**Describe your environment.**
Python 3.10
opencensus==0.11.2

**Steps to reproduce.**
pip install opencensus==0.11.2
and try to do: from opencensus.common.schedule import Queue, QueueEvent, QueueExitEvent

**What is the expected behavior?**
Modules get imported

**What is the actual behavior?**
Error message: "ModuleNotFoundError: No module named 'six'"

**Additional context.**
"opencensus" is using "six" and "google-api-core"
"google-api-core" is using "google-auth"
"google-auth" is using "six" in the versions previous to "2.23.0" (it was released 2 days ago)
So now if we try to do pip install opencensus==0.11.2, google-auth-2.23.0 will get installed and "six" won't which would break "opencensus"

I guess "opencensus" should add "six" as a dependency as it is being used in its source code.

PR #1223, opened 2023-09-18
Import six

lzchen:
Fixes https://github.com/census-instrumentation/opencensus-python/issues/1219


setup.py
```diff
@@ -45,6 +45,7 @@
         'opencensus-context >= 0.2.dev0',
         'google-api-core >= 1.0.0, < 2.0.0; python_version<"3.6"',
         'google-api-core >= 1.0.0, < 3.0.0; python_version>="3.6"',
+        "six ~= 1.16.0",
```
aabmass, line 48:
```suggestion
        "six ~= 1.16",
```

I imagine any major version 1 should be OK

aabmass (approved):

PR #1229, opened 2023-09-27
[Bug fix] Allow redirects for proxy

GrantAnt:
In the current implementation, proxies are offered but not usable. 
In case requests library has proxy options set, allow_redirect needs to be set to true.

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/common/transport.py
@@ -116,7 +116,7 @@ class TransportMixin(object):
 
             response = requests.post(
                 url=endpoint,
-                data=json.dumps(envelopes),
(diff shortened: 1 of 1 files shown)


GrantAnt:
@lzchen Build Job for py39 failed. It seems not be connected to my code changes and last week the build was working on master. 
Just restart build tool?

lzchen:
@GrantAnt 

Looks like build is failing for lint.

GrantAnt:
@lzchen which linter is executed? Can I run it locally?

lzchen:
@GrantAnt 

image

Looks like it's flake8. You can see the lint steps here: https://github.com/census-instrumentation/opencensus-python/blob/master/tox.ini#L68

GrantAnt:
@lzchen How could I be so blind. Thank you so much :) Whitespaces are resolved now

lzchen (approved):

PR #1234, opened 2023-11-15
fix typename data drop

jeremydvoss:
When using FastAPI's HTTPException, ERROR_NAME cannot be pulled from the exception by the trace exporter. Defaulting to 'Exception'.


lzchen:
@jeremydvoss 

Can we add a test for this? Also looks like you need to sign the CLA

lzchen (approved):

PR #1239, opened 2023-12-27
Update httpx and fastapi bitmap

lzchen:


opencensus/trace/integrations.py
```diff
@@ -32,8 +32,8 @@ class _Integrations:
     PYRAMID = 512
```
jeremydvoss, line 32: Update the changelog

jeremydvoss (approved):

PR #1248, opened 2025-02-06
Remove custom createLock method in BaseLogHandler

danigm:
This patch removes the override of createLock method so this class will use the default logging.Handler method.

The tests are failing for python 3.13 because in this version there's a direct usage of self.lock instead of calling self.acquire(): https://github.com/python/cpython/blob/3.13/Lib/logging/__init__.py#L1025-L1026

### DIFF
--- a/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
+++ b/contrib/opencensus-ext-azure/opencensus/ext/azure/log_exporter/__init__.py
@@ -102,9 +102,6 @@ class BaseLogHandler(logging.Handler):
             self._worker.stop(timeout)
         super(BaseLogHandler, self).close()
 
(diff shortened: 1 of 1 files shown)


EpicWink:
To keep the current behaviour (of not synchronising handling), you instead may want to set `lock` to some new dummy value, which implements the `threading.Lock` interface. For Python 3.12 and below compatibility, you will need to implement `__enter__`, `__exit__`, `acquire`, `release`, `_at_fork_reinit`

EpicWink (approved):

bjorn-einar-bjartnes-4ss:
👍 We were affected by this, causing tests to fail at Python 3.13. I tried to add these changes to the package directly in the project, and then things works again. Any idea of when this will be officially released in the public package?

